PR #3, opened 2015-10-24
PKZip crypto header extensions

beargiles:
Addition of five classes that recognize PKZip crypto header extensions. Actual encryption is not supported at this time, this patch is solely so users can identify and ignore otherwise unknown extensions.


bodewig:
Oops, this looks bigger than intended :-)

It seems as if out migration to git has made the wrong branch the default branch.  I'll ask the ASF infra team to fix it.  After that, please re-create the PR against the master branch.

Also, I'll try to get in touch with the PKWare folks, just to be safe.

beargiles:
Let me know what you need me to do. I've never done branching and merging with git before, much less across projects, so I'm pretty much fumbling at this point.

It looks like the first step is closing this pull request and waiting for you to merge/change default branch.

bodewig:
Don't feel forced to create a pull request just because we are using git now.  The established workflow of attaching a patch to JIRA is still in place.  Whatever is more convenient to you.

This repo's default branch is master now, so a fresh github clone would be based on the correct branch.

Within the existing clone you can add a master branch that tracks our master

```
git checkout -b master remotes/apache/master
git push origin master -u
```

this is assuming you have a remote named apache defined for our git repository and origin for your github clone of it.

There are several ways you can try to merge your changes to the new master branch.  This ranges from simply copying over the code (the low tech version, that always works) to `cherry-pick`ing the commits from the COMPRESS-294 branch.

PR #5, opened 2015-11-24
Pkware headers

beargiles:
Added several classes reflecting the strong crypto headers for PKWARE encryption. This request does not implement encryption/decryption.

This request does not reflect the strong crypto headers for WinZip encryption.

This request does not reflect the legacy (weak) encryption algorithm for ZIP files.


bodewig:
Thanks

I've merged you patch have a few questions - and would like to see tests :-)  Not sure whether you'd prefer to discuss this here, inside a new PR or on the dev list.

I'm note sure why you've introduced the base class `PKWareExtraHeader`. It might be useful as a container but I don't see why the others inherit from it.

Second, don't you want to register the new classes `ExtraFieldUtils`?

PR #7, opened 2016-02-03
COMPRESS-333: adds buffering on top of RandomAccessFile.

dweiss:
Speeds up 7Z handling an order of magnitude as a result.


tcurdt:
Small change, big impact - that's how we like it :)
Thanks for digging into this!

PR #10, opened 2016-03-01
Zip64Mode.Always now impacts central file header same as local file

weirdkid:
These changes bring the file header entries in the central directory into alignment with the local file headers when zip64Mode is set to Zip64Mode.Always. This change makes the generated Zip64 files more compatible with the [Open Packaging Conventions](https://en.wikipedia.org/wiki/Open_Packaging_Conventions) for Microsoft Office Open XML files, which use zip64 as a container.


bodewig:
Thanks a lot Matt, this makes things quite a bit more consistent.

PR #11, opened 2016-03-16
Bugfix for COMPRESS-343

rpreissel:
The class ...sevenz.Coders.DeflateDecoder does not close (end()) the Deflater and Inflater. This can lead to native memory issues: see https://bugs.openjdk.java.net/browse/JDK-8074108.

This bugfix solves the problem.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -45,6 +45,7 @@ jar, tar, zip, dump, 7z, arj.
     <!-- configuration bits for cutting a release candidate -->
     <commons.release.version>${project.version}</commons.release.version>
     <commons.rc.version>RC1</commons.rc.version>
+    <powermock.version>1.6.4</powermock.version>
   </properties>
 
   <issueManagement>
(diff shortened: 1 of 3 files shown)


garydgregory:
Hi,

Thank you for the patch.

Do you have a unit test that show the problem and it being fixed by the patch?

Gary

rpreissel:
This is a memory issue and from my point of view there is no simple way to test it with an unit test.

I have tried to write an unit test and used ManagementFactory.getMemoryMXBean().getNonHeapMemoryUsage() to check the memory usage before and after creating an archive with many files. But there was no difference with the patch.

But if I look at the process with system utils I see that the version without the patch requires many GB and the version with the patch only 100MB. 

Any ideas?

garydgregory:
Yes, one of:

- Use Mockito or Powermock to assert that the end() method has been called (best)
- add an isClosed method somewhere that sets an ivar once end() is called.

rpreissel:
I added tests with PowerMock/Mockito.

bodewig:
I had naively assumed `InflaterInputStream#close` would free up the resources. Come to think of it, if we pass in the `Inflater` ourselves it makes sense that we are responsible for cleanup.

Many thanks!

I'll grep through the code to ensure the same kind of bug doesn't hit us anywhere else.

PR #12, opened 2016-12-03
COMPRESS-375 Allow the deferred creation of `ZipArchiveEntry` for parallel zips

plamentotev:
This is my attempt to implement the changes suggested in [COMPRESS-375](https://issues.apache.org/jira/browse/COMPRESS-375).


plamentotev:
Hmm, looks like @coveralls acts strange. The first time it says my commit are increasing the code coverage and after that they are decreasing it. And there is no code changes - I only fixed the commit message.

bodewig:
Don't worry about Coveralls, I see the same behaviour for other projects as well.

I think this is related to our Travis CI setup using different JDKs and Coveralls picks whichever build finishes first. For different JDKs you get slightly different coverage values.

PR #13, opened 2017-01-15
BZip2: Use BitInputStream

thomasmey:
And some minor spelling errors I stumbled upon.

### DIFF
--- a/src/main/java/org/apache/commons/compress/compressors/bzip2/BZip2CompressorInputStream.java
+++ b/src/main/java/org/apache/commons/compress/compressors/bzip2/BZip2CompressorInputStream.java
@@ -26,8 +26,11 @@ package org.apache.commons.compress.compressors.bzip2;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.nio.ByteOrder;
+import java.util.Arrays;
 
 import org.apache.commons.compress.compressors.CompressorInputStream;
+import org.apache.commons.compress.utils.BitInputStream;
 
 /**
  * An input stream that decompresses from the BZip2 format to be read as any other stream.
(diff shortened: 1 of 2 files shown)


ebourg:
Is the performance equivalent with a BitInputStream?

thomasmey:
Don't know. How to test? I think it very much depends on target architecture and JVM implementation. I could try to establish a baseline with a test case to check for regression. what do you think?

bodewig:
Some rough comparisons for larger files would be a good indicator.

The existing code is pretty ugly because this seemed to be necessary for acceptable performance back then. "Back then" is JDK 1.3 or 1.4, though. Another aspect was that the code stayed close to libbzip2's C implementation which in theory would have allowed us to port changes - which I don't think has ever happened.

Do you think you could create separate PRs for your javadoc typo fixes and making methods static in GZIP? They are separate commits, which is most important to me, but separate PRs looks even cleaner :-)

bodewig:
Sorry, I didn't realize you had updated the PR, will have a look later.

bodewig:
I'm planning to set up a JMH based benchmark soonish and would like to get some statistical information before merging this.

bodewig:
For starters, https://github.com/bodewig/commons-compress-benchmarks - right now I'm not sure where it'll end up.

bodewig:
@thomasmey I've made some minor changes and pushed it to the branch PR13

My benchmarks at https://github.com/bodewig/commons-compress-benchmarks/wiki/PR13 sees the changed code within the error margins of the results for the master branch. @ebourg do you think this is sufficient or do you want to run tests of your own?

bodewig:
I just realized I had benchmarked compression rather than decompression, I've now updated https://github.com/bodewig/commons-compress-benchmarks/wiki/PR13

Fortunately the result is the same, the results are basically the same (the mean of one measurement is within the error margin of the other). If anything then decompressing bigger files has become a bit faster.

PR #16, opened 2017-01-27
Don't duplicate count() functionality.

thomasmey:


bodewig:
Hmm, the no-arg `read` returns the byte read. So when reading an `'A'` your code would count 65 bytes rather than 1, wouldn't it?

thomasmey:
Oops.

True, my mistake. Maybe we should cover this with a unit test...

Sorry for the fuss.

bodewig:
True, a unit-test wouldn't hurt.

I'm afraid I can't close the PR myself.

PR #17, opened 2017-03-29
CPIO crc overflow resolved for large files

dcollin:
When unpacking a CPIO file containing a large file the crc check will
overflow and throw an IOException("CRC Error...").

Did not find a nice way to test this since it requires a very large
input file.


bodewig:
Great, many thanks,

I've just found https://people.freebsd.org/~kientzle/libarchive/man/cpio.5.txt states

> The CRC format is identical to the new ASCII format described in the previous section except that the magic field is set to ``070702'' and the check field is set to the sum of all bytes in the file data.  This sum is computed treating all bytes as unsigned values and using unsigned arithmetic.  Only the least-significant 32 bits of the sum are stored.

so your patch is certainly correct.

I'm afraid we've got the same bug when writing CPIO archives. I'll take care of it.

PR #20, opened 2017-04-14
COMPRESS-382

tballison:
first draft of preventing OOM in LZMA


src/test/java/org/apache/commons/compress/compressors/DetectCompressorTestCase.java
```diff
@@ -167,6 +171,26 @@ private String detect(String testFileName) throws IOException, CompressorExcepti
...
+        testMemoryLimit("COMPRESS-386");
+    }
+
+    private void testMemoryLimit(String fileName) throws IOException, CompressorException {
+        CompressorStreamFactory fac = new CompressorStreamFactory(true,
+                100);
+        try (InputStream is = new BufferedInputStream(
+                new FileInputStream(getFile(fileName)))) {
+            InputStream compressorInputStream = fac.createCompressorInputStream(is);
+            fail("Should have thrown CompressorMemoryLimitException");
```
kinow: Maybe instead we could let the CompressorMemoryLimitException happen, and declare in the @Test annotation the expected exception.

bodewig: +1 and make it two tests :-)

tballison:
+1 to both suggestions.  Will do.  Thank you.

src/main/java/org/apache/commons/compress/compressors/lzma/LZMACompressorInputStream.java
```diff
@@ -56,7 +56,7 @@ public LZMACompressorInputStream(final InputStream inputStream)
             in = new LZMAInputStream(inputStream, MemoryLimit.getMemoryLimitInKb());
         } catch (org.tukaani.xz.MemoryLimitException e) {
             //convert to commons-compress exception
-            throw new MemoryLimitException("exceeded calculated memory limit", e);
+            throw new MemoryLimitException(e.getMemoryNeeded(), e.getMemoryLimit());
```
kinow: Not looking further into the source code of both MemoryLimitException classes, couldn't this hide the root cause of the exception? As the original `e` exception is ignored?

garydgregory: I always like linking to the cause exception...

tballison: Y.  Good point.  I'll update our MemoryLimitException initializers to both a 2-parameter (needed, limit) and 3-parameter (needed, limit, e)?

kinow: Sounds good. There are a few other places in this pull request where the root exception is discarded too.

tballison: Y, any time we're wrapping tukaani's.

PR #21, opened 2017-04-23
COMPRESS-388: Fix concurrent reads performance

kvr000:
Ticket: https://issues.apache.org/jira/browse/COMPRESS-388

Concurrent reads on the ZipFile archive is terribly slow on multiprocessor systems. On my 4 CPU laptop it shows 26 reads/s vs 2 reads/s on 100MB samples for example.

The cause is the use of synchronized blocks to access the underlying file channel. This may be required for generic SeekableByteChannel but most commonly there is FileChannel implementation which supports lock-free reading from any position (i.e. using pread/pwrite system calls or their equivalent).

With the fix the performance is about 10 times faster (on 4 CPU system, with more processor the difference should grow significantly).


src/main/java/org/apache/commons/compress/archivers/zip/ZipFile.java
```diff
@@ -1081,16 +1082,26 @@ private boolean startsWithLocalFileHeader() throws IOException {
...
-        private static final int MAX_BUF_LEN = 8192;
-        private final ByteBuffer buffer;
-        private long remaining;
-        private long loc;
-        private boolean addDummyByte = false;
+        protected static final int MAX_BUF_LEN = 8192;
+        protected final ByteBuffer buffer;
+        protected long remaining;
+        protected long loc;
+        protected boolean addDummyByte = false;
```
sebbASF: Do these really need to be protected rather than private or package-protected?

src/main/java/org/apache/commons/compress/archivers/zip/ZipFile.java
```diff
@@ -1111,14 +1122,11 @@ public int read() throws IOException {
...
             }
-            synchronized (archive) {
-                archive.position(loc++);
-                int read = read(1);
-                if (read < 0) {
-                    return read;
-                }
-                return buffer.get() & 0xff;
+            int read = read(loc++, 1);
+            if (read < 0) {
```
sebbASF: Surely the increment of loc needs to be synchronised?

bodewig: I think it depends on what we think `synchronized` is supposed to protect against.

As it stands the `synchronized` block also protected concurrent reads from the same `BoundedInputStream` from overwriting their results (protecting `loc` and `buffer` not just the file position).

If we want to keep that level of protection the changes are not going to work. If we say each input stream returned by `getInputStream` can only be used by a single thread, then I think it is fine to keep the increment outside of the block and only synchronise the code that protects the interaction between different streams.

src/main/java/org/apache/commons/compress/archivers/zip/ZipFile.java
kvr000: They can be package protected too, it's in private class anyway. The fields are accessed by the FileChannel specialization so private won't suffice.

src/main/java/org/apache/commons/compress/archivers/zip/ZipFile.java
kvr000: Correct, bodewig. Additionally, note that in these terms the original implementation wasn't thread-safe - in the byte[] read method the loc incrementing is outside the synchronized too.

Anyway, there are two options - either make the full methods synchronized (the single stream would hardly ever be accessed concurrently so keeping smaller scope won't make any difference).

Or, better option would be to keep this implementation unsynchronized as it's wrapped into another decompressing stream anyway and this one already has to be synchronized. The only exception is STORED compression method where we would have to wrap into synchronized proxy.

sebbASF: OK. This needs to be clarified in the Javadoc.

kvr000:
Update, improving few things:
- made the fields private again
- simplified to single read(long pos, ByteBuffer buf) method
- allocating the instance buffer only for single byte read which is rather rare (so far it seems that only Bzip2 uses it for reading the header), all other decompressors and even standard java core readers use internal cache.
- wrapping the byte array passed by parameters instead of creating temporary ByteBuffer and copying the bytes

Performance improved in comparison to previous commit:
- 15%-20% for stored stream
- 1%-3% for deflated stream

kvr000:
Additionally, I was thinking about exposing the entry raw stream starting offset and length via public API so in case of need one can either map it into memory, directly access the raw data (especially useful when zip is just kind of flat storage, being quite popular in games but not only). For me it would help to implement off-heap read-only storage, using standard file format widely supported by lot of tools.

It's quite zip specific (although can be applied to similar containers too) but anyway the API already has lot of zip specific stuff... That piece of information would have to be only moved from ZipFile.Entry to ZipFileEntry. What do you think about it?

PR #23, opened 2017-05-02
Add Brotli Support

pmouawad:
Fix for:
https://issues.apache.org/jira/browse/COMPRESS-392


pom.xml
```diff
@@ -68,6 +68,12 @@ jar, tar, zip, dump, 7z, arj.
       <scope>test</scope>
     </dependency>
     <dependency>
+      <groupId>org.brotli</groupId>
+      <artifactId>dec</artifactId>
+      <version>0.1.1</version>
+      <scope>compile</scope>
```
bindul: Shouldn't this be an optional dependency, just like the xz one?

pom.xml
```diff
@@ -85,6 +91,13 @@ jar, tar, zip, dump, 7z, arj.
       <version>${powermock.version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>commons-codec</groupId>
+      <artifactId>commons-codec</artifactId>
+      <version>1.10</version>
+      <scope>test</scope>
+    </dependency>
+
```
bodewig: I'd prefer encoding things by hand rather than adding a new dependency just for tests.

bodewig:
Many thanks Philippe

see my comments on the test.

Also I haven't checked whether the brotli lib is an OSGi bundle, we may need to update the bundle-plugin configuration.

pom.xml
bodewig: +1

garydgregory:
Thank you for your patch. The only thing missing at first glance is a patch to `org.apache.commons.compress.compressors.CompressorStreamFactory`.

PR #24, opened 2017-05-05
COMPRESS-391: Allow alignment on zip content

kvr000:


bodewig:
Many thanks @kvr000 

Have you thought about adding a class implementing `ZipExtraField` for padding? You could add that to the entry if you detect padding is necessary and could remove code that now needs to know about the encoding of extra fields from `ZipArchiveOutputStream`. The central directory size should be 0 then.

It would also preserve the padding information when an archive is read which may be useful when re-packing existing entries as you could strip or modify the padding extra field directly.

Finally I wonder whether `setAlignment` should reject values bigger than 64k right away.

kvr000:
@bodewig : Thanks for suggestions. There are definitely some pros with the approach, especially if it reaches Zip RFC :-)

On the other hand, I don't think it could be made completely independent on ZipArchiveOutputStream as it has to know current stream offset on which to base the alignment.

But I'll check in details...

bodewig:
I don't see a chance of making it independent of `ZipArchiveOutputStream`, you are certainly correct.

I was thinking along the lines of `ZipArchiveOutputStream` calculates the length needed for proper alignment, creates an instance of the new extra field passing in the size information necessary and adds it to the `ZipArchiveEntry` - after that the code in `ZipArchiveOutputStream` can be left unchanged.

bodewig:
@kvr000 what do you think about https://github.com/apache/commons-compress/commit/620196621e15a87cdd5e3382504bd8a9829f4698 ?

PR #27, opened 2017-06-05
COMPRESS-400 

sesuncedu:
Add extra header map to tar archive entry.
Move PAX header processing code from TarArchiveInputStream to TarAchiveEntry.
Use same code for processing user supplied extra headers - thus setting "size "changes the value of getSize().
Add any extra PAX headers to output map when putting entry in TarArchiveOutputStream.
Add simple tests for getting/setting xattr, setting "size", and round tripping.

This PR uses COMPRESS-399 as a base.  
To make it easier to cherry-pick + rebase the PR has been split in two.

 1d9b3c88455caceca81f0ff6b7eca0958c631359 contains just the code changes and test changes,  and does not bump the minor package version .  This will cause the bundle:baseline verify goal in 399 to  break the build.

82405c13bd2688817108d3f2854387b3417a764d  increases the package minor version to the correct value


bodewig:
Thanks, do you think you could re-create the PR without the commits that belong to #26 ?

PR #28, opened 2017-06-11
COMPRESS-405 Create Fixed Length Block OutputStream / WriteableByteChannel

sesuncedu:
This PR provides a new class that is an OutputStream and WritableByteChannel, and which supports writing to a destination output stream or byte channel in fixed size blocks. Internally, all writes are made using NIO.
If the destination is a  FileOutputStream the existing channel is used.
Other OutputStreams are wrapped with a custom channel implementation which does not attempt to split writes into chunks.

If the target channel fails to write the entire buffer in a single call, an exception is thrown,

Incoming data is accumulated in a ByteBuffer until a complete block is ready, then written to the target.
If  WritableByteChannel::write(ByteBuffer)  method is called, the code will attempt to avoid copying data into the buffer if the buffer is empty, and a complete block is available.

**The class and UnitTest are in compress/utils . This is a MINOR change - thus the version number for the package should be increated to 1.15.0**

### DIFF
--- a/src/main/java/org/apache/commons/compress/utils/FixedLengthBlockOutputStream.java
+++ b/src/main/java/org/apache/commons/compress/utils/FixedLengthBlockOutputStream.java
@@ -1,3 +1,21 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/commons/compress/utils/FixedLengthBlockOutputStream.java
```diff
@@ -0,0 +1,227 @@
...
+import java.io.IOException;
+import java.io.OutputStream;
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+import java.nio.channels.ClosedChannelException;
+import java.nio.channels.WritableByteChannel;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+public class FixedLengthBlockOutputStream extends OutputStream implements WritableByteChannel,
+    AutoCloseable {
```
bodewig: picking nits here but I'm curious. Why do you add `AutoCloseable` explicitly? It's already inherited from both `OutputStream` and `Channel`.

sesuncedu: Huh. Not only can I think of no good reason, I can't think of what I was thinking. 
 It might be because I've been writing a bunch of stuff that I wanted to use with try-with-resources and which didn't have it (reactive stuff, and stuff that wrapped  closeables), so I added AC automatically when I implemented close :-)

PR #33, opened 2017-06-13
add-some-Unit-Tests Added some Unit Tests to increase code coverage.

TheRealHaui:
I have added some Unit Tests to increase code coverage and want to contribute them.

### DIFF
--- a/src/test/java/org/apache/commons/compress/compressors/xz/XZCompressorOutputStreamTest.java
+++ b/src/test/java/org/apache/commons/compress/compressors/xz/XZCompressorOutputStreamTest.java
@@ -1,0 +1,51 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.commons.compress.compressors.xz;
+
+import org.junit.Test;
+
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+
+import static org.junit.Assert.assertEquals;
+
+
+/**
+ * Unit tests for class {@link XZCompressorOutputStream}.
+ *
+ * @date 16.06.2017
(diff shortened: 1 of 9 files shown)


TheRealHaui:
@bodewig 
Could you be so kind an review my contribution and - of course - most of all pull it into the code base.

bodewig:
Fortunately I'm not the only one who could merge this :-)

More seriously, it may take a bit of time until I get there, but I will.

bodewig:
Thanks @TheRealHaui 

I agree with your comment on `ChecksumCalculatingInputStreamTest#testGetValueThrowsNullPointerException` this looks like a bug. So I'd rather remove the test and see the bug fixed. Do you want to take a stab?

Also, the Commons Compress community has decided to not use author tags, I'd have to strip yours before merging the PR. Could you please remove them yourself if you modify this PR? Feel free to add yourself as a contributor to the POM.

TheRealHaui:
@bodewig 
Thank you for your kind response!
Really appreciate that!

Therefore I've made all the changes you requested/proposed.
And of course added myself as a contributor as heavily requested by you.
Really couldn't disappoint you regarding that special topic. :-)

Furthermore I've created a Jira Task for the bespoken bug: https://issues.apache.org/jira/browse/COMPRESS-412
And fixed it in an own branch which I've commited, created tests for and pushed too.
And which pull request I am going to link here in the next comment after I've created it.

TheRealHaui:
Link to the pull request: https://github.com/apache/commons-compress/pull/35

bodewig:
Thanks a lot.

src/test/java/org/apache/commons/compress/compressors/xz/XZCompressorOutputStreamTest.java
```diff
@@ -0,0 +1,51 @@
...
+
+    @Test
+    public void testWrite() throws IOException {
+
+        ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream(4590);
+        XZCompressorOutputStream xZCompressorOutputStream = new XZCompressorOutputStream(byteArrayOutputStream);
+        xZCompressorOutputStream.write(4590);
+
+        assertEquals(24, byteArrayOutputStream.size());
+        assertEquals("\uFFFD7zXZ\u0000\u0000\u0004\uFFFD\u05B4F\u0002\u0000!\u0001\u0016\u0000\u0000\u0000t/\uFFFD", byteArrayOutputStream.toString());
```
sesuncedu: It's usually not a good idea to convert an arbitrary byte array to a string, as on can't rely on  a specific default  charset, and you may end up with characters mangled /mapped. 

I thinkThat hamcrest has array matchers; if not,  assertTrue(Arrays.equals(exp,act)) should be OK.
 I can't remember if there's an assertEquals for byte[]

bodewig: Thank you for catching this. I've reformulated the test with commit b53ead4 - I think we are not that much interested in the exact bytes created but rather that using `write` creates a valid XZ output.

JUnit's `Assert` class contains `assertArrayEquals`.

TheRealHaui:
@bodewig 
You're welcome, was a pleasure!

PR #36, opened 2017-06-17
COMPRESS-410 Remove Non-NIO character set encoders.

sesuncedu:
As a special case, the UTF-8 encoder will replace malformed / unmappable input with '?'.
This behavior is required for compatibility with existing behavior.


sesuncedu:
reopening just to see if I didn't need to create a new branch and cherry pick

PR #37, opened 2017-06-19
COMPRESS-410 - cherry picked  to fix merging issues

sesuncedu:
Remove non-NIO character set encoders. 

Fixes rebasing related issues in previous PR


sesuncedu:
There seems to be something amiss with Hudson (it can't checkout origin/master ... )

bodewig:
Thanks a lot Simon,

Please take a look at fde6670 - I've only performed some minor changes that I didn't feel like hackling you for. Things that may be a tiny bit more than just cosmetics:

* I'm not sure why you went for the literal string `UTF-8` rather than the constant and `isUTF8` method in `ZipEncodingHelper`, I've changed that (kinda) back
* I've removed the `assert` from `NioZipEncoding.encode`

PR #38, opened 2017-06-19
Testing hudson/jenkins (added contributor line to pom.xml so that there would be a change)

sesuncedu:
They've been failing PRs on the checkout of master.  Lets try this.


sesuncedu:
Hudkins is happy. Happy happy Hudkins.

So I will close this.  Though I think some of the changes might actually be useful.

PR #39, opened 2017-06-22
COMPRESS-413:  Performance improvements for Travis builds 

sesuncedu:
Change travis configuration to
- skip pre-build install (which runs the install goal).
- remove after_success code.
- use mvnw to use up-to-date version of maven.
- run all goals/phases previously included in after_sucess phase, using -Ptravis-jacoco profile if one exists.
  - clean
  - apache-rat:check
  - test
  - jacoco:report
  - coveralls:report

  - This will fail the build if any of the above goals fail. The goals that are most likely to fail are: apache-rat:check and test. coveralls configuration is changed in pom to not fail the build if the service is not available (e.g. for forks that do not have repository registered with coveralls).

 -  Note that the apache-rat: check could be added to the pom (executing by default in the validate phase)
 -  Also note that the jacoco:check goal can  be added to the pom (executing by default in the verify phase).
 -  the script could then be changed to execute the verify goal.
 - coveralls stuff could be run in an after_success script. That script could restrict coveralls notification to a single point in the jdk matrix


sesuncedu:
Randomly adding  coverage tests  for coveralls

bodewig:
Thanks Simon.

TBH I'm not too sure why we've got Travis builds at all, probably so we can run Coveralls. I'll leave this request open for those more interested in the Travis CI setup. I'll rather invest some time in adding a Jenkins matrix build for the JDKs we want to support instead.

Commons Compress aims to be compatible with Java7, no matter how old or deprecated, but we already cover that with Jenkins - the pull request builder over there uses Oracle JDK 7, so we've got that covered independent of Travis CI. We do run SonarQube for Compress already.

Personally I dislike `mvnw` scripts, but this is probably a matter of taste. I won't veto its introduction.

Please strip the changes to `AsiExtraField` from this PR, they are completely unrelated.

sesuncedu:
Closing because there's a new PR without the coveralls bait. 
Coveralls complains. 

The mvnw is not essential yet, but it's good to have  3.5.0 in place.

sesuncedu:
Replaced by PR https://github.com/apache/commons-compress/pull/43

PR #42, opened 2017-06-23
COMPRESS-407:  Allow use of invalid record sizes if a system property is set.

sesuncedu:
I do not recommend using this approach.  #40 is the preferred solution.


sesuncedu:
The system property is can be discovered by reading the source code. This information could be added to the exception message.

PR #43, opened 2017-06-24
COMPRESS-413 : Improved build performance with travis

sesuncedu:
Now without unrelated sacrifices to coveralls


sesuncedu:
That's why there was random coveralls stuff in the PR... 
coverage/coveralls — Coverage decreased (-0.02%) to 84.742%

bodewig:
We all know coveralls is flaky at the sub-percent/sub-permille level.

The first paragraphs of https://github.com/apache/commons-compress/pull/39#issuecomment-310845353 are still true :-) The matrix build is https://builds.apache.org/view/Apache%20Commons/job/Commons-Compress%20JDK-Matrix%20Build/

sesuncedu:
I think there is also some jacoco oddness related to synthetic methods in certain JDKs , but mostly I blame coveralls, since I'm not sure jacoco is wrong :-)  
IntelliJ has a bunch of different coverage tracers, including jacoco, and they have a spread of class and line counts that seems to be all synthetic accessor related. 
Oh for statement level "ignore".

sesuncedu:
Travis has some nice properties, especially when working with forks. It's very easy to enable travis for a fork, so CI can be done before a PR goes in.

 Travis can also run really quickly, but only when tuned properly, and when the system is not running at capacity. If the stars are aligned, you can get every job in a big 'ol  matrix running in parallel (each container build gets 2 cores and max 4GB of memory), which is  nice if they're free.  The critical thing is to be using the container environments and not the VM, as the startup time is tiny.  The annoying thing about Travis is that the default environment is running Ubuntu 12.04, which is well past its LTS date. The trusty (14.04) container is *still* marked as beta, despite being on a production level update cycle (set group to edge for that real beta experience). The 12.04 containers all have to do a lot of package replacement on startup, which takes away some of the benefits of having a container ). I kept the home server on 12.04 until the end of LTS because Travis, then went  to xenial (16.04) which is the current LTS, and has nice things like proper zfs (hey- an apache connection). 

Hudkins farms in my mental model usually seem to be more congested, but there are good reasons not to believe that.

bodewig:
Builds for PRs are already covered - https://builds.apache.org/job/Commons-Compress%20PullRequest%20Builder/ - and we could add support for building arbitrary branches if we wanted to using build pipelines. We just haven't got that many branches outside of PRs (master and the stalled compress-2.0 experiment). :-)

.travis.yml
```diff
@@ -13,17 +13,26 @@
...
 sudo: false
+# Use Trusty as dist
+dist: trusty
+language: java
+
 
 jdk:
   - openjdk7
-  - oraclejdk7
-  - oraclejdk8
```
garydgregory: Why drop the Oracle JDK builds?

sesuncedu:
Oracle JDK 7 has been dead for two years. Openjdk 7 has some back ports of security patches, but both are gone from most Linux distros. 

Openjdk and oraclejdk 8 are essentially in sync, with the only differences really being the oracle proprietary stuff. There's not much win testing against both unless the oracle commercial stuff is being used.

garydgregory:
We all know that Java 7 is EOL but that is the required platform currently. As we have seen in Commons with other components, the IBM JDK sometimes gives surprising results. It is my opinion that we do loose confidence that all is well everywhere by omitting building on Oracle's JDK.

PR #45, opened 2017-06-25
execute-Unit-Tests-parallel-with-Maven

TheRealHaui:
Maven surefire plugin executes Unit Tests in parallel now.
Further information if desired: https://dzone.com/articles/running-junit-tests-parallel

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -374,6 +374,7 @@ jar, tar, zip, dump, 7z, arj.
               <includes>
                 <include>**/zip/*IT.java</include>
               </includes>
+              <parallel>classes</parallel>
             </configuration>
           </plugin>
         </plugins>
@@ -389,6 +390,7 @@ jar, tar, zip, dump, 7z, arj.
               <includes>
                 <include>**/tar/*IT.java</include>
               </includes>
+              <parallel>classes</parallel>
             </configuration>
           </plugin>
         </plugins>


TheRealHaui:
Obviously, coveralls makes any kind of mistake measuring the coverage due to the fact that I've not edited any code.

bodewig:
Coveralls seems to be sensitive on which of the several JDKs we build with on Travis finishes first. Don't worry, this seems to be normal.

bodewig:
Your change is for the IT profile tests which aren't run on a regular basis. Each of these surefire instances will run the tests of a single class only. There won't be any difference between running them parallel "per class" or not in parallel at all.

As these tests are resource hogs - file system and memory in particular - they wouldn't be good candidates for running in parallel anyway.

For "normal" tests we inherit the settings from [Commons Parent](https://svn.apache.org/repos/asf/commons/proper/commons-parent/trunk/),

TBH I don't know whether jacoco would be compatible with a parallel execution of tests.

TheRealHaui:
You're right the change is for the *IT tests.
Was sure I made the change for the normal Unit Tests ...

However, I've tested the change on my local machine excessively and didn't encounter any problems.

Thought this might be a good idea to incorporate.
However ... / Anyway it wouldn't have been in any way a great improvement.

Thank you for your reply.
I close the pull request.

PR #47, opened 2017-06-26
Add additional unit tests

TheRealHaui:
Added additional Unit Tests to increase coverage.


bodewig:
Thanks @TheRealHaui 

I'd like to cherry-pick 7e11b54 rather than merging the PR as is, as the PR includes the "parallel" changes to the ITs that haven't got any effect (as there is only one class per execution anyway).

TheRealHaui:
You're welcome.
Don't know why the change included the pom.xml changes.
Explicitly created an own branch for that.
Anyway ...

PR #49, opened 2017-07-04
COMPRESS-413 set execute back on mvnw

sesuncedu:
mvnw seems to have lost it's execute mode bits. 

Merge asap


bodewig:
I've made `mvnw` executable again with eef4e3c but didn't want to mix it with the unrelated changes

sesuncedu:
Did you also re-add .mvn?

bodewig:
The directory itself has been there, including the properties file, only the jar was missing (and I added it).

PR #50, opened 2017-08-02
Missing Test Case for octal to binary conversion

ileontiuc:
In order to ensure that there is at least one test that will enter the octal string conversion, we  added a test that for value = Long.MAX\_VALUE with an insufficient small length of 8 for the buffer. This will check that the value can not be converted into an octal and throw an exception with an appropriate message which we check. This missing test case was an outcome of our mutation testing analysis performed on your system.


garydgregory:
Thank you for your patch. I will take a look today.

bodewig:
Many thanks @ileontiuc - patch applied.

PR #54, opened 2017-10-05
Let parseName stop at first NUL

rspilker:
Fixes COMPRESS-421


rspilker:
I suspect that the coverage went down. not because I added untested lines, but because I reduced the number of lines in the file.

bodewig:
Don't worry about the coverage too much, it also depends on which JDK build of travis has finished first and fed the results into coveralls.

I vaguely recall there is a reason why we look for the last NUL, will have to re-read the code, the tar spec(s) and maybe even the code history.

bodewig:
The reason is not buried inside the spec, it merely seems to have been a performance optimization. Before introducing the encoding support we've been looking for NUL from the start of the name fields and I changed it with 69ceb4e many moons ago.

I'll apply this PR, many thanks!

PR #56, opened 2017-11-20
[COMPRESS-429] Provide information about source of name / comment field values (e.g. Unicode)

dalbani:
https://issues.apache.org/jira/browse/COMPRESS-429

### DIFF
--- a/src/main/java/org/apache/commons/compress/archivers/zip/ZipUtil.java
+++ b/src/main/java/org/apache/commons/compress/archivers/zip/ZipUtil.java
@@ -239,7 +239,7 @@ public abstract class ZipUtil {
                                                            originalNameBytes);
         if (newName != null && !originalName.equals(newName)) {
             ze.setName(newName);
-            ze.setHasUnicodeName(true);
+            ze.setNameSource(ZipArchiveEntry.NameSource.UNICODE_EXTRA_FIELD);
         }
 
         if (commentBytes != null && commentBytes.length > 0) {
@@ -249,7 +249,7 @@ public abstract class ZipUtil {
                 getUnicodeStringIfOriginalMatches(cmt, commentBytes);
             if (newComment != null) {
                 ze.setComment(newComment);
-                ze.setHasUnicodeComment(true);
+                ze.setCommentSource(ZipArchiveEntry.CommentSource.UNICODE_EXTRA_FIELD);
             }
         }
     }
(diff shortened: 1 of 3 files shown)


bodewig:
Many thanks @dalbani this looks good. I've got two wishes, though:

* could you please revert the change in imports to `ZipFile`? We don't want to use `*`-imports add all.
* the new enums could use some javadoc inclusing a @since marker

Of course, adding tests woudln't hurt , either;-)

bodewig:
@dalbani I've extended your PR with https://github.com/apache/commons-compress/tree/COMPRESS-429 - could you please have a look?

src/main/java/org/apache/commons/compress/archivers/zip/ZipUtil.java
```diff
@@ -239,6 +239,7 @@ static void setNameAndCommentFromExtraFields(final ZipArchiveEntry ze,
                                                            originalNameBytes);
         if (newName != null && !originalName.equals(newName)) {
```
dalbani: When I think about it, I would remove the second test in this `if`, so that the Unicode field always have "priority" over the original name as source.
Scenario where it would matter is when the `originalName` contains a non ASCII value (which is properly decoded) and which matches with the Unicode field value.

dalbani:
@bodewig: my apologies for not getting back to you sooner, I was on vacation.
What you wrote looks just fine to me — I would definitely not have done any better.
Kudos!
I have just added one comment though for a very specific use case.

src/main/java/org/apache/commons/compress/archivers/zip/ZipUtil.java
bodewig: I think the second part is there for a reason, even though my memory is failing me right now - and I stumbled over it myself when I realized the same check is not present for the comment.

I understand what you mean, though, changed with c2906092

bodewig: After some digging through the history it seems the reason no longer exists, I've removed the second part with ced2075c.

bodewig:
Don't worry about any delay @dalbani many thanks.

I've merged the branch to master by now, will be included in Compress 1.16.

dalbani:
@bodewig: thanks for your great overall reactivity for this feature request! I look forward for release 1.16.

PR #58, opened 2018-01-06
An example failing deflate64 zip file

dweiss:


bodewig:
Thanks @dweiss !

It looks as if 07ed5450 has fixed this. As I'm not really happy with the way I fixed it, I'm going to add your archive to the testcases as a way to prevent regressions if I look for a solution I like better.

PR #60, opened 2018-01-14
Replace custom 'dummy' byte addition for the odd inflater contract wi…

dweiss:
…th a sequence input stream. Clean up the code a bit. I did not use NOSONAR flags because I don't use sonar, so it may require some tweaking.

### DIFF
--- a/src/main/java/org/apache/commons/compress/archivers/zip/ZipFile.java
+++ b/src/main/java/org/apache/commons/compress/archivers/zip/ZipFile.java
@@ -18,11 +18,13 @@
 package org.apache.commons.compress.archivers.zip;
 
 import java.io.BufferedInputStream;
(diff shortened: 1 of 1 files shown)


dweiss:
Also, to be honest -- this boundedinputstream has synchronized read methods and I've no idea why they're there. I don't think we have to create thread-safe streams here (is this a requirement anyhow?). This isn't included in the patch, but I'd drop synchronized from boundedinputstream entirely.

bodewig:
your PR and 43e8d3d collided mid-flight, sorry about that, I'll incorporate a few of your changes that I have missed.

Honestly, I don't recall why the methods are synchronized anymore either.

dweiss:
No problem at all; sorry I missed that other place that was using the inflater (glad you spotted it); had kids on my back and couldn't focus.

PR #61, opened 2018-01-14
Code Cleanup.

shahabkondri:


src/main/java/org/apache/commons/compress/archivers/ar/ArArchiveEntry.java
```diff
@@ -179,12 +179,7 @@ public boolean equals(final Object obj) {
...
         if (name == null) {
-            if (other.name != null) {
-                return false;
-            }
-        } else if (!name.equals(other.name)) {
-            return false;
-        }
-        return true;
+            return other.name == null;
+        } else return name.equals(other.name);
```
bodewig: please add braces for the else block.

src/main/java/org/apache/commons/compress/archivers/cpio/CpioArchiveEntry.java
```diff
@@ -886,12 +886,7 @@ public boolean equals(final Object obj) {
...
         if (name == null) {
-            if (other.name != null) {
-                return false;
-            }
-        } else if (!name.equals(other.name)) {
-            return false;
-        }
-        return true;
+            return other.name == null;
+        } else return name.equals(other.name);
```
bodewig: please add braces for the else block.

src/main/java/org/apache/commons/compress/archivers/dump/DumpArchiveEntry.java
```diff
@@ -418,12 +418,8 @@ public boolean equals(final Object o) {
...
         // summary is always null right now, but this may change some day
-        if ((summary == null && rhs.summary != null) // NOSONAR
-            || (summary != null && !summary.equals(rhs.summary))) { // NOSONAR
-            return false;
-        }
-
-        return true;
+        return (summary != null || rhs.summary == null) // NOSONAR
+                && (summary == null || summary.equals(rhs.summary));
     }
```
bodewig: I find this version a lot harder to read than the original expression. I'm totally fine with getting rid of the `if`, but please keep the original logic.

shahabkondri:
Thanks for reviewing. 
The suffix L is preferred, because the letter l (ell) is often hard to distinguish from the digit 1 (one).

src/main/java/org/apache/commons/compress/archivers/dump/DumpArchiveEntry.java
shahabkondri: I've double checked and reverted to the original logic.

bodewig:
merged, many thanks!

PR #62, opened 2018-01-20
Code cleanups.

shahabkondri:


bodewig:
I'm afraid removing throws clauses from public methods is a source-incompatible change, we cannot do that in a  minor release (and a major release is currently not planned to do).

PR #63, opened 2018-03-31
(doc) fix the error in the gzip compression example

lazar2606:
1. GZipCompressorOutputStream does not exist
2. BufferedInputStream is invalid class for creating output stream


bodewig:
Many thanks, obviously there has been a little bit too much copy-paste going on. :-)

PR #65, opened 2018-07-01
fix typo in syntax

boyarsky:
Files.newInputStream is a static method so the "new" keyword shouldn't be here.


bodewig:
Merged, many thanks!

PR #67, opened 2018-07-09
[COMPRESS-459] Fix reading of multibyte name entries

ctron:
This fixes COMPRESS-459 by using the name number of bytes from the field
in the stream instead of relying on the assumption that each character
is exactly one byte, which isn't true for UTF-8, UTF-16 or other
multi-byte character encodings.


bodewig:
Thanks @ctron!

Don't we need to adjust `getHeaderPadCount` in `readOldBinaryEntry` as well?

At first I stumbled over the `-1` in the call to `getHeaderPadCount` but I think it is because the method already accounts for the trailing NUL, I guess a comment could help :-)

One more nit, you removed a since tag from one of `CpioArchiveEntry`'s constructors by accident.

Also, but this probably is a separate issue: `CpioArchiveOutputStream` seems to have the same problem (even worse, as the length we write into the header may be wrong). It seems we should deprecate the no-arg version of `getHeaderPadCount` as it only works for single bye encodings.

ctron:
Thanks @bodewig for reviewing this PR. Yes you are right, with everything :wink: 

I amended the PR, fixing the issues with the original PR and added the functionality to the output stream as well. The made them two commits for to easier understand what was fixed and what added. It can be squashed later on if required.

bodewig:
Many thanks. We are not that squash-happy over here, two commits is fine.

I would have gone for `ZipEncoding` rather than `Charset` for consistency, but can change that later myself, I really don't want to play "fetch me a rock" here. Will merge the PR soonish.

ctron:
Yes, that might have been a solution as well. But I was thinking to use something more Java-standard. If you want me to change this, I can do this as well.

bodewig:
`ZipEncoding` is a better choice for our internal use as it is used to encode the name (and deals with "use null as the encoding to use the platform's default encoding" transparently). You don't have to make the changes yourself (but if you do, please add spaces around operators :-) ), but I won't complain if you are faster than me.

bodewig:
@ctron I've just started merging your changes locally

bodewig:
Many thanks!

PR #68, opened 2018-07-24
Allow specifying Zstandard compression level

cgrushko:
ZstdCompressorOutputStream allows data compression using Zstandard, but the default compression level of 3 is hard-coded.
Programs that wish to use a different compression level must reimplement ZstdCompressorOutputStream.
This PR adds a constructor that takes a compression level and passes it to 
`new ZstdOutputStream(...)`.

### DIFF
--- a/src/main/java/org/apache/commons/compress/compressors/zstandard/ZstdCompressorOutputStream.java
+++ b/src/main/java/org/apache/commons/compress/compressors/zstandard/ZstdCompressorOutputStream.java
@@ -34,8 +34,20 @@ public class ZstdCompressorOutputStream extends CompressorOutputStream {
 
     private final ZstdOutputStream encOS;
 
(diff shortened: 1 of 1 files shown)


bodewig:
Thank you @cgrushko 

could you please add javadocs that explain the parameters (and a 1.18 @since marker would be a bonus).

bodewig:
Understood. I somehow assumed you'd know what the parameters did :-) I'll merge the PR as is.

In a way I wonder why anybody who needs that much control over zstd output is using Commons Compress as indirection at all.

cgrushko:
heh :)
Commons compress provides tar support which is almost a must for using
zstd.

Thanks!

bodewig:
Well, you could still use zstd-jni directly instead of Commons Compress for the zstd part.

Many thanks for the PR.

PR #70, opened 2018-09-28
Use Java 7 Files.createTempDirectory instead of creating file, deleti…

DaGeRe:
…ng it and create directory

The self-written mkdir deletes a temporary file with a prefix, deletes it and creates an directory with the same name. Creating a temporary file by Files.createTempDirectory so is cleaner and faster.


bodewig:
Many thanks. I agree this is cleaner, the existing method is a leftover of the time before we made Java7 our baseline.

While I don't think we consider the tests a public API of ours, the change would be smaller and keep perfect backwards compatibility if you "just" changed the implementation of `mkdir`.

bodewig:
I've just commited a minimal implementation of the suggested change, many thanks.

PR #74, opened 2019-02-11
Correct EOF check in ArArchiveInputStream

akbertram:
The original code appears to be checking for end-of-file
using the InputStream.available() method.

This however, misunderstands the InputStream API. The available()
method only returns an estimate, and cannot be used
to check for the remaining bytes in the file. From the documentation:

> Returns an estimate of the number of bytes that can be read (or
> skipped over) from this input stream without blocking by the next
> invocation of a method for this input stream. The next invocation
> might be the same thread or another thread.  A single read or skip of this
> many bytes will not block, but may read or skip fewer bytes.
> **Note that while some implementations of InputStream will return
> the total number of bytes in the stream, many will not.**  It is
> never correct to use the return value of this method to allocate
> a buffer intended to hold all data in this stream.

This patch includes a unit test that demonstrates the bug and
verifies the fix.

### DIFF
--- a/src/main/java/org/apache/commons/compress/archivers/ar/ArArchiveInputStream.java
+++ b/src/main/java/org/apache/commons/compress/archivers/ar/ArArchiveInputStream.java
@@ -121,13 +121,12 @@ public class ArArchiveInputStream extends ArchiveInputStream {
             trackReadBytes(1);
         }
 
-        if (input.available() == 0) {
-            return null;
-        }
-
         {
             final int read = IOUtils.readFully(input, metaData);
             trackReadBytes(read);
+            if (read == 0) {
+                return null;
+            }
             if (read < metaData.length) {
                 throw new IOException("truncated ar archive");
             }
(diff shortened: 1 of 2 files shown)


akbertram:
FYI: The coverage decrease reported by coveralls is in bzip2, which is untouched by this pull request.

https://coveralls.io/builds/21576720/source?filename=src/main/java/org/apache/commons/compress/compressors/bzip2/BlockSort.java

The Jenkins failure relates to a javadoc problem.

bodewig:
Thanks, I haven't looked at you PR so far, hope to find time the coming days.

But I fixed the javadoc error wit 844faded and made a note to self to check why none of the existing Jenkns builds for master caught this.

Don't worry about the coverage change, permille values seem to depend on the JRE and so the value depends on which travis build finishes first.

bodewig:
Had a look after I typed the last comment and realized this was good to go as it is. Many thanks again.

akbertram:
Terrific, thanks!

PR #76, opened 2019-04-08
COMPRESS-481 Allow passing memory limit to SevenZFile

theobisproject:


src/main/java/org/apache/commons/compress/archivers/sevenz/LZMA2Decoder.java
```diff
@@ -33,9 +34,12 @@
 
     @Override
     InputStream decode(final String archiveName, final InputStream in, final long uncompressedLength,
-            final Coder coder, final byte[] password) throws IOException {
+            final Coder coder, final byte[] password, int maxMemoryLimitInKb) throws IOException {
         try {
             final int dictionarySize = getDictionarySize(coder);
+            if ((dictionarySize / 1024) > maxMemoryLimitInKb) {
```
bodewig: I think you should call `LZMAInputStream.getMemoryUsage` rather than evaluate the dictionary size only. Too bad the constructor we use doesn't support the `memoryLimit` parameter other constructors allow.

src/main/java/org/apache/commons/compress/archivers/sevenz/LZMADecoder.java
```diff
@@ -34,11 +35,13 @@
 
     @Override
     InputStream decode(final String archiveName, final InputStream in, final long uncompressedLength,
-            final Coder coder, final byte[] password) throws IOException {
+            final Coder coder, final byte[] password, int maxMemoryLimitInKb) throws IOException {
         final byte propsByte = coder.properties[0];
         final int dictSize = getDictionarySize(coder);
         if (dictSize > LZMAInputStream.DICT_SIZE_MAX) {
             throw new IOException("Dictionary larger than 4GiB maximum size used in " + archiveName);
+        } else if ((dictSize / 1024) > maxMemoryLimitInKb) {
```
bodewig: I think you should call `LZMAInputStream.getMemoryUsage` rather than evaluate the dictionary size only. Too bad the constructor we use doesn't support the `memoryLimit` parameter other constructors allow.

src/main/java/org/apache/commons/compress/archivers/sevenz/LZMA2Decoder.java
bodewig: sorry about the previous comment which applied to `LZMAInputStream`. `LZMA2InputStream` also provides a static `getMemoryUsage` which might be a better option to calculate memory consumption.

bodewig:
Many thanks. Looks good, but see my inline comments.

Also I'd ask you to add  to `SevenZFile`'s javadocs that only some of the codecs will honor the memory limit.

Adding a blurb to http://commons.apache.org/proper/commons-compress/examples.html#Restricting_Memory_Usage would be good as well.

src/main/java/org/apache/commons/compress/archivers/sevenz/LZMADecoder.java
theobisproject: Didn't know about that. I changed the code according to your comment. Do

src/main/java/org/apache/commons/compress/archivers/sevenz/LZMA2Decoder.java
theobisproject: Code changed according to your comment.

theobisproject:
I added a paragraph to the example page. Feel free to change it afterwards.

bodewig:
Great, many thanks!

The Travis Build fails on Java 7 for unrelated reasons.

PR #81, opened 2019-06-17
Performance Improvement: Call toArray with 0 Array Size

DaGeRe:
Since Java 6, calling toArray with an array of size is faster than calling it with the size of the array (https://shipilev.net/blog/2016/arrays-wisdom-ancients/). This is also a PMD recommendation: https://pmd.github.io/latest/pmd_rules_java_performance.html#optimizabletoarraycall 
I changed this in the code.


bodewig:
many thanks, @DaGeRe

PR #82, opened 2019-10-21
Made some variable names more consistent with the other parts.

euske:
Hello, we're developing an automated system that detects inconsistent variable names in a large software project. Our system checks if each variable name is consistent with other variables in the project in its usage pattern, and proposes correct candidates if inconsistency is detected. This is a part of academic research that we hope to publish soon, but as a part of the evaluation, we applied our systems to your projects and got a few interesting results. We carefully reviewed our system output and manually created a patch to correct a few variable names. We would be delighted if this patch is found to be useful. If you have a question or suggestion regarding this patch, we'd happily
answer. Thank you.

### DIFF
--- a/src/main/java/org/apache/commons/compress/archivers/zip/BitStream.java
+++ b/src/main/java/org/apache/commons/compress/archivers/zip/BitStream.java
@@ -48,14 +48,14 @@ class BitStream extends BitInputStream {
     /**
      * Returns the integer value formed by the n next bits (up to 8 bits).
      *
-     * @param n the number of bits read (up to 8)
-     * @return The value formed by the n bits, or -1 if the end of the stream has been reached
+     * @param count the number of bits read (up to 8)
+     * @return The value formed by the count bits, or -1 if the end of the stream has been reached
      */
-    long nextBits(final int n) throws IOException {
-        if (n < 0 || n > 8) {
-            throw new IOException("Trying to read " + n + " bits, at most 8 are allowed");
+    long nextBits(final int count) throws IOException {
+        if (count < 0 || count > 8) {
+            throw new IOException("Trying to read " + count + " bits, at most 8 are allowed");
         }
-        return readBits(n);
+        return readBits(count);
     }
 
     int nextByte() throws IOException {
--- a/src/main/java/org/apache/commons/compress/archivers/zip/StreamCompressor.java
+++ b/src/main/java/org/apache/commons/compress/archivers/zip/StreamCompressor.java
@@ -259,9 +259,9 @@ public abstract class StreamCompressor implements Closeable {
     }
 
     void deflate() throws IOException {
-        final int len = def.deflate(outputBuffer, 0, outputBuffer.length);
-        if (len > 0) {
-            writeCounted(outputBuffer, 0, len);
+        final int length = def.deflate(outputBuffer, 0, outputBuffer.length);
+        if (length > 0) {
+            writeCounted(outputBuffer, 0, length);
         }
     }
 
--- a/src/main/java/org/apache/commons/compress/compressors/z/ZCompressorInputStream.java
+++ b/src/main/java/org/apache/commons/compress/compressors/z/ZCompressorInputStream.java
@@ -103,12 +103,12 @@ public class ZCompressorInputStream extends LZWInputStream {
     @Override
     protected int addEntry(final int previousCode, final byte character) throws IOException {
         final int maxTableSize = 1 << getCodeSize();
-        final int r = addEntry(previousCode, character, maxTableSize);
+        final int idx = addEntry(previousCode, character, maxTableSize);
         if (getTableSize() == maxTableSize && getCodeSize() < maxCodeSize) {
             reAlignReading();
             incrementCodeSize();
         }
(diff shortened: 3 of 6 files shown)


src/main/java/org/apache/commons/compress/compressors/z/ZCompressorInputStream.java
```diff
@@ -103,12 +103,12 @@ private void reAlignReading() throws IOException {
...
     protected int addEntry(final int previousCode, final byte character) throws IOException {
         final int maxTableSize = 1 << getCodeSize();
-        final int r = addEntry(previousCode, character, maxTableSize);
+        final int idx = addEntry(previousCode, character, maxTableSize);
         if (getTableSize() == maxTableSize && getCodeSize() < maxCodeSize) {
             reAlignReading();
             incrementCodeSize();
         }
-        return r;
+        return idx;
```
garydgregory: I am -1 to abbreviated names like "idx". I prefer "index" or, as a fallback for the brevity fans, the classic "i".

src/main/java/org/apache/commons/compress/utils/ByteUtils.java
```diff
@@ -71,16 +71,16 @@ public static long fromLittleEndian(byte[] bytes) {
     /**
      * Reads the given byte array as a little endian long.
      * @param bytes the byte array to convert
-     * @param off the offset into the array that starts the value
+     * @param offset the offset into the array that starts the value
```
garydgregory: +1

src/main/java/org/apache/commons/compress/archivers/zip/StreamCompressor.java
```diff
@@ -259,9 +259,9 @@ private void deflateUntilInputIsNeeded() throws IOException {
     }
 
     void deflate() throws IOException {
-        final int len = def.deflate(outputBuffer, 0, outputBuffer.length);
-        if (len > 0) {
-            writeCounted(outputBuffer, 0, len);
+        final int length = def.deflate(outputBuffer, 0, outputBuffer.length);
```
garydgregory: +1

src/main/java/org/apache/commons/compress/archivers/zip/BitStream.java
```diff
@@ -48,14 +48,14 @@ int nextBit() throws IOException {
...
      *
-     * @param n the number of bits read (up to 8)
-     * @return The value formed by the n bits, or -1 if the end of the stream has been reached
+     * @param count the number of bits read (up to 8)
+     * @return The value formed by the count bits, or -1 if the end of the stream has been reached
      */
-    long nextBits(final int n) throws IOException {
-        if (n < 0 || n > 8) {
-            throw new IOException("Trying to read " + n + " bits, at most 8 are allowed");
+    long nextBits(final int count) throws IOException {
```
garydgregory: +1

bodewig:
I am on the fence with this PR.

Personally I don't see any benefit in consistent naming of local variables as long as the local variables have names that can be understood. To me a local variable named `length` is in no way clearer than one called `len`.

We will never get a consistency at the level your tool is striving for on a project that relies on numerous volunteer contributors. I don't see myself frustrating contributors by asking them to rename their local variables on patches.

@garydgregory if you see value in this and want to merge the PR after the changes you've requested have been made I won't stand in the way but I'll otherwise ignore it myself.

garydgregory:
I don't care that the changes come from a tool, I am just looking to see if the code is easier to read, which admittedly is subjective of course. But yeah, I can bring in the PR at some point. But if this is just a PR where the poster does not come back and update the code or comment, then it will go nowhere.

src/main/java/org/apache/commons/compress/compressors/z/ZCompressorInputStream.java
euske: This suggestion for "idx" comes from a similar expression in UnshrinkingInputStream.java:

    60:        final int idx = addEntry(previousCode, character, MAX_TABLE_SIZE);
    61:        if (idx >= 0) {
    62:            isUsed[idx] = true;
    63:        }
    64:        return idx;
 
The use of "idx" is slightly more common than "index" throughout the project. In this case, there's already a variable that has a similar use, so the system picked it up. Our system doesn't create a new name but it tries to choose the best looking candidate to make the code consistent with a minimal amount of change.

euske:
> Personally I don't see any benefit in consistent naming of local variables as long as the local variables have names that can be understood. To me a local variable named `length` is in no way clearer than one called `len`.
> 
> We will never get a consistency at the level your tool is striving for on a project that relies on numerous volunteer contributors. I don't see myself frustrating contributors by asking them to rename their local variables on patches.

I think that's a valid concern. Our system proposed numerous small changes, but I only sent some of the most conspicuous ones. I would say that overall the variable naming in this project is pretty consistent. We're not insisting the change.

garydgregory:
Please rebase on master. Recent changes should allow all builds to be green including Java 16 and 17-EA.

garydgregory:
I think the consensus is that is not really worth it. Closing.

PR #85, opened 2019-10-28
COMPRESS-497 Handle missing endheader offset

Skyr:
Detect incomplete header (all zero bytes)
Try to guess end header by trying to parse the file at the end


src/main/java/org/apache/commons/compress/archivers/sevenz/SevenZFile.java
```diff
@@ -441,17 +441,81 @@ private Archive readHeaders(final byte[] password) throws IOException {
...
+            }
+        } else {
+            headerLooksValid = true;
+        }
+
+        if (headerLooksValid) {
+            final StartHeader startHeader = readStartHeader(startHeaderCrc);
+            return initializeArchive(startHeader, password, true);
+        } else {
+            // No valid header found - probably first file of multipart archive was removed too early. Scan for end header.
```
bodewig: Could you please extract the block starting here into a new method?

bodewig: actually I can do this kind of cosmetic change myself,  :-)

PR #87, opened 2019-11-25
COMPRESS-124 : Add support for extracting sparse entries from tar archives

PeterAlfreadLee:
[COMPRESS-124](https://issues.apache.org/jira/browse/COMPRESS-124)
Add support for extracting sparse entries from tar archives, including :

1. Old GNU Format
The sparse map of old GNU is stored in tar header with only 4 sparse headers. If more spar headers exist, they should be stored in the extension block

2. PAX 0.0 Format
The sparse map of PAX 0.0 Format is stored in tar headers with `GNU.sparse.offset` and `GNU.sparse.numbytes`. They may appear more than 1 time, so they cannot be stored in a map.

3. PAX 0.1 Format
The sparse map of PAX 0.1 Format is stored in tar headers with `GNU.sparse.map`.

3. PAX 1.0 Format
The sparse map of PAX 1.0 Format is stored in the head of data file block. Therefore we need to handle it separately when we encounted a PAX 1.0 tar.

The implemention is mainly implemented by the TarArchiveSparseInputStream. It returns 0 when reading the "holes", and the file data from original tar file input stream when reading the sparse content. These are all decided by the sparse headers. The TarArchiveSparseInputStream reads all sparse headers, creates all-zero input streams surrounded by BoundedInputStream and actual tar file input stream surrounded by BoundedInputStream, and combines them together to a single TarArchiveSparseInputStream. Then we can straightly read from TarArchiveSparseInputStream when we encounted a sparse tar entry.

TODO : I noticed in the code that there are some kind of "star sparse data". I didn't find it in [GNU Tar Manual](https://www.gnu.org/software/tar/manual), so I don't know how to parse such kind of tars.

Please let me know if something need to be modified. :-) @bodewig


src/main/java/org/apache/commons/compress/archivers/tar/TarArchiveInputStream.java
```diff
@@ -243,8 +258,16 @@ public long skip(final long n) throws IOException {
             return 0;
         }
 
-        final long available = entrySize - entryOffset;
-        final long skipped = IOUtils.skip(is, Math.min(n, available));
+        long available;
+        long skipped;
+        if(!currEntry.isSparse()) {
```
tcurdt: This is creating a lot of duplicate code paths. Maybe the sparse handling could be just moved into get `getRealSize()` and we can keep the single code path?

src/main/java/org/apache/commons/compress/archivers/tar/TarArchiveSparseInputStream.java
```diff
@@ -0,0 +1,230 @@
...
+
+/**
+ * For sparse tar entries, there are many "holes"(consisting of all 0) in the file. Only the non-zero data is
+ * stored in tar files, and they are stored separately. The structure of non-zero data is introduced by the
+ * sparse headers using the offset, where a block of non-zero data starts, and numbytes, the length of the
+ * non-zero data block.
+ * This class is used to construct an input stream that combines the "holes" and the non-zero data together using
+ * the sparse headers. When reading from this input stream, the actual data is read out with "holes" and non-zero
+ * data combined together according to the sparse headers.
+ */
```
tcurdt: From the description I don't understand why this needs a separate class. Why no have a single `TarArchiveInputStream` that also handles support for sparse entries? Can you elaborate?

src/main/java/org/apache/commons/compress/archivers/tar/TarArchiveInputStream.java
PeterAlfreadLee: I agree. I will modify this and move the handling to the `getRealSize`.

src/main/java/org/apache/commons/compress/archivers/tar/TarArchiveSparseInputStream.java
PeterAlfreadLee: Of course I can make `TarArchiveInputStream` be able to handle sparse entries, therefore I need to modify the `read()` and `skip()` in `TarArchiveInputStream` when we are reading sparse entries.

The reason why I create a new class `TarArchiveSparseInputStream` is that I don't want to modify `TarArchiveInputStream` too much. I want to try my best to keep the `TarArchiveInputStream` as the original logic as much as possible.

With a separate `TarArchiveSparseInputStream`, I can replace the input stream to a new `TarArchiveSparseInputStream`. I don't need to modify the `read()` or `skip()` in `TarArchiveInputStream`.

PeterAlfreadLee: I'm not sure if creating a single `TarArchiveSparseInputStream` is a good idea or not. Maybe we can discuss more about this.

src/main/java/org/apache/commons/compress/archivers/tar/TarArchiveStructSparse.java
```diff
@@ -0,0 +1,57 @@
...
+ * <pre>
+ * struct sparse {
+ * char offset[12];   // offset 0
+ * char numbytes[12]; // offset 12
+ * };
+ * </pre>
+ */
+public class TarArchiveStructSparse {
+    private long offset;
+
```
melloware: Should this struct implement equals() hashcode() and toString() to be a good citizen?

src/main/java/org/apache/commons/compress/archivers/tar/TarArchiveInputStream.java
```diff
@@ -80,6 +85,9 @@
     // the global PAX header
     private Map<String, String> globalPaxHeaders = new HashMap<>();
 
+    // the global sparse headers, this is only used in PAX Format 0.X
+    private List<TarArchiveStructSparse> globalSparseHeaders = new ArrayList<>();
+
```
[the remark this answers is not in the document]
PeterAlfreadLee: Agree. But the HashMap was not created in this PR. I will mark the List as final.

src/main/java/org/apache/commons/compress/archivers/tar/TarArchiveSparseInputStream.java
```diff
@@ -0,0 +1,230 @@
...
+public class TarArchiveSparseInputStream extends InputStream {
+    /** the sparse headers describing the sparse information */
+    private List<TarArchiveStructSparse> sparseHeaders;
+
+    /** the input stream of the tar file */
+    private InputStream inputStream;
```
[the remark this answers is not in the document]
PeterAlfreadLee: Agree.

src/main/java/org/apache/commons/compress/archivers/tar/TarArchiveSparseInputStream.java
```diff
@@ -0,0 +1,230 @@
...
+ * stored in tar files, and they are stored separately. The structure of non-zero data is introduced by the
+ * sparse headers using the offset, where a block of non-zero data starts, and numbytes, the length of the
+ * non-zero data block.
+ * This class is used to construct an input stream that combines the "holes" and the non-zero data together using
+ * the sparse headers. When reading from this input stream, the actual data is read out with "holes" and non-zero
+ * data combined together according to the sparse headers.
+ */
+public class TarArchiveSparseInputStream extends InputStream {
+    /** the sparse headers describing the sparse information */
+    private List<TarArchiveStructSparse> sparseHeaders;
```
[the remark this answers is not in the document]
PeterAlfreadLee: Agree.

src/main/java/org/apache/commons/compress/archivers/tar/TarArchiveStructSparse.java
PeterAlfreadLee: Seems a lot of data classes in commons-compress are lack of `equals()`, `hashcode()` and `toString()`, but I agree that implementing these methods in data class is a better programming practice. I will add these methods.

PeterAlfreadLee:
Code just pushed basing on reviews of @tcurdt and @melloware .
Exception for one review opinion :
Is creating a separate class `TarArchiveSparseInputStream` a good idea or not?

Of course I can make `TarArchiveInputStream` be able to handle sparse entries, therefore I need to modify the `read()` and `skip()` in TarArchiveInputStream when we are reading sparse entries.

The reason why I create a new class `TarArchiveSparseInputStream` is that I don't want to modify `TarArchiveInputStream` too much. I want to try my best to keep the `TarArchiveInputStream` as the original logic as much as possible.

With a separate `TarArchiveSparseInputStream`, I can replace the input stream to a new `TarArchiveSparseInputStream`. I don't need to modify the `read()` or `skip()` in `TarArchiveInputStream`.

I'm not sure if creating a single `TarArchiveSparseInputStream` is a good idea or not.

What do you think?

tcurdt:
While I appreciate being careful, I am not a fan of duplicating functionality and interfaces like that. So I am in favor of just one class. Maybe it means that testing needs to improved to be sure all is kept working.

Since @bodewig has the best overview over the project these days I think it would be good to get input from him first.

bodewig:
I'll carve out some time at the weekend. Sorry, haven't got that much time for OSS right now.

bodewig:
First of all, many thanks @PeterAlfreadLee 

I agree with @tcurdt that providing the functionality via a single `TarArchiveInputStream` would be preferable. Maybe we could extract the meet of the read and skip methods into strategy implementations and make the handling of sparse entries another strategy that is triggered when reading said type of entry? That way we could extend the class without making it even more complex itself.

Here I'm just thinking loud, I haven't got any code to show or even just tried to sketch it up. I must admit I haven't even looked at the code in detail, so maybe the idea is complete nonsense.

BTW, star is a pretty old tar implementation by Jörg Schilling that used to be pretty popular before POSIX started to add support for many of the shortcoming that GNU tar addressed in its own way - long file names, big files, non-ASCII file names and all that. It is still available via Sourceforge, I found the manual page at http://schilytools.sourceforge.net/man/man1/star.1.html . IIRC it is BSD licensed so was preferred over GNU tar by the BSDs and some other Unices.

PeterAlfreadLee:
@bodewig Sure. I'll modify my code.
The `TarArchiveInputStream` could have another handling of `read` and `skip` when the entry is a sparse entry. This could easily implemented by calling `currEntry.isSparse()`. Then I can expand the original code with only a little modification of original code. Therefore I don't need a separate class `TarArchiveSparseInputStream`.
I'll push my code in these two days.

For the implemenation of reading the sparse entry, I'm combining the `BoundedInputStream` with the original tar file input stream and the `TarArchiveSparseZeroInputStream`. They are constructed by reading the sparse headers of `currEntry`. What do you think about this? Maybe you have some better ideas? @bodewig @tcurdt

PeterAlfreadLee:
Code pushed.
The `TarArchiveInputStream` can deal with the sparse entries now. There's no spearate class that deals with the sparse entries.

src/main/java/org/apache/commons/compress/archivers/tar/TarArchiveInputStream.java
```diff
@@ -516,6 +713,22 @@ private void paxHeaders() throws IOException{
                                 final String value = new String(rest, 0,
                                                           restLen - 1, CharsetNames.UTF_8);
                                 headers.put(keyword, value);
+
+                                // for 0.0 PAX Headers
+                                if(keyword.equals("GNU.sparse.offset")) {
+                                    sparseHeader = new TarArchiveStructSparse(Long.parseLong(value), 0);
```
tcurdt: better extract the keys into a static String

src/main/java/org/apache/commons/compress/archivers/tar/TarArchiveInputStream.java
```diff
@@ -26,14 +26,14 @@
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
-import java.util.HashMap;
-import java.util.Map;
+import java.util.*;
```
tcurdt: unless rules have changed - please avoid * imports

tcurdt:
I didn't do a thorough code review but from the diffs it looks OKish. Still not sure about some of the new classes. Either way - thanks for all the work!

It would probably be good to check this out and look at the final code as this is a big change.

I am wondering if our test cases are enough of a safety net.

src/main/java/org/apache/commons/compress/archivers/tar/TarArchiveInputStream.java
PeterAlfreadLee: Agree. There're some other keys exist in `TarArchiveEntry`, like this:

https://github.com/apache/commons-compress/blob/master/src/main/java/org/apache/commons/compress/archivers/tar/TarArchiveEntry.java#L1126-L1129

I'm wondering if a refactor is needed or not for the existing code.

src/main/java/org/apache/commons/compress/archivers/tar/TarArchiveInputStream.java
PeterAlfreadLee: Agree. Sorry for my careless.

PeterAlfreadLee:
> I am wondering if our test cases are enough of a safety net.

I agree with this. I will try to add more test cases.

src/main/java/org/apache/commons/compress/archivers/tar/TarArchiveInputStream.java
bodewig: resolved with b465bfa6

bodewig:
@PeterAlfreadLee commit b465bfa contains some minor changes

I've created the site build for the current master branch and published it at https://stefan.samaflost.de/staging/commons-compress-1.20-snapshot/ - looking at the Jacoco results, for example https://stefan.samaflost.de/staging/commons-compress-1.20-snapshot/jacoco/org.apache.commons.compress.archivers.tar/TarArchiveInputStream.java.html we seem to be covering the new sparse code pretty well. Of course more test cases won't hurt.

I'm wondering whether we really want to make the API details public or whether things like the getters and setters for sparse headers in `TarArchiveEntry` should be hidden instead. As a matter of taste I would have preferred `TarArchiveStructSparse` to be immutable and have no setters at all.

PeterAlfreadLee:
> I'm wondering whether we really want to make the API details public or whether things like the getters and setters for sparse headers in TarArchiveEntry should be hidden instead. As a matter of taste I would have preferred TarArchiveStructSparse to be immutable and have no setters at all.

+1

For `TarArchiveEntry`, I have the same feeling that the details(getters/setters) should be hidden when I was reading the code.

For `TarArchiveStructSparse`, I agree that it should be immutable. I will modify this and push another PR.

PR #88, opened 2019-12-19
COMPRESS-499: Fixed edge case in `SeekableInMemoryByteChannel#truncate`

trejkaz:
Initial proposed patch.

It's unknown whether the bug has any actual impact because the edge case is very specific and honestly is a misuse-case in my book.

Test cases cover what we deemed to be all possible size vs new size vs position relationships.

### DIFF
--- a/src/main/java/org/apache/commons/compress/utils/SeekableInMemoryByteChannel.java
+++ b/src/main/java/org/apache/commons/compress/utils/SeekableInMemoryByteChannel.java
@@ -96,9 +96,7 @@ public class SeekableInMemoryByteChannel implements SeekableByteChannel {
 
     @Override
     public SeekableByteChannel truncate(long newSize) {
-        if (size > newSize) {
-            size = (int) newSize;
-        }
+        size = (int) newSize;
         repositionIfNecessary();
         return this;
     }
--- a/src/test/java/org/apache/commons/compress/utils/SeekableInMemoryByteChannelTest.java
+++ b/src/test/java/org/apache/commons/compress/utils/SeekableInMemoryByteChannelTest.java
@@ -155,15 +155,97 @@ public class SeekableInMemoryByteChannelTest {
 
     @Test
     public void shouldSetProperPositionOnTruncate() throws IOException {
-        //given
-        SeekableInMemoryByteChannel c = new SeekableInMemoryByteChannel(testData);
-        //when
-        c.position(testData.length);
-        c.truncate(4L);
-        //then
-        assertEquals(4L, c.position());
-        assertEquals(4L, c.size());
-        c.close();
+        // In absence of proper parametrised test cases :(
+        long[][] testCases = {
+                // { Position, NewSize, ExpectedPosition }
+
+                // Position == Size == NewSize
+                // - Size doesn?t change
+                // - Position doesn?t change
+                { 9L, 9L, 9L },
+
+                // NewSize < Position == Size :
+                // - Size changes to NewSize
+                // - Position changes to NewSize
+                { 4L, 9L, 4L },
+
+                // Position == Size < NewSize :
+                // - Size changes to NewSize and you get undefined data between
+                // - Position doesn?t change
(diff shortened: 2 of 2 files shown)


src/main/java/org/apache/commons/compress/utils/SeekableInMemoryByteChannel.java
```diff
@@ -96,9 +96,7 @@ public long size() {
 
     @Override
     public SeekableByteChannel truncate(long newSize) {
-        if (size > newSize) {
-            size = (int) newSize;
-        }
+        size = (int) newSize;
```
trejkaz: The fact that removing the condition made the last test case pass was very interesting but definitely warrants double or even triple-checking. Also I'd check all the test cases again too. Two of us here looked at them for quite a while but I found one mistake in our reasoning while implementing the actual test code already. There might be another.

PeterAlfreadLee: I think removing the size comparion is not right. There're also some problems in your test cases, and I think that's why your code passed the test cases.

src/test/java/org/apache/commons/compress/utils/SeekableInMemoryByteChannelTest.java
```diff
@@ -155,15 +155,97 @@ public void shouldTruncateContentsProperly() throws IOException {
...
+                long expectedPosition = testCase[2];
+
+                //given
+                try (SeekableInMemoryByteChannel c = new SeekableInMemoryByteChannel(testData)) {
+                    //when
+                    c.position(position);
+                    c.truncate(newSize);
+                    //then
+                    assertEquals(expectedPosition, c.position());
+                    assertEquals(newSize, c.size());
```
PeterAlfreadLee: I think this is not right. Basing on the java docs, I think it should be like this :
```
//when
long originalSize = c.size();
c.position(position);
c.truncate(newSize);
//then
assertEquals(expectedPosition, c.position());
long expectedSize = newSize < originalSize ? newSize : originalSize;
assertEquals(expectedSize, c.size());
```

PeterAlfreadLee:
As I said on JIRA, I think we should wait a while until we have figure out if `SeekableInMemoryByteChannel` is doing this intentionally or not.
I would say it may be better that we should be patient and have a discussion in mailing list before the code is merged.

Besides, I think it's better to have a modification on `SeekableInMemoryByteChannel.repositionIfNecessary` if this is identified as a problem. The problem is about the `position` in `SeekableInMemoryByteChannel`, not the `size`. `SeekableInMemoryByteChannel.repositionIfNecessary` is the method that actually changes the `position`, and it's also called in `SeekableInMemoryByteChannel.read`.

And I think a double check on other classes that implements `SeekableByteChannel` would be a better software practice - we shoud check other related potential problems - but not too rush, I believe a discussion before it may be better.

Hope this helps :)

src/test/java/org/apache/commons/compress/utils/SeekableInMemoryByteChannelTest.java
trejkaz: You might be right about that. If that's the case I'd prefer to put expectedSize into the array though, otherwise it's dangerously like replicating the logic in the class being tested.

PeterAlfreadLee: > if the position is past the end and the truncation point, then the position is updated to the truncation point, but the size is not updated to it.

I don't see any problems with the `size`. I think that's what the name `truncate` means. It's called `truncate`, not `resize` or some other name, which means the size can only become smaller when this method is called. If the `newSize` is bigger than the `size`, I think there's nothing needed to do on the `size`.
I think the problem is about the `position`, not the `size`.

bodewig:
My reading of the spec is that the current behavior for size was correct but not for position. I've added a bunch of tests and think I've fixed the issue with af63df69

PR #91, opened 2020-01-20
Make bz2 decompressor robust to selector overflow

jobar:
As explained in https://gnu.wildebeest.org/blog/mjw/2019/08/02/bzip2-and-the-cve-that-wasnt/
bz2 has historically been able to decompress wrongly formatted compressed
files, more precisely files stating more selectors that the algorithm
actually uses.
This patch replicates https://sourceware.org/ml/bzip2-devel/2019-q3/msg00007.html
and provides a test file extracted from 20191201 German Xikipedia XML dump:
https://dumps.wikimedia.org/dewiki/20191201/dewiki-20191201-pages-meta-history5.xml-p8161946p8319641.bz2
(the link will disappear soon, as new dumps are generated).


src/test/java/org/apache/commons/compress/compressors/bzip2/BZip2NSelectorsOverflowTest.java
```diff
@@ -0,0 +1,49 @@
...
+
+public class BZip2NSelectorsOverflowTest extends AbstractTestCase {
+
+    /**
+     * See https://sourceware.org/ml/bzip2-devel/2019-q3/msg00007.html
+     */
+    @Test()
+    public void shouldDecompressBlockWithNSelectorOverflow() throws Exception {
+        File decompressed = getFile("rec02841dewiki-20191201-pages-meta-history5.xml-p8161946p8319641.bz2");
+        final File toDecompress = getFile("rec02841dewiki-20191201-pages-meta-history5.xml-p8161946p8319641.bz2");
```
PeterAlfreadLee: Why the decompressed file and toDecompress file are the same file?
Maybe we should compare the decompressed file with src/test/resources/rec02841dewiki-20191201-pages-meta-history5.xml-p8161946p8319641?

bodewig:
Many thanks @jobar 

The test fails on my Linux box and seems to fail on Travis as well, Could this be a line-separator OS-difference issue on the XML file? If so, it might be better to pick a different testcase - if possible - anyway as [it's licesne(https://dumps.wikimedia.org/legal.html) may force us to jump through a few extra hoops, see https://apache.org/legal/resolved.html#cc-sa

One more thing, to me it looks as if the `currentWorkingBlock` variable you introduced was never read and can safely be removed.

jobar:
Thank you for the reviews @bodewig and @PeterAlfreadLee. I'm sorry I had some erroneous leftovers :(
I have removed the unused variable and corrected the test so that it uses the same test file as lbzip2 and actually do some testing ( the previous version was really not correct, sorry again).
I hope using the lbzip2 test will be ok from a license perspective, I'm not super good at understanding licences :S)

bodewig:
Thank you, I'll need to double check on the license myself. :-)

bodewig:
many thanks @jobar I've merged this, it will be incorparated in 1.20. I've made a few small changes and listed you as contributor in our changelog.

jobar:
Thanks @bodewig for the merge and updates :) I'm glad to have this fixed upstream.

PR #92, opened 2020-02-05
COMPRESS-503 : "open when actually needed" for MultiReadOnlySeekableByteChannel

PeterAlfredLee:
Please refer (#90)[https://github.com/apache/commons-compress/pull/90]

When I was adding zip64 support for split zip, I encountered a problem : 
When adding testcases in `Zip64SupportIT`, I created a split zip with 10,000+ split segments. Then I found that I was unable to unzip it because there would be too many open files when extracting it. We can oepn the files when actually needed and therefore we can successfully extract such split zips with great amount of segments.

I have set a threshold of 20 in `MultiReadOnlySeekableByteChannel`. The "open when actually needed" procedure will only work when the number of split segments is greater than the threshold.

Actually this is a pretty rare case cause most split zips would not have too many segments. So you can decide whether to merge this PR or not. :-)


bodewig:
@PeterAlfreadLee I think it would be good to track this as a separate JIRA issue.

PeterAlfreadLee:
Sure. I will add a new JIRA issue and modify the name of this PR.

PR #94, opened 2020-03-04
Deprecated constructors of ZstdOutputStream fix

PeterAlfreadLee:
The zstd-jni was updated to 1.4.4-7, and some constructors of ZstdOutputStream were deprecated.

This PR fix them as the comment was recommended :
/**
*  @deprecated
*  Use ZstdOutputStream() or ZstdOutputStream(level) and set the other params with the setters
**/


bodewig:
many thanks

PR #100, opened 2020-05-13
make openjdk14 and openjdk-ea allow_failures in travis.

XenoAmess:


bodewig:
thanks @XenoAmess 

why did you also enable the openjdk-ea build with this commit?

XenoAmess:
> thanks @XenoAmess
> 
> why did you also enable the openjdk-ea build with this commit?

Because I see openjdk-ea commented in original travis.yml
If you really want not to support openjdk-ea, then you will delete the commented part.
So I just assume you want the result of openjdk-ea but hesitate it will block the build.
So I thought we might also build openjdk-ea with allow_failures.

Of course we can also chose to delete the 
```
#    - dist: trusty
#      jdk: openjdk-ea
```
to make sure we really do not need result of openjdk-ea.

bodewig:
ah, my bad, I overlooked you also added it under allow-failures. thanks!

PR #101, opened 2020-05-16
COMPRESS-520: Implement ZCompressorInputStreamTest without powermock

theobisproject:
Noticed this test is skipped while checking the unit tests log. The mockito dependency was transitive before.
Is there a Jira ticket needed for a change like this?


PeterAlfredLee:
> Is there a Jira ticket needed for a change like this?

That's exactly what I'm thinking about. A change in pom should be tracked with a new Jira ticket.

theobisproject:
I've created the issue and added the number to the commit and the PR description.

bodewig:
Many thanks @theobisproject 

Actually it is nice you have created an issue but we don't require a Jira issue for every PR. Personally I prefer Jira if something needs discussion and skip it when talking about a straight forward PR. YMMV

PR #102, opened 2020-05-19
Compress-515: Add fileName/path info to ZipFile constructor IOException

ian-lavallee:


ian-lavallee:
@elharo

src/main/java/org/apache/commons/compress/archivers/zip/ZipFile.java
```diff
@@ -374,6 +374,8 @@ private ZipFile(final SeekableByteChannel channel, final String archiveName,
             }
             fillNameMap();
             success = true;
+        } catch (IOException e) {
```
elharo, line 377: Does archiveName here include the path?

src/main/java/org/apache/commons/compress/archivers/zip/ZipFile.java
```diff
@@ -374,6 +374,8 @@ private ZipFile(final SeekableByteChannel channel, final String archiveName,
...
+            throw new IOException("Error on ZipFile " + archiveName + e.getMessage(), e);
```
elharo, line 378: add "; " before `+ e.getMessage()` or drop the message completely.

src/main/java/org/apache/commons/compress/archivers/zip/ZipFile.java
ian-lavallee: If it used a constructor that gets the path sent into then yes, otherwise it will say "unknown archive"

src/main/java/org/apache/commons/compress/archivers/zip/ZipFile.java
ian-lavallee: dropped the message

ian-lavallee:
@bodewig

PR #103, opened 2020-05-26
refine travis-ci scripts

XenoAmess:


PeterAlfredLee:
The `$HOME/.m2` has already been cached in [pom.xml](https://github.com/apache/commons-compress/blob/master/.travis.yml#L50-L51).
Why do we need another one?

XenoAmess:
> The `$HOME/.m2` has already been cached in [pom.xml](https://github.com/apache/commons-compress/blob/master/.travis.yml#L50-L51).
> Why do we need another one?
@PeterAlfredLee oh sorry for the mistake.

PR #104, opened 2020-06-02
[COMPRESS-533] Refine javadoc and fix typos

XenoAmess:
Refine javadoc and fix typos

### DIFF
--- a/src/main/java/org/apache/commons/compress/compressors/brotli/BrotliUtils.java
+++ b/src/main/java/org/apache/commons/compress/compressors/brotli/BrotliUtils.java
@@ -36,7 +36,7 @@ public class BrotliUtils {
         try {
             Class.forName("org.osgi.framework.BundleEvent");
         } catch (final Exception ex) { // NOSONAR
-            setCacheBrotliAvailablity(true);
+            setCacheBrotliAvailability(true);
         }
     }
 
@@ -72,7 +72,7 @@ public class BrotliUtils {
      * <p>This defaults to {@code false} in an OSGi environment and {@code true} otherwise.</p>
      * @param doCache whether to cache the result
      */
-    public static void setCacheBrotliAvailablity(final boolean doCache) {
+    public static void setCacheBrotliAvailability(final boolean doCache) {
         if (!doCache) {
             cachedBrotliAvailability = CachedAvailability.DONT_CACHE;
         } else if (cachedBrotliAvailability == CachedAvailability.DONT_CACHE) {
--- a/src/main/java/org/apache/commons/compress/compressors/brotli/BrotliUtils.java
+++ b/src/main/java/org/apache/commons/compress/compressors/brotli/BrotliUtils.java
@@ -36,7 +36,7 @@ public class BrotliUtils {
         try {
             Class.forName("org.osgi.framework.BundleEvent");
         } catch (final Exception ex) { // NOSONAR
-            setCacheBrotliAvailability(true);
+            setCacheBrotliAvailablity(true);
         }
     }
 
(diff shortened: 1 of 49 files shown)


src/main/java/org/apache/commons/compress/compressors/lzma/LZMAUtils.java
```diff
@@ -146,7 +146,7 @@ public static String getCompressedFilename(final String fileName) {
      * <p>This defaults to {@code false} in an OSGi environment and {@code true} otherwise.</p>
      * @param doCache whether to cache the result
      */
-    public static void setCacheLZMAAvailablity(final boolean doCache) {
```
bodewig, line 149: Even if its spelling is wrong, we cannot change the name of a public method as it would break backwards compatibility.

src/main/java/org/apache/commons/compress/compressors/xz/XZUtils.java
```diff
@@ -163,7 +163,7 @@ public static String getCompressedFilename(final String fileName) {
      * @param doCache whether to cache the result
      * @since 1.9
      */
-    public static void setCacheXZAvailablity(final boolean doCache) {
+    public static void setCacheXZAvailability(final boolean doCache) {
```
bodewig, line 166: same as above, this would break backwards compatibility.

src/main/java/org/apache/commons/compress/compressors/zstandard/ZstdUtils.java
```diff
@@ -85,7 +85,7 @@ private static boolean internalIsZstdCompressionAvailable() {
      * <p>This defaults to {@code false} in an OSGi environment and {@code true} otherwise.</p>
      * @param doCache whether to cache the result
      */
-    public static void setCacheZstdAvailablity(final boolean doCache) {
+    public static void setCacheZstdAvailability(final boolean doCache) {
```
bodewig, line 88: and again, we will have to live with a misspelled method name.

src/main/java/org/apache/commons/compress/compressors/brotli/BrotliUtils.java
```diff
@@ -72,7 +72,7 @@ private static boolean internalIsBrotliCompressionAvailable() {
      * <p>This defaults to {@code false} in an OSGi environment and {@code true} otherwise.</p>
      * @param doCache whether to cache the result
      */
-    public static void setCacheBrotliAvailablity(final boolean doCache) {
+    public static void setCacheBrotliAvailability(final boolean doCache) {
```
bodewig, line 75: this change breaks backwards compatibility,

XenoAmess:
@bodewig 
done.
I also have a question.
Do `archiveName` and `archiverName` means one same thing in this repo?
If so, should we consider uniform them to `archiveName` or `archiverName`?

Asked because I found both several `archiveName` and several `archiverName`, and I got kindof confused.

src/main/java/org/apache/commons/compress/compressors/xz/XZUtils.java
XenoAmess: > same as above, this would break backwards compatibility.

done.

src/main/java/org/apache/commons/compress/compressors/zstandard/ZstdUtils.java
XenoAmess: > and again, we will have to live with a misspelled method name.

done.

src/main/java/org/apache/commons/compress/compressors/brotli/BrotliUtils.java
XenoAmess: > this change breaks backwards compatibility,

done.

src/main/java/org/apache/commons/compress/compressors/lzma/LZMAUtils.java
XenoAmess: > Even if its spelling is wrong, we cannot change the name of a public method as it would break backwards compatibility.

done.

bodewig:
I'm not sure we've got an agreed upon language at all :-)
In this particular case it may depend on the context. `archiveName` would be the file name of an archive, `archiverName` represent the archive format tar/zip/7z...

XenoAmess:
> I'm not sure we've got an agreed upon language at all :-)
> In this particular case it may depend on the context. `archiveName` would be the file name of an archive, `archiverName` represent the archive format tar/zip/7z...

Ok, I get it now, thx.

PR #105, opened 2020-06-03
[COMPRESS-534] use StandardCharsets.

XenoAmess:
In most cases using StandardCharsets is faster.
You can get more information at `decode(String charsetName, byte[] ba, int off, int len)` and `decode(Charset cs, byte[] ba, int off, int len)` in class java.lang.StringCoding


bodewig:
Many thanks @XenoAmess

PR #109, opened 2020-06-04
unitized array styles.

XenoAmess:
from both `byte a[]` and `byte[] a` to only `byte[] a`


PeterAlfredLee:
LGTM, and thank you. @XenoAmess

PR #110, opened 2020-06-04
use System.currentTimeMillis instead of new Date().getTime

XenoAmess:
to avoid Date() object creation thus faster.


PeterAlfredLee:
LGTM. Thank you.

PR #112, opened 2020-06-04
remove `== null` as it is covered by `instance of`

XenoAmess:
Or, is it for special optimization strategy?

### DIFF
--- a/src/main/java/org/apache/commons/compress/archivers/zip/ZipEightByteInteger.java
+++ b/src/main/java/org/apache/commons/compress/archivers/zip/ZipEightByteInteger.java
@@ -212,7 +212,7 @@ public final class ZipEightByteInteger implements Serializable {
      */
     @Override
     public boolean equals(final Object o) {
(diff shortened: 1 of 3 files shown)


PeterAlfredLee:
Not sure if this is caused by Sonarqube or some other scans. From my view the null check is not needed here.

bodewig:
@PeterAlfredLee are you talking about the decreased coverage? This happens because the builds on different JDKs end with slightly different results and so the coverage fluctuates a bit depending on which JDK's build finishes first.

PeterAlfredLee:
> are you talking about the decreased coverage

Nope, I was just wondering if the null check here is caused by some type of static analysis like FindBugs or Sonarqube. This PR looks good to me - if it's not because of some static analysis tool.

> This happens because the builds on different JDKs end with slightly different results and so the coverage fluctuates a bit depending on which JDK's build finishes first.

So that's the reason why the coverage varies when no code is changed. Thank you for your explanation.

bodewig:
thanks @XenoAmess

PR #113, opened 2020-07-07
COMPRESS-540: Implement TarFile to allow random access to tar files

theobisproject:
Basic idea behind the implementation is to read all headers at instantiation and save the position to the entry data. This will then be accessed as a random access file when reading the entry.
I tried to touch as less exisiting code as possible and only did refactor existing code were I felt it was really needed. 

I have the following TODOs left in the code where I think your opinion is needed

- TarArchiveInputStream#canReadEntryData: as noted in the TODO sparse entries are now supported. I see no need for this function any longer and would not include it in the TarFile api. Also we should update the implementation to reflect the now supported sparse entries.
- In the SparseFilesTest extracting of sparsefile-0.1 in pax_gnu_sparse.tar is skipped because of problems with tar on Ubuntu 16.04. I tested this on Ubuntu 18.04 and everything worked fine. Should we reenable extracting of the file?

- [x] Incorporate fix for COMPRESS-544


garydgregory:
From the sidelines here... I am wondering if the introduction of utility streams would not best be done by depending on Apache Commons IO which already contains a bounded stream for example. What's missing in Commons IO could be added there if general enough.

XenoAmess:
Hi.
suggestions:
for BoundedSeekableByteChannelInputStream, please tell users if it is thread-safe in its javadoc.
for BoundedNIOInputStream , please tell users if it is thread-safe in its javadoc.
for BoundedNIOInputStream , please tell users if it is thread-safe in its javadoc.
for TarArchiveSparseZeroInputStream, maybe this class should be added to commons-io instead.

theobisproject:
> I am wondering if the introduction of utility streams would not best be done by depending on Apache Commons IO which already contains a bounded stream for example

As I understood commons-compress does purposefully not depend on Commons IO. This is something which I cannot decide to change.

> for BoundedSeekableByteChannelInputStream, please tell users if it is thread-safe in its javadoc.
for BoundedNIOInputStream , please tell users if it is thread-safe in its javadoc.

I will add the information if the general design is accepted from the maintainers.

> for TarArchiveSparseZeroInputStream, maybe this class should be added to commons-io instead.

In my opinion this adds no value for Commons IO and should not be used outside of the tar implementation. I only moved this class to reuse it in both implementations and made it package protected so nobody outside the tar package will use it.

PeterAlfredLee:
> In my opinion this adds no value for Commons IO and should not be used outside of the tar implementation.

+1. The `TarArchiveSparseZeroInputStream` will only return 0 if you are reading it. It's simple enough and I think this should be limited to tar in Commons Compress.

src/main/java/org/apache/commons/compress/archivers/tar/TarFile.java
```diff
@@ -0,0 +1,712 @@
...
+import org.apache.commons.compress.utils.BoundedInputStream;
+import org.apache.commons.compress.utils.BoundedNIOInputStream;
+import org.apache.commons.compress.utils.BoundedSeekableByteChannelInputStream;
+import org.apache.commons.compress.utils.SeekableInMemoryByteChannel;
+
+/**
+ * The TarFile provides random access to UNIX to archives.
+ * @since 1.21
+ */
+public class TarFile implements Closeable {
```
PeterAlfredLee, line 49: Many methods in `TarFile` looks pretty like the ones in `TarArchiveInputStream`(e.g. `buildSparseInputStreams`, `getLongNameData`,  `buildSparseInputStreams`). Maybe we can find some way to share these methods?

src/main/java/org/apache/commons/compress/archivers/tar/TarArchiveInputStream.java
```diff
@@ -1106,35 +917,8 @@ public int compare(final TarArchiveStructSparse p, final TarArchiveStructSparse
...
+        if (!sparseInputStreams.isEmpty()) {
             currentSparseInputStreamIndex = 0;
         }
     }
-
-    /**
-     * This is an inputstream that always return 0,
-     * this is used when reading the "holes" of a sparse file
-     */
-    private static class TarArchiveSparseZeroInputStream extends InputStream {
```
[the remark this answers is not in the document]
theobisproject: I moved this class out of the `TarArchiveInputStream` because I also needed it in the `TarFile`. Otherwise I would had to introduce more duplicated code. I made the class package private to prevent accidental usage.
In my opinion before the changes done here it was reasonable to have this as inner class as this was only used here.

src/main/java/org/apache/commons/compress/utils/BoundedSeekableByteChannelInputStream.java
```diff
@@ -0,0 +1,56 @@
...
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.nio.channels.SeekableByteChannel;
+
+/**
+ * InputStream that delegates requests to the underlying SeekableByteChannel, making sure that only bytes from a certain
+ * range can be read.
+ * @since 1.21
+ */
+public class BoundedSeekableByteChannelInputStream extends BoundedNIOInputStream {
```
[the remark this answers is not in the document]
theobisproject: You are correct about the origin of the code. I moved it out of the `ZipFile` class because it is needed for reading sparse tar conent (see `TarFile` line 377)

src/main/java/org/apache/commons/compress/utils/BoundedNIOInputStream.java
```diff
@@ -0,0 +1,97 @@
...
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.ByteBuffer;
+
+/**
+ * NIO backed bounded input stream for reading a predefined amount of data from.
+ * @since 1.21
+ */
+public abstract class BoundedNIOInputStream extends InputStream {
```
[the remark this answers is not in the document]
theobisproject: I agree that the name of the class it not perfect. Just wan't to share the reason why I took the name.

The newly defined read method reads the content into an `java.nio.ByteBuffer` and the `read` methods of the `InputStream` are calling it. So the name comes more from the fact that an inherited class only needs to implement the new `read` method to the `ByteBuffer` and not from the actual underlying implementation. My experience shows that reading into a ByteBuffer from a non NIO source is not so much fun and therefore I put the `NIO` in the classname.

If someone has a better suggestion for a classname it would be welcome.

src/main/java/org/apache/commons/compress/archivers/tar/TarFile.java
theobisproject: The duplication is really on purpose from my side in the current state of the PR. I didn't wanted to limit myself to early from making the needed changes to get this work. Also I wanted to limit changes to existing code to avoid breaking it. Sharing some of the code can be quite a challenge because the data source is so different.
`buildSparseInputStream`: The difference here is that the sparse streams are saved differently in the TarFile than in the Stream. In the TarFile the streams need to be saved on a per entry basis to allow random access. I think this could also be done in the stream but for the first version I didn't wanted to change too much existing code.
`getLongNameData`: The only difference is the source from where the data is read (InputStream vs. SeekableByteChannel). A problem could be the call to `getNext(Tar)Entry` which has a lot of sideeffects.

bodewig:
Unfortunately I've been unable to do a real review so far. $life is just setting different priorities right now. A few more or less unsorted thoughts:

* I really wish Compress had abstractions that weren't so tightly bound to Java Streams, but we'll likely never find the energy and time to revamp it
* I understand you want to get feedback on whether something like `TarFile` would be accepted rater than the concrete implementation. IMHO this would be a nice addition even though I don't expect many people to require random access to a raw tar archive. Most people likely deal with tar archives that have been compressed and so you'd need to decompress it to a temporary file (or memory) for random access.
* we already have a bunch of things copied from commons-io and our very own BoundedInputStream may even predate the one of commons-io. I wouldn't want to add a new dependency to 1.x unless the reason was better than avoiding dupkiicating a few lines of code.
* have you considered making `TarArchiveEntry` implement `EntryStreamOffsets `?
* I'd be in favour of moving stream implementations to the utils package, in particular if you copy code from the zip package
* nitpicking: the methods you've moved to `TarUtils` probably don't need to be public.

theobisproject:
I know you are working on this library in your free time. So I did expect this big change to take some time before it is integrated.

>   IMHO this would be a nice addition even though I don't expect many people to require random access to a raw tar archive. Most people likely deal with tar archives that have been compressed and so you'd need to decompress it to a temporary file (or memory) for random access.

As you pointed out this only brings better speed for some special cases e.g. random access or somthing like the `TarLister` where no content is read and therefore the stream needs to read and discard all data.  
Specifially for compressed archives it can even take longer depending on the target where the file is decompressed to.

>  have you considered making `TarArchiveEntry` implement `EntryStreamOffsets `?

Seems like I missed this interface. Since this exactly what I wanted to implement I will change the implementation to implement the `EntryStreamOffsets` interface.

> I'd be in favour of moving stream implementations to the utils package, in particular if you copy code from the zip package

The code I moved out of the `ZipFile` class is already located in the utils package. Only the Tar specific streams I didn't create in the utils package because in my opinion nobody outside the library should access them and currently no other archive needs similar streams.

> nitpicking: the methods you've moved to `TarUtils` probably don't need to be public.

I agree and will change the code which was moved to the `TarUtils` to be not public.

theobisproject:
I rebased my changes and included a fix for COMPESS-544 in the TarFile implementation

PeterAlfredLee:
IMO, it's OK to have a random access for Tar - even through it would be slower as we read the `tar archive` to get the positions of entries. But I think we should hear more voice from others about this idea, cause this is a big change and `tar archives` are not designed to have a random access.

 @theobisproject And as you said, there are still some work(e.g. the refactor of the duplication) to do.

theobisproject:
Is anybody able to give me an information how we will progress with this change?
As I currently understand you wanted to hear for more feedback about this changes. Since I will have to do quite some work with the removal of code duplication I'd like to know how long you are willing to wait for feedback.
Also it would be good to get a review of the current state of the changes to guide me into the right direction with the refactorings which need to be done.

I will keep backporting changes from master regarding the TarArchiveInputStream and fix bugs found in the implementation in the meantime since I would like to avoid heading into the wrong direction with any further changes.

PeterAlfredLee:
Hi @theobisproject 
I have started a thread in mailing list about this.

> Since I will have to do quite some work with the removal of code duplication I'd like to know how long you are willing to wait for feedback.

I'm hoping to have this change dealed within version 1.21.

Thanks for your great work once again.

kinow:
>In the SparseFilesTest extracting of sparsefile-0.1 in pax_gnu_sparse.tar is skipped because of problems with tar on Ubuntu 16.04. I tested this on Ubuntu 18.04 and everything worked fine. Should we reenable extracting of the file?

+1, if that works with newer versions. Would be good to leave a NOTE/comment somewhere though, in case the test fails again. Even better if we were able to pinpoint what broke it in Ubuntu 16.04 (but can be done later :+1: )

To keep this PR simpler, we can leave as-is, and create a new JIRA to follow-up on this.

src/main/java/org/apache/commons/compress/archivers/tar/TarArchiveEntry.java
```diff
@@ -553,6 +554,23 @@ public TarArchiveEntry(final byte[] headerBuf, final ZipEncoding encoding, final
         parseTarHeader(headerBuf, encoding, false, lenient);
     }
 
+    /**
+     * Construct an entry from an archive's header bytes for random access tar. File is set to null.
+     * @param headerBuf The header bytes from a tar archive entry.
+     * @param encoding encoding to use for file names
+     * @param lenient when set to true illegal values for group/userid, mode, device numbers and timestamp will be
+     * ignored and the fields set to {@link #UNKNOWN}. When set to false such illegal fields cause an exception instead.
+     * @param dataOffset Position of the entry data in the random access file
```
kinow, line 563: Formatting is inconsistent. Some descriptions start with an upper case letter, others not. Some end with a period, others not.  But it's a nit-pick, and can be fixed later :+1:

src/main/java/org/apache/commons/compress/utils/BoundedNIOInputStream.java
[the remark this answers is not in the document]
kinow: Thanks for explaining your rationale @theobisproject . I don't have a better suggestion, but I agree with @PeterAlfredLee that it is confusing. Perhaps the old name?

src/main/java/org/apache/commons/compress/archivers/tar/TarFile.java
```diff
@@ -0,0 +1,732 @@
...
+     *
+     * GNU.sparse.map
+     *    Map of non-null data chunks. It is a string consisting of comma-separated values "offset,size[,offset-1,size-1...]"
+     *
+     *
+     * For PAX Format 1.X:
+     * The sparse map itself is stored in the file data block, preceding the actual file data.
+     * It consists of a series of decimal numbers delimited by newlines. The map is padded with nulls to the nearest block boundary.
+     * The first number gives the number of entries in the map. Following are map entries, each one consisting of two numbers
+     * giving the offset and size of the data block it describes.
```
kinow, line 422: I think the comment above could be formatted with HTML to use `<pre>`, `<p>`, etc. But I didn't render the Javadoc to see how it looks like.

kinow (approved):
No objections from me. I don't have a use case for that as I haven't worked with tar files that required random access. But I can imagine users could use it.

@theobisproject , great work! Thanks a lot for the contribution!

@theobisproject , @PeterAlfredLee , my only other comment is about the name `TarFile`. Is there any risk in using this class name, with a feature that is not part of the tar format (I don't know the format well enough, but I think you mentioned it in your e-mail to the mailing list Lee?)?

I mean, could it be that in the future we may need to create a `TarFile`, that doesn't/cannot support the random read, and in which case we would have to find another name?

Other than that, it looks good to me :+1: Change going into new files, there should be no regression in the existing code / API, and tests and comments added.

Thanks!
Bruno

PeterAlfredLee:
> In the SparseFilesTest extracting of sparsefile-0.1 in pax_gnu_sparse.tar is skipped because of problems with tar on Ubuntu 16.04. I tested this on Ubuntu 18.04 and everything worked fine. Should we reenable extracting of the file?

Ah. I missed this in my last reviews. It's wired. :)
I tested it on Ubuntu 18.04 and it do works now. That test was failed on Ubuntu 16.04 but unfortunly I don't have a Ubuntu 16.04 now. Agree with @kinow that we can leave it as-is and have it in a separate PR.

theobisproject:
@kinow Thank you for your feedback!


> > In the SparseFilesTest extracting of sparsefile-0.1 in pax_gnu_sparse.tar is skipped because of problems with tar on Ubuntu 16.04. I tested this on Ubuntu 18.04 and everything worked fine. Should we reenable extracting of the file?
> 
> +1, if that works with newer versions. Would be good to leave a NOTE/comment somewhere though, in case the test fails again. Even better if we were able to pinpoint what broke it in Ubuntu 16.04 (but can be done later 👍 )
> 
> To keep this PR simpler, we can leave as-is, and create a new JIRA to follow-up on this.

> Ah. I missed this in my last reviews. It's weird. :)
> I tested it on Ubuntu 18.04 and it do works now. That test was failed on Ubuntu 16.04 but unfortunly I don't have a Ubuntu 16.04 now. Agree with @kinow that we can leave it as-is and have it in a separate PR.


I can create a new JIRA and fix this on master next weekend if nothing get's in my way.




> @theobisproject , @PeterAlfredLee , my only other comment is about the name `TarFile`. Is there any risk in using this class name, with a feature that is not part of the tar format (I don't know the format well enough, but I think you mentioned it in your e-mail to the mailing list Lee?)?
> 
> I mean, could it be that in the future we may need to create a `TarFile`, that doesn't/cannot support the random read, and in which case we would have to find another name?

I don't think there is any risk even that the random access is not supported by the format but emulated by the implementation. You still can use the `TarFile` class to read the content sequentially since the order of the entries is preserved. Also this naming would be consistent with the already exisiting `ZipFile` implementation. 
From a performance point of view the overhead needed to allow the random access is as small as possible since only the headers of the entries are read (excluding some special cases with e.g. long filenames where the information is stored in a different entry) and all file data is jumped over by setting the position in the file to the start of the next header. My testing showed the `TarFile` and `TarArchiveStream` performance is equivalent for sequential access. The performance gain for accessing a single entry in the `TarFile` is huge because the `TarArchiveStream` needs to read and skip all data until it reaches the entry.  
If you would like some numbers for some cases feel free to name them. I can create some benchmarks and share them with you.

PeterAlfredLee:
> My testing showed the TarFile and TarArchiveStream performance is equivalent for sequential access. The performance gain for accessing a single entry in the TarFile is huge because the TarArchiveStream needs to read and skip all data until it reaches the entry.

> If you would like some numbers for some cases feel free to name them. I can create some benchmarks and share them with you.

That's interesting. Would you please upload the tests so I can have a look?

theobisproject:
The benchmark was added to the https://github.com/bodewig/commons-compress-benchmarks repository for easier setup.  

```java
package de.samaflost.commons_compress.tar;

import org.apache.commons.compress.archivers.tar.TarArchiveEntry;
import org.apache.commons.compress.archivers.tar.TarArchiveInputStream;
import org.apache.commons.compress.archivers.tar.TarFile;
import org.apache.commons.compress.utils.IOUtils;
import org.openjdk.jmh.annotations.Benchmark;
import org.openjdk.jmh.annotations.Fork;
import org.openjdk.jmh.annotations.Measurement;
import org.openjdk.jmh.annotations.OutputTimeUnit;
import org.openjdk.jmh.annotations.Scope;
import org.openjdk.jmh.annotations.State;
import org.openjdk.jmh.annotations.Warmup;
import org.openjdk.jmh.util.NullOutputStream;

import java.io.IOException;
import java.io.InputStream;
import java.nio.file.Files;
import java.nio.file.Path;
import java.nio.file.Paths;
import java.util.concurrent.TimeUnit;

@State(Scope.Benchmark)
@Fork(1)
@OutputTimeUnit(TimeUnit.SECONDS)
@Measurement(iterations = 10)
@Warmup(iterations = 10)
public class ReadLargeTarBenchmark {

    private InputStream getLargeTarStream() throws IOException {
        return Files.newInputStream(getLargeTarPath());
    }

    private Path getLargeTarPath() {
        return Paths.get("path-to-file");
    }

    private void readEntry(final InputStream inputStream) throws IOException {
        IOUtils.copy(inputStream, new NullOutputStream(), 8192);
    }

    @Benchmark
    public void readAllEntries_tarFile() throws IOException {
        try (TarFile tarFile = new TarFile(getLargeTarPath())) {
            for (TarArchiveEntry entry : tarFile.getEntries()) {
                try (InputStream entryInput = tarFile.getInputStream(entry)) {
                    readEntry(entryInput);
                }
            }
        }
    }

    @Benchmark
    public void readAllEntries_tarStream() throws IOException {
        try (TarArchiveInputStream tarIn = new TarArchiveInputStream(getLargeTarStream())) {
            TarArchiveEntry entry;
            while ((entry = tarIn.getNextTarEntry()) != null) {
                readEntry(tarIn);
            }
        }
    }

    @Benchmark
    public void readFirstEntry_tarFile() throws IOException {
        try (TarFile tarFile = new TarFile(getLargeTarPath())) {
            try (InputStream entryInput = tarFile.getInputStream(tarFile.getEntries().get(0))) {
                readEntry(entryInput);
            }
        }
    }

    @Benchmark
    public void readFirstEntry_tarStream() throws IOException {
        try (TarArchiveInputStream tarIn = new TarArchiveInputStream(getLargeTarStream())) {
            tarIn.getNextTarEntry();
            readEntry(tarIn);
        }
    }

    @Benchmark
    public void readSecondEntry_tarFile() throws IOException {
        try (TarFile tarFile = new TarFile(getLargeTarPath())) {
            try (InputStream entryInput = tarFile.getInputStream(tarFile.getEntries().get(1))) {
                readEntry(entryInput);
            }
        }
    }

    @Benchmark
    public void readSecondEntry_tarStream() throws IOException {
        try (TarArchiveInputStream tarIn = new TarArchiveInputStream(getLargeTarStream())) {
            tarIn.getNextTarEntry();
            tarIn.getNextTarEntry();
            readEntry(tarIn);
        }
    }

    @Benchmark
    public void readLastEntry_tarFile() throws IOException {
        try (TarFile tarFile = new TarFile(getLargeTarPath())) {
            try (InputStream entryInput = tarFile.getInputStream(tarFile.getEntries().get(tarFile.getEntries().size() - 1))) {
                readEntry(entryInput);
            }
        }
    }

    @Benchmark
    public void readLastEntry_tarStream() throws IOException {
        try (TarArchiveInputStream tarIn = new TarArchiveInputStream(getLargeTarStream())) {
            tarIn.getNextTarEntry();
            tarIn.getNextTarEntry();
            tarIn.getNextTarEntry();
            readEntry(tarIn);
        }
    }
}

```

This is the result on my machine (Windows 10, Java 11.0.9, File on SSD) with the testfile containing 3 times the Ubuntu 20.04 Desktop image.
```
Benchmark                                         Mode  Cnt  Score   Error  Units
ReadLargeTarBenchmark.readAllEntries_tarFile     thrpt   10  0,228 ± 0,003  ops/s
ReadLargeTarBenchmark.readAllEntries_tarStream   thrpt   10  0,256 ± 0,002  ops/s
ReadLargeTarBenchmark.readFirstEntry_tarFile     thrpt   10  0,680 ± 0,010  ops/s
ReadLargeTarBenchmark.readFirstEntry_tarStream   thrpt   10  0,766 ± 0,006  ops/s
ReadLargeTarBenchmark.readLastEntry_tarFile      thrpt   10  0,689 ± 0,006  ops/s
ReadLargeTarBenchmark.readLastEntry_tarStream    thrpt   10  0,127 ± 0,001  ops/s
ReadLargeTarBenchmark.readSecondEntry_tarFile    thrpt   10  0,690 ± 0,012  ops/s
ReadLargeTarBenchmark.readSecondEntry_tarStream  thrpt   10  0,218 ± 0,003  ops/s
```
I think this demonstrates the strength and weakness of the implementation.

**Strength**
- Performance of random access is constant
- Extracting all entries is near to the performance of the `TarArchiveStream`

**Weakness**
- Only extracting the first entry is slower because of the extra work which is done for the random access

The performance gap should be smaller when the archive is smaller. Feel free to experiment with different tar contents and number of files and report back your result.

kinow:
That's still a good approach for users that are aware of the weakness of the approach @theobisproject . Your performance test could be incorporated in commons-compress with JMH too, disabled by default. This way in case developers need to work on these classes, they may re-run the test to check for regressions.

PeterAlfredLee:
I have tested using your bench on my computer and got this :
```
Benchmark                                        Mode  Cnt   Score    Error  Units
ReadLargeTarBenchmark.readAllEntries_tarFile     avgt   10   5.601 ±  0.061   s/op
ReadLargeTarBenchmark.readAllEntries_tarStream   avgt   10   6.728 ±  0.012   s/op
ReadLargeTarBenchmark.readFirstEntry_tarFile     avgt   10   2.279 ±  0.017   s/op
ReadLargeTarBenchmark.readFirstEntry_tarStream   avgt   10   0.001 ±  0.001   s/op
ReadLargeTarBenchmark.readLastEntry_tarFile      avgt   10   2.266 ±  0.020   s/op
ReadLargeTarBenchmark.readLastEntry_tarStream    avgt   10  13.068 ±  0.030   s/op
ReadLargeTarBenchmark.readSecondEntry_tarFile    avgt   10   2.257 ±  0.012   s/op
ReadLargeTarBenchmark.readSecondEntry_tarStream  avgt   10   0.001 ±  0.001   s/op
```

The score is average time by setting `BenchmarkMode` to `Mode.AverageTime`. I'm testing with linux-4.4.tar, which has a size of 618MB with 55708 entries. I'm testing on my machine : Windows10, SSD, Java 8.

The result is interesting. Seems the `tarFile` needs an extra 2.65s to read all entries. `readAllEntries_tarStream` is slower than `readAllEntries_tarFile`, and `readLastEntry_tarStream` is even slower than eithor. Will try to figure out what's going on here.

theobisproject:
> `readLastEntry_tarStream` is even slower than eithor

My guess would be that the performance characteristic of `IOUtils.skip` is quite different than the writing to the `NullOutputStream` in the benchmark.

If someone is only interested in the overhead of the `TarFile` implementation this benchmark can be added.
```java
@Benchmark
public void tarFileOverhead() throws IOException {
    try (TarFile tarFile = new TarFile(getLargeTarPath())) {
        // Do nothing to only measure the overhead of reading all files header information
    }
}
```

PeterAlfredLee:
Hey @theobisproject .
Finally I can carve out some time these days. :(
I'm thinking that we can make this PR merged within this Christmas. WDYT?

theobisproject:
Hey @PeterAlfredLee I can invest some time starting from tomorrow. It would be good if you are able to go through the changes once again and comment on the TODOs I left in there.
Regarding the removal of the code duplication in my opinion this would require some major changes to the existing `TarArchiveInputStream`. So we should decide what we want to do about it in this PR.

PeterAlfredLee:
> So we should decide what we want to do about it in this PR.

Let's start with the easiest one : there are some conflicts now. Could you resolve them? :)

BTW : wish you a great christmas. :-)

theobisproject:
> > So we should decide what we want to do about it in this PR.
> 
> Let's start with the easiest one : there are some conflicts now. Could you resolve them? :)

I resolved the conflicts. As far as I could see there where no additional issues fixed in the new commits for the `TarArchiveInputStream` 😀 


> BTW : wish you a great christmas. :-)

Also wish you a merry christmas 🎄

src/main/java/org/apache/commons/compress/archivers/tar/TarFile.java
theobisproject: I formatted the Javadoc with html tags. Should be easier to read now when it is rendered

src/main/java/org/apache/commons/compress/archivers/tar/TarArchiveEntry.java
theobisproject: I formatted all the added Javadoc in `TarArchiveEntry` to start with lower case letter and end with a period.

src/main/java/org/apache/commons/compress/utils/BoundedNIOInputStream.java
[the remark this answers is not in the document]
theobisproject: Since the `BoundedInputStream` class is still there how about naming it `BoundedArchiveInputStream`?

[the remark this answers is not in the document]
PeterAlfredLee: +1
I don't have a better idea.

src/main/java/org/apache/commons/compress/archivers/tar/TarFile.java
```diff
@@ -0,0 +1,736 @@
...
+     * @throws IOException when reading the next TarEntry fails
+     */
+    private ByteBuffer getRecord() throws IOException {
+        ByteBuffer headerBuf = readRecord();
+        setAtEOF(isEOFRecord(headerBuf));
+        if (isAtEOF() && headerBuf != null) {
+            // Consume rest
+            tryToConsumeSecondEOFRecord();
+            // TODO: This is present in the TarArchiveInputStream but I don't know if we need this in the random access implementation. All tests are passing ...
+            // consumeRemainderOfLastBlock();
```
PeterAlfredLee, line 552: I think we should leave the `consumeRemainderOfLastBlock` as it was.

According to the comment :
```
    /**
     * This method is invoked once the end of the archive is hit, it
     * tries to consume the remaining bytes under the assumption that
     * the tool creating this archive has padded the last block.
     */
```

This is used to deal with the last block with padding zero. Even through some tar implementions don't implement this, this is part of tar specifiction.

See also [this](https://www.gnu.org/software/tar/manual/html_node/Standard.html) : 

> When writing an archive, the last record of blocks should be written at the full size, with blocks after the zero block containing all zeros.

theobisproject: I added the implementation. The test which checks this for the `TarArchiveInputStream` was a bit obscured. I added a bit of documentation so the next one finds it easier.

PeterAlfredLee:
The coverage of TarFile looks good enough. Are there any other work left now? @theobisproject

PeterAlfredLee:
And the japicmp report no BC breaking change.

theobisproject:
I've gone through the changes once more and noticed the changes from COMPRESS-559 is missing for the sparse entry test with `TarFile`.
> Are there any other work left now? @theobisproject

I'd like to hear your opinion about my TODO comments in the `SparseFilesTest`. I left out the `canReadEntryData` method which the `TarArchiveInputStream` has but kept the checks in my added tests. I don't think we need this method because the `TarFile#getInputStream` only takes an instance of `TarArchiveEntry`. Should I add this method or just remove the commented out code?

PeterAlfredLee:
> I don't think we need this method because the TarFile#getInputStream only takes an instance of TarArchiveEntry.

+1.

The `canReadEntryData` in `TarArchiveInputStream` was checking if the entry is a sparse entry or not - cause the sparse entries were not supported. The sparse entry was supported later but we could not remove the `canReadEntryData` because it's a  API breaking change.

I don't think we need it in `TarFile` any more.

theobisproject:
I removed the code. Now everything I wanted to address is done.

src/main/java/org/apache/commons/compress/archivers/tar/TarArchiveEntry.java
```diff
@@ -553,6 +554,23 @@ public TarArchiveEntry(final byte[] headerBuf, final ZipEncoding encoding, final
...
+     * @param dataOffset position of the entry data in the random access file.
+     * @since 1.21
+     * @throws IllegalArgumentException if any of the numeric fields have an invalid format.
+     * @throws IOException on error.
+     */
+    public TarArchiveEntry(final byte[] headerBuf, final ZipEncoding encoding, final boolean lenient,
+            final long dataOffset) throws IOException {
+        this(headerBuf, encoding, lenient);
+        this.dataOffset = dataOffset;
+    }
```
garydgregory, line 572: Should we throw an IAE if the offset is negative or larger than the parent can hold?

src/main/java/org/apache/commons/compress/utils/BoundedArchiveInputStream.java
```diff
@@ -0,0 +1,98 @@
+/*
```
garydgregory, line 1: Is it time to depend on Commons IO instead of creating a mini Commons IO in Compress?

src/test/java/org/apache/commons/compress/archivers/TarTestCase.java
```diff
@@ -335,6 +482,18 @@ public void testCOMPRESS178() throws Exception {
         in.close();
     }
 
+    @Test
+    public void testTarFileCOMPRESS178() throws Exception {
+        final File input = getFile("COMPRESS-178.tar");
+        try (final TarFile tarFile = new TarFile(input)) {
+            // Compared to the TarArchiveInputStream all entries are read when instantiating the tar file
+            fail("Expected IOException");
+        } catch (final IOException e) {
```
garydgregory, line 491: Use assertThrows().

src/main/java/org/apache/commons/compress/utils/BoundedArchiveInputStream.java
PeterAlfredLee: I'm +1 for this.
But IIRC there was a issue or mailing discussing about this : the `skip` implemention in Commons IO didn't use the `skip` of the InputStream it's reading from. I will try to find this.

src/main/java/org/apache/commons/compress/archivers/tar/TarArchiveEntry.java
theobisproject: It now throws an IAE if the offset is negative. Detecting if the offset is larger than the complete file is not possible here. I testet setting the offset of an entry to the end of the file. When reading the inputstream there is immediately an IOException thrown.

src/test/java/org/apache/commons/compress/archivers/TarTestCase.java
theobisproject: `assertThrows` is not available in Junit 4. The test is equivalent to the already exisiting test for the `TarArchiveInputStream`. Should I change both tests to use the `expected` property in the `@Test` annotation?

PeterAlfredLee: I prefer to leave it as is.

PeterAlfredLee:
> I removed the code. Now everything I wanted to address is done.

Looks good. I will do some final checks these days.

PeterAlfredLee:
Thanks @theobisproject for your hard work!

theobisproject:
Thank you all for taking the time to review this change!

PR #120, opened 2020-07-27
COMPRESS-542: Lazy allocation of SevenZArchiveEntry to prevent OOM on corrupt files

theobisproject:
This PR is nearly the same as the patch I attached to the Jira ticket. I made an additional null check before passing the initialized files to the archive to avoid NullPointerExceptions later on.
As said in the ticket it would be nice to check if the headers are ok before initializing the files but with my current knowledge of 7z this is out of scope for me. This patch works for the files I encountered even if I have to admit the patch is not very beautiful.


PeterAlfredLee:
Hi @theobisproject 
I'm a little interested about the amount of `SevenZArchiveEntry`. How many entries do you have in your 7z archive that take so much memory?

theobisproject:
Hi @PeterAlfredLee 
as explained in the Jira Ticket it was a corrupted archive where the `numFiles` variable read from the header in `readFilesInfo` was about 138 million entries. I don't know what would be the correct number for the archive but this was the number commons-compress was reading.

PeterAlfredLee:
I see.
This PR looks good now. Will look more into it these days.

garydgregory:
My biggest issue with this PR is that it is missing a unit test. Without a failing unit test, there is no way to know that this fixes anything or that a future change will regress to the previous behavior. Please add a test ;-)

I also see a lot duplicate code like (5 times!)
```
if (files[i] == null) {
      files[i] = new SevenZArchiveEntry();
}
```
which should be easily refactored.

theobisproject:
As you are a bit more familiar with the 7z archive maybe you have an idea how to avoid the entry allocation completely before we are sure the header is not corrupted. It was more a hack from me to process the data somehow.

theobisproject:
> My biggest issue with this PR is that it is missing a unit test. Without a failing unit test, there is no way to know that this fixes anything or that a future change will not regress to the previous behavior. Please add a test ;-)

As this was encountered with production data there is no way I can share those and creating an artifical archive which is reproducing this is way beyond my current understanding of the format. The only way I see to currently add a test is when making the `readFilesInfo` method protected.

akelday:
> avoid the entry allocation completely before we are sure the header is not corrupted

Probably not possible with the current code... **tryToLocateEndHeader** is the real cause because it does no CRC check and cannot, because by definition it's already a corrupt file.

I have crafted a 233 byte malformed 7z which would attempt to allocate 268,435,455 files but I'm not certain it's wise to post it here. This is in some way related to my own problems with a very large 7z because the "kName" section allocates an enormous buffer for filenames (fixable by streaming the bytes instead).

theobisproject:
Just a short update. I will try to generate a reproducting file via fuzzing which is hopefully successful. This might take some days but I think at the end of the weekend I can share somthing.

theobisproject:
The test files @akelday attached to the jira issue are replicating the issue. Since this issue is dependend on the available Java heap it could be difficult to add this as unit test since I don't know a way to force the Java heap for a single test in JUnit. As far as I know it could only be set at the maven-surefire-plugin.

src/main/java/org/apache/commons/compress/archivers/sevenz/SevenZFile.java
```diff
@@ -981,8 +979,13 @@ private void readFilesInfo(final ByteBuffer header, final Archive archive) throw
                     int nextName = 0;
                     for (int i = 0; i < names.length; i += 2) {
                         if (names[i] == 0 && names[i+1] == 0) {
-                            files[nextFile++].setName(new String(names, nextName, i-nextName, StandardCharsets.UTF_16LE));
+                            String fName = new String(names, nextName, i - nextName, StandardCharsets.UTF_16LE);
+                            if (files[nextFile] == null) {
+                                files[nextFile] = new SevenZArchiveEntry();
+                            }
+                            files[nextFile].setName(fName);
```
PeterAlfredLee, line 986: I think we do not need a variable `fName` here :

```
if (files[nextFile] == null) {
    files[nextFile] = new SevenZArchiveEntry();
}
files[nextFile].setName(new String(names, nextName, i - nextName, StandardCharsets.UTF_16LE));
```

src/main/java/org/apache/commons/compress/archivers/sevenz/SevenZFile.java
```diff
@@ -997,6 +1000,9 @@ private void readFilesInfo(final ByteBuffer header, final Archive archive) throw
                         throw new IOException("Unimplemented");
                     }
                     for (int i = 0; i < files.length; i++) {
+                        if (files[i] == null) {
+                            files[i] = new SevenZArchiveEntry();
+                        }
```
PeterAlfredLee, line 1005: Agree with @garydgregory about this duplicate code. We can have a simple refactor here.

src/main/java/org/apache/commons/compress/archivers/sevenz/SevenZFile.java
```diff
@@ -1090,7 +1108,13 @@ private void readFilesInfo(final ByteBuffer header, final Archive archive) throw
                 ++emptyFileCounter;
             }
         }
-        archive.files = files;
+        List<SevenZArchiveEntry> entries = new ArrayList<>();
```
PeterAlfredLee, line 1111: Maybe a final here?
```
final List<SevenZArchiveEntry> entries = new ArrayList<>();
```

garydgregory: +1
I use `final` as much as possible to make it obvious when reading or debugging whether I need to even think about a given local variable, for more details on my reasoning, please see https://garygregory.wordpress.com/2013/01/26/the-final-kiss-in-java/

src/main/java/org/apache/commons/compress/archivers/sevenz/SevenZFile.java
theobisproject: Yes, I got rid of the variable in the latest commit

src/main/java/org/apache/commons/compress/archivers/sevenz/SevenZFile.java
theobisproject: Code is refactored as part of the latest commit

theobisproject:
In the files I got from the fuzzing I noticed that even the array creation can lead to an OOM exception. I tried to replace it with a list and let it grow as needed but it didn't worked well. The current solution with the Map works (all tests pass on a Raspberry Pi 3) but has high memory allocation because of autoboxing the integer.

PeterAlfredLee:
Looks good. Thank you for your work @theobisproject .
BTW how did you do the fuzzing? I was thinking if we could have some fuzz tests in Compress.

theobisproject:
I am using this project https://github.com/rohanpadhye/jqf to do the fuzzing via the AFL bridge provided. This was the fuzz method implementation I used
```java
@Fuzz
public void sevenzOom(@From(InputStreamGenerator.class) InputStream in) {
    try (SevenZFile sevenZFile = new SevenZFile(new SeekableInMemoryByteChannel(IOUtils.toByteArray(in)))) {
        Assume.assumeTrue(sevenZFile.getNextEntry() != null);
    } catch (IOException e) {
        // Ignore
    }
}
```

PR #152, opened 2020-11-22
COMPRESS-559: Extract sparsefile-0.1 also on Linux

theobisproject:
As discussed in https://github.com/apache/commons-compress/pull/113 the problem can no longer be seen in newer Ubuntu versions

### DIFF
--- a/src/test/java/org/apache/commons/compress/archivers/tar/SparseFilesTest.java
+++ b/src/test/java/org/apache/commons/compress/archivers/tar/SparseFilesTest.java
@@ -189,11 +189,9 @@ public class SparseFilesTest extends AbstractTestCase {
                         IOUtils.toByteArray(sparseFileInputStream));
             }
 
-            // TODO : it's wired that I can only get a 0 size sparsefile-0.1 on my Ubuntu 16.04
-            //        using "tar -xf pax_gnu_sparse.tar"
             paxGNUEntry = tin.getNextTarEntry();
             assertTrue(tin.canReadEntryData(paxGNUEntry));
-            try (InputStream sparseFileInputStream = extractTarAndGetInputStream(file, "sparsefile-0.0")) {
+            try (InputStream sparseFileInputStream = extractTarAndGetInputStream(file, "sparsefile-0.1")) {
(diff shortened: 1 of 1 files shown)


theobisproject:
The described problem seems to be still there on Ubuntu 16.04 since the test fails in Travis CI. Should we investigate further?

PeterAlfredLee:
> Should we investigate further?

Trying to figure out what's wrong here. And I think we can have a comment here to indicate the test may fail on Ubuntu16.04.

PeterAlfredLee:
> Should we investigate further?

Ubuntu 16.04 is using gnu tar whose version is 1.28, and 18.04 is using 1.29 instead.
According to the [releases description](https://www.gnu.org/software/tar/), the sparse detection is modified in version 1.29. After cloning the [gnu tar git](https://git.savannah.gnu.org/git/tar.git), I found that the commit `b68432` modified the sparse detection a lot.

I believe there're some potential problems about sparse detection in gnu tar 1.28, which was fixed in 1.29.

theobisproject:
I also think the tar version 1.28 is simply broken for the archive we are extracting in the test. The file size is always 0. I could do a detection like the check for windows either if we are on Ubuntu 16.04 or the tar version available is 1.28. WDYT?
Maybe we should also think about switching to some newer Ubuntu version in Travis since for 16.04 the standard support will end in April 2021.

PeterAlfredLee:
> I could do a detection like the check for windows either if we are on Ubuntu 16.04 or the tar version available is 1.28. WDYT?

+1

> Maybe we should also think about switching to some newer Ubuntu version in Travis since for 16.04 the standard support will end in April 2021.

I'm working on this.

PeterAlfredLee:
> Maybe we should also think about switching to some newer Ubuntu version in Travis since for 16.04 the standard support will end in April 2021.

I have made a [PR about this](https://github.com/apache/commons-compress/pull/156) and everything looks good.

theobisproject:
I finally found the time to add the check and test it. I've done the check for the tar version since this might also affect other distros with the same tar version if we only check for Ubuntu 16.04

PR #153, opened 2020-11-22
COMPRESS-560: Do not return false if the entry is a tar sparse entry

theobisproject:
Also some unit test cleanup to use assumeTrue/assumeFalse with the OS check. Now tests will be shown as skipped instead of success when they are not run.

### DIFF
--- a/src/test/java/org/apache/commons/compress/archivers/tar/SparseFilesTest.java
+++ b/src/test/java/org/apache/commons/compress/archivers/tar/SparseFilesTest.java
@@ -75,7 +75,7 @@ public class SparseFilesTest extends AbstractTestCase {
 
     @Test
     public void testExtractSparseTarsOnWindows() throws IOException {
-        assumeTrue(isOnWindows);
+        assumeTrue("This test should be ignored if not running on Windows", isOnWindows);
 
         final File oldGNUSparseTar = getFile("oldgnu_sparse.tar");
         final File paxGNUSparseTar = getFile("pax_gnu_sparse.tar");
@@ -115,7 +115,7 @@ public class SparseFilesTest extends AbstractTestCase {
 
     @Test
     public void testExtractOldGNU() throws IOException, InterruptedException {
-        assumeFalse(isOnWindows);
+        assumeFalse("This test should be ignored on Windows", isOnWindows);
 
         try {
             final File file = getFile("oldgnu_sparse.tar");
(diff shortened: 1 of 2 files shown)


src/test/java/org/apache/commons/compress/archivers/tar/SparseFilesTest.java
```diff
@@ -73,54 +75,57 @@ public void testPaxGNU() throws Throwable {
 
     @Test
     public void testExtractSparseTarsOnWindows() throws IOException {
-        if (!isOnWindows) {
-            return;
-        }
+        assumeTrue(isOnWindows);
```
PeterAlfredLee, line 78: Maybe we can make it more detailed with something like :
```
assumeTrue("This test should be ignored if not running on Windows",isOnWindows);
```

src/test/java/org/apache/commons/compress/archivers/tar/SparseFilesTest.java
```diff
@@ -73,54 +75,57 @@ public void testPaxGNU() throws Throwable {
...
             }
         }
     }
 
     @Test
     public void testExtractOldGNU() throws IOException, InterruptedException {
-        if (isOnWindows) {
-            return;
-        }
+        assumeFalse(isOnWindows);
```
PeterAlfredLee, line 118: Maybe we can make it more detailed with something like :
```
assumeFalse("This test should be ignored on Windows", isOnWindows);
```

src/test/java/org/apache/commons/compress/archivers/tar/SparseFilesTest.java
```diff
@@ -129,14 +134,13 @@ public void testExtractOldGNU() throws IOException, InterruptedException {
 
     @Test
     public void testExtractExtendedOldGNU() throws IOException, InterruptedException {
-        if (isOnWindows) {
-            return;
-        }
+        assumeFalse(isOnWindows);
```
PeterAlfredLee, line 137: Same one.

src/test/java/org/apache/commons/compress/archivers/tar/SparseFilesTest.java
```diff
@@ -169,31 +173,32 @@ public void testExtractExtendedOldGNU() throws IOException, InterruptedException
 
     @Test
     public void testExtractPaxGNU() throws IOException, InterruptedException {
-        if (isOnWindows) {
-            return;
-        }
+        assumeFalse(isOnWindows);
```
PeterAlfredLee, line 176: Same one.

src/test/java/org/apache/commons/compress/archivers/tar/SparseFilesTest.java
theobisproject: Good idea. I added the messages everywhere

PR #165, opened 2021-02-10
OMPRESS-565 : add BufferPool support in zstd

PeterAlfredLee:
Add `BufferPool` as a configuration of `ZstdCompressorInputStream`

### DIFF
--- a/src/main/java/org/apache/commons/compress/compressors/zstandard/ZstdCompressorInputStream.java
+++ b/src/main/java/org/apache/commons/compress/compressors/zstandard/ZstdCompressorInputStream.java
@@ -21,6 +21,7 @@ package org.apache.commons.compress.compressors.zstandard;
 import java.io.IOException;
 import java.io.InputStream;
 
+import com.github.luben.zstd.BufferPool;
 import com.github.luben.zstd.ZstdInputStream;
 import org.apache.commons.compress.compressors.CompressorInputStream;
 import org.apache.commons.compress.utils.CountingInputStream;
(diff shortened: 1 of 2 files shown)


heuermh:
See also #163

heuermh:
Thanks @PeterAlfredLee!

This LGTM for the use case where callers are using the `ZstdCompressorInputStream` constructor directly.

How might we support the [`CompressorStreamFactory.createCompressorInputStream(InputStream)`](
http://commons.apache.org/proper/commons-compress/javadocs/api-release/org/apache/commons/compress/compressors/CompressorStreamFactory.html#createCompressorInputStream-java.io.InputStream-) use case?

PeterAlfredLee:
> See also #163

I can't see much differences between 1.4.8-1 and 1.4.8-3 of `zstd-jni`. Besides that, 1.4.8-4 is on the way(released 2 days ago).
I think we can leave the pom as it is.

PeterAlfredLee:
> How might we support the CompressorStreamFactory.createCompressorInputStream(InputStream) use case?

-1. We do not support detailed configurations in `CompressorStreamFactory`. It's a simple input stream factory with compressor type detectors. I think we should keep it simple enough.

heuermh:
> > How might we support the CompressorStreamFactory.createCompressorInputStream(InputStream) use case?
> 
> -1. We do not support detailed configurations in `CompressorStreamFactory`. It's a simple input stream factory with compressor type detectors. I think we should keep it simple enough.

Fair enough.  I can do the stream format check beforehand; I use `CompressorStreamFactory` as a fallback.

PR #168, opened 2021-02-19
COMPRESS-566 allow gzip buffer size to be configured

bokken:


src/main/java/org/apache/commons/compress/compressors/gzip/GzipParameters.java
```diff
@@ -118,4 +120,22 @@ public int getOperatingSystem() {
     public void setOperatingSystem(final int operatingSystem) {
         this.operatingSystem = operatingSystem;
     }
+
```
garydgregory, line 123: You're missing a Javadoc comment here.

garydgregory (changes requested):
See comment.

src/main/java/org/apache/commons/compress/compressors/gzip/GzipParameters.java
bokken: The rest of the attributes in the class only had javadoc for setter method. I had started writing javadoc but removed it to match the rest. I can certainly add.

src/main/java/org/apache/commons/compress/compressors/gzip/GzipParameters.java
```diff
@@ -118,4 +120,28 @@ public int getOperatingSystem() {
...
+    /**
+     * Sets size of the buffer used to retrieve compressed data from
+     * {@link Deflater} and write to underlying {@link OutputStream}.
+     * 
+     * @param bufferSize the bufferSize to set. Must be a positive value.
+     * @since 1.21
+     */
+    public void setBufferSize(int bufferSize) {
+        if (bufferSize <= 0) {
```
PeterAlfredLee, line 142: Do we need a high limit check for `bufferSize` here?

bokken: I personally do not see the value, but can add if desired. 
There is certainly some threshold where larger values are counter productive, but my opinion is consumers can determine that on their own.

bokken:
Do I need to do anything else here @garydgregory ?

bodewig:
@garydgregory @PeterAlfredLee is there anythin gleft that needs to be done before merging this?

PeterAlfredLee:
This looks good to me. :-)

PR #169, opened 2021-02-22
COMPRESS-565 : add a new AlwaysWithCompatibility in Zip64Mode

PeterAlfredLee:
Add a new AlwaysWithComestibles in Zip64Mode, this is a compromise for some libraries including 7z and Expand-Archive Powershell utility(and likely Excel).

With `Zip64Mode.AlwaysWithComestibles`, the following fields in Central File Header will be set as needed:
1. disk number offset in split zip
2. relative offset of LFH
3. relative header offset in zip64 extra
4. disk number start in zip64 extra

These changes are consistent with commit `2b7281`.

It would be good if the generated file in test could tested by Expand-Archive or 7z, but I didn't find a way to detect them in Java.

It's welcome for suggestions about the name `Zip64Mode.AlwaysWithComestibles`  :)

See [COMPRESS-565](https://issues.apache.org/jira/projects/COMPRESS/issues/COMPRESS-565?filter=allissues&orderby=created+DESC%2C+priority+DESC%2C+updated+DESC) for more detailed information.

### DIFF
--- a/src/main/java/org/apache/commons/compress/archivers/zip/Zip64Mode.java
+++ b/src/main/java/org/apache/commons/compress/archivers/zip/Zip64Mode.java
@@ -43,5 +43,15 @@ public enum Zip64Mode {
      * Use Zip64 extensions for all entries where they are required,
      * don't use them for entries that clearly don't require them.
      */
-    AsNeeded
+    AsNeeded,
+    /**
+     * Always use Zip64 extensions for LFH and central directory as
+     * {@link Zip64Mode#Always} did, and at the meantime encode
+     * the relative offset of LFH and disk number start as needed in
+     * CFH as {@link Zip64Mode#AsNeeded} did.
+     * <p>
+     * This is a compromise for some libraries including 7z and
+     * Expand-Archive Powershell utility(and likely Excel).
+     */
+    AlwaysWithCompatibility
 }
--- a/src/main/java/org/apache/commons/compress/archivers/zip/ZipArchiveOutputStream.java
+++ b/src/main/java/org/apache/commons/compress/archivers/zip/ZipArchiveOutputStream.java
@@ -749,7 +749,8 @@ public class ZipArchiveOutputStream extends ArchiveOutputStream {
     }
 
     private boolean isZip64Required(final ZipArchiveEntry entry1, final Zip64Mode requestedMode) {
-        return requestedMode == Zip64Mode.Always || isTooLargeForZip32(entry1);
+        return requestedMode == Zip64Mode.Always || requestedMode == Zip64Mode.AlwaysWithCompatibility
+                || isTooLargeForZip32(entry1);
     }
 
     private boolean isTooLargeForZip32(final ZipArchiveEntry zipArchiveEntry){
@@ -940,6 +941,7 @@ public class ZipArchiveOutputStream extends ArchiveOutputStream {
      */
     private boolean shouldAddZip64Extra(final ZipArchiveEntry entry, final Zip64Mode mode) {
         return mode == Zip64Mode.Always
+            || mode == Zip64Mode.AlwaysWithCompatibility
             || entry.getSize() >= ZIP64_MAGIC
             || entry.getCompressedSize() >= ZIP64_MAGIC
             || (entry.getSize() == ArchiveEntry.SIZE_UNKNOWN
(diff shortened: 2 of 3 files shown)


bodewig:
I like the code change.

It is good you always include a central directory zip64 when there has been one in LFH, something I likely would have overlooked. Thanks. The javadoc of the new option is not completely accurate, as it will always add zip64 entries to the central directory, but not necessarily encode the disk number and relative offset.

Oh, I believe we are overlooking something. We must ensure we add the relative LFH offset to the Zip64 extended information entry inside the central directory even if it is not too big in the case where the disk number is too big - or we'd create a gap inside of the extra field. This is currently wrong inside master as well, unless I'm wrong.

I don't like the name of the new option, but I'm the last one who should suggest a different one - I've got a long track record of picking bad names. Let's discuss this on the dev list.

PeterAlfredLee:
> The javadoc of the new option is not completely accurate, as it will always add zip64 entries to the central directory, but not necessarily encode the disk number and relative offset.

Yes, it is. I pushed a new commit with only the javadoc of the new option changed.

> We must ensure we add the relative LFH offset to the Zip64 extended information entry inside the central directory even if it is not too big in the case where the disk number is too big - or we'd create a gap inside of the extra field. This is currently wrong inside master as well, unless I'm wrong.

This means we should encode the relative offset and disk number at the same time - even through only eithor of them is too big.  Am I right about it?

> Let's discuss this on the dev list.

+1.

PeterAlfredLee:
And I made a typo in the name - I was thinking about the name `AlwaysWithCompatibility`.

missingdays:
Hi!
Any change this fix will be releases in the following couple of days? We have a release soon, and if this fix won't be released, I'll have to use custom jar with this fix, which will lead to some problems for us.

bodewig:
I'm sorry but the master branch isn't quite ready for a new release, yet, so unfortunately you can't expect a release within days.

bodewig:
> This means we should encode the relative offset and disk number at the same time - even through only eithor of them is too big. Am I right about it?

No, this is not what I meant.

In the layout of the ZIP64 extended information extra field you have the two sizes - which must always be present - then the LFH offset, then the disk number. If you don't need to encode the LFH offset but have to encode the disk number the disk number would appear directly after the sizes. I'm afraid there might be parsers that try to read the LFH offset anyway if the size of the extra field is bigger than the two size fields. I may be overcautious (the parser should consult the corresponding field inside of the central directory actually contains FFs and it should see the size is only 20 bytes not 24 or even 28) but prefer to stay on the safe side.

`AlwaysWithCompatibility` sounds a lot better, thanks :-)

missingdays:
> I'm sorry but the master branch isn't quite ready for a new release

Ok, but if I build the jar from this branch, is this ok to use in release?

garydgregory:
> > I'm sorry but the master branch isn't quite ready for a new release
> 
> Ok, but if I build the jar from this branch, is this ok to use in release?

If you want to release the software on your own, and especially if you plan on pushing it out to the world through Maven Central or elsewhere, you must change the Maven Group and Artifact IDs.

missingdays:
> If you want to release the software on your own, and especially if you plan on pushing it out to the world through Maven Central or elsewhere, you must change the Maven Group and Artifact IDs.

I mean bundle the jar to our release, not release the library anywhere. 
Is this branch stable enough for the production usage?

bodewig:
I'd say it depends on which parts of Compress you use. If you are only using code that hasn't changed or has only seen bugfixes, you should be pretty safe. Usually we do some extra code reviews when preparing a release and this may not have happened fully for new code.

PeterAlfredLee:
I'm thinking about merge this PR. 
Are there anything to be modified about this PR? @bodewig

garydgregory:
Surely we can come up with a better name than a "food item" (comestible) or is that term used in the zip spec itself?

bodewig:
you may want to change the name of the test method. Other than that, please just go ahead.

@garydgregory "comestible" is almost only present in the PR's title. Peter has fixed the occurrences in code already.

bodewig:
@PeterAlfredLee I had another look and I don't think you've addressed my comment from https://github.com/apache/commons-compress/pull/169#issuecomment-788865243 - but we can deal with that after the PR has been merged.

PeterAlfredLee:
> Surely we can come up with a better name than a "food item" (comestible) or is that term used in the zip spec itself?

That was a typo and I'm using `Compatibility` now. But I forgot to change the title of this PR, my bad.

PeterAlfredLee:
> I had another look and I don't think you've addressed my comment from #169 (comment) 

I must have missed that. Sorry about that.
I will fix this today and push a commit in this PR.

PeterAlfredLee:
I pushed a commit that fixes name of tests, and also fixes the [comment](https://github.com/apache/commons-compress/pull/169#issuecomment-788865243).

WDYT? @bodewig @garydgregory

PeterAlfredLee:
The Github Actions is weird - I can successfully compile and test the class `Zip64SupportIT`, but Actions is reporting compilation failure, which is caused by the missing of `FileOutputStream` and `FileInputStream`. These 2 classes are already imported but Actions seems missed them.

Maybe this is a bug of Github Actions?

bodewig:
looks good to me

PR #178, opened 2021-03-25
Minor Improvements

arturobernalg:
* Simplify conditions
* Use lambda and method reference


bodewig:
thank you @arturobernalg

PR #192, opened 2021-05-21
Fix typo and grammar

HelderMagalhaes:


bodewig:
thank you. Please remove the hunk that just changes whitespace at the top of TarArchiveEntry

HelderMagalhaes:
> Please remove the hunk that just changes whitespace at the top of TarArchiveEntry

The tabs vs. spaces mix does lead to unexpected visual formatting (only OK in a properly configured text editor):
image

The only reason that occurred to leave as-is was the snippet being used for a straight comparison (`diff`) with the _xstar_ repository. Removing that as per the request.

bodewig:
Well, first of all we are not all the same, my opinion may differ from the one of others who also contribute to COMPRESS. So take what I say with a big grain of salt, please. To me readable diffs are far more important than sticking to arbitrary formatting rules, so my first rule would be to never change the formatting of existing code - if it really seems necessary, do it in a different step so that the noise of whitespace changes doesn't distract from what you really wanted to change.

bodewig:
I've merged your first two commits, thanks again.

PR #194, opened 2021-05-22
COMPRESS-578 - Java 8 improvements

arturobernalg:


src/main/java/org/apache/commons/compress/archivers/sevenz/SevenZFile.java
```diff
@@ -937,7 +936,7 @@ private void sanityCheckSubStreamsInfo(final ByteBuffer header, final ArchiveSta
             for (int i = 0; i < stats.numberOfFolders; i++) {
                 numUnpackSubStreamsPerFolder.add(assertFitsIntoNonNegativeInt("numStreams", readUint64(header)));
             }
-            stats.numberOfUnpackSubStreams = numUnpackSubStreamsPerFolder.stream().collect(Collectors.summingLong(Integer::longValue));
+            stats.numberOfUnpackSubStreams = numUnpackSubStreamsPerFolder.stream().mapToLong(Integer::longValue).sum();
```
bodewig, line 939: what is the advantage of doing it this way? This is a curious question. When I wrote the code I considered both options and considered the later to better express my intention - but I'm more than willing to change it if this is not just a metter of taste.

arturobernalg: Hi @bodewig 
 It allows to avoid creating redundant temporary objects when traversing a collection.
TY

bodewig: Why do you think the current alternative would require temporary objects? The summingLong collector is working on primitive longs just like sum() is, isn't it?

arturobernalg: Hi @bodewig 
please check https://rules.sonarsource.com/java/RSPEC-4034

bodewig: well, the examples are quite different from this case and I agree with all of them.

Please do me a favor and separate this change from the rest of the PR and the other team members can chime in with their opinion. I'm not totally against the change just consider it non-necessary. "readablity" is not really an objective measure as you can tell from our conversation.

arturobernalg: Agree. Changed

bodewig: thank you  @arturobernalg

PR #195, opened 2021-05-22
COMPRESS-579 - Remove redundant local variable

arturobernalg:


src/test/java/org/apache/commons/compress/archivers/TarTestCase.java
```diff
@@ -569,7 +569,7 @@ public void testTarFileLongNameLargerThanBuffer() throws IOException {
     }
 
     private String createLongName(final int nameLength) {
-        final StringBuffer buffer = new StringBuffer();
+        final StringBuilder buffer = new StringBuilder();
```
bodewig, line 572: I don't see any redundant variable here :-)

I assume you added changes you intended to create as separate commits by accident. Please make sure your PR title and description matches what your change actually does.

arturobernalg: Yep, sorry. 
It was intentionally, I didn't think another PR was necessary. Next time will be more careful
TY

PR #196, opened 2021-05-22
COMPRESS-580 - Remove redundant operation

arturobernalg:
- Redundant array creation
- Redundant type arguments
- Redundant type cast
- Redundant String operation
- Unnecessarily escaped character


src/main/java/org/apache/commons/compress/archivers/arj/ArjArchiveInputStream.java
```diff
@@ -115,10 +115,10 @@ private String readString(final DataInputStream dataIn) throws IOException {
                 buffer.write(nextByte);
             }
             if (charsetName != null) {
-                return new String(buffer.toByteArray(), charsetName);
+                return buffer.toString(charsetName);
```
bodewig, line 118: oh, I was unaware of ByteArrayOutputStream#toString(String). Learn something new every day. Thank you!

PR #206, opened 2021-06-12
Minor changes

arturobernalg:
* Simplify conditions
* Make final var

### DIFF
--- a/src/main/java/org/apache/commons/compress/archivers/zip/ZipArchiveEntry.java
+++ b/src/main/java/org/apache/commons/compress/archivers/zip/ZipArchiveEntry.java
@@ -776,8 +776,7 @@ public class ZipArchiveEntry extends java.util.zip.ZipEntry
      */
     @Override
     public boolean isDirectory() {
-        final String n = getName();
-        return n != null && n.endsWith("/");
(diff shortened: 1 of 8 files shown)


garydgregory:
Please rebase on master. Recent changes should allow all builds to be green including Java 16 and 17-EA.

arturobernalg:
> Please rebase on master. Recent changes should allow all builds to be green including Java 16 and 17-EA.

Hi @garydgregory 
done
TY

src/main/java/org/apache/commons/compress/archivers/zip/ZipArchiveEntry.java
```diff
@@ -777,7 +777,7 @@ public String getName() {
     @Override
     public boolean isDirectory() {
         final String n = getName();
-        return n != null && n.endsWith("/");
+        return n.endsWith("/");
```
garydgregory, line 780: It does seem that `getName()` can never be null, so maybe just `return getName().endsWith("/");` ? Same for other getName() call sites?

garydgregory: ping?

PR #210, opened 2021-07-02
Pack200 harmony as fallback

bodewig:
Hi @garydgregory 

I'm only half-serious about this PR.

Personally I'd like to use java.util.jar.Pack200 if it is available and fallback to Harmony's code if it isn't. I've managed to do so with a bunch of facades and a little reflection, but it seems to work - all tests pass for all JDKs tested.

I had to pick up a few changes from master to make things work in github actions, but we need to rebase anyway.

Right now only javadoc stuff fails AFAICT, and even that is probably going to pass once the change is rebased.

If you like my changes, then let's move forward with that. If you prefer your pure Harmony solution I can live with that very well - feel free to merge it to master however you see fit.

Thanks


bodewig:
I have rebased the branch against master and close the PR. This can be revisited later if anybody is interested.

PR #214, opened 2021-08-03
[COMPRESS-584] Fix IOUtils.readRange() can read more from a channel than asked for

matthijsln:
See https://issues.apache.org/jira/browse/COMPRESS-584

### DIFF
--- a/src/test/java/org/apache/commons/compress/utils/IOUtilsTest.java
+++ b/src/test/java/org/apache/commons/compress/utils/IOUtilsTest.java
@@ -23,6 +23,7 @@ import java.io.EOFException;
 import java.io.FilterInputStream;
 import java.io.InputStream;
 import java.io.IOException;
+import java.lang.reflect.Field;
 import java.nio.ByteBuffer;
 import java.nio.channels.ReadableByteChannel;
 
@@ -145,6 +146,43 @@ public class IOUtilsTest {
             final ByteBuffer b = ByteBuffer.allocate(1);
             Assert.assertEquals(1, in.read(b));
             Assert.assertArrayEquals(new byte[] { 4 }, b.array());
+        }
+    }
+
+    @Test
+    public void readRangeFromChannelDoesntReadMoreThanAskedForWhenItGotLessInFirstReadCall() throws IOException {
+        try (ReadableByteChannel in = new SeekableInMemoryByteChannel(new byte[] { 1, 2, 3, 4, 5, 6, 7 }) {
+            @Override
+            public int read(ByteBuffer buf) throws IOException {
+                // Trickle max 2 bytes at a time to trigger COMPRESS-584
+                final ByteBuffer temp = ByteBuffer.allocate(Math.min(2, buf.remaining()));
+                final int read = super.read(temp);
+                if (read > 0) {
+                    buf.put(temp.array(), 0, read);
+                }
+                return read;
+            }
+        }) {
+            final byte[] read = IOUtils.readRange(in, 5);
+            Assert.assertArrayEquals(new byte[] { 1, 2, 3, 4, 5 }, read);
+        }
+    }
+
+    @Test
+    public void readRangeMoreThanCopyBufferSize() throws Exception {
+        final Field COPY_BUF_SIZE = IOUtils.class.getDeclaredField("COPY_BUF_SIZE");
+        COPY_BUF_SIZE.setAccessible(true);
+        final int copyBufSize = (int)COPY_BUF_SIZE.get(null);
+
+        // Make an input that requires two read loops to trigger COMPRESS-585
+        final byte[] input = new byte[copyBufSize + 10];
+
+        try (SeekableInMemoryByteChannel in = new SeekableInMemoryByteChannel(input)) {
+            // Ask for less than the input length, but more than the buffer size
+            final int toRead = copyBufSize + 1;
+            final byte[] read = IOUtils.readRange(in, toRead);
+            Assert.assertEquals(toRead, read.length);
+            Assert.assertEquals(toRead, in.position());
         }
     }
 
--- a/src/main/java/org/apache/commons/compress/utils/IOUtils.java
+++ b/src/main/java/org/apache/commons/compress/utils/IOUtils.java
@@ -367,6 +367,8 @@ public final class IOUtils {
         final ByteBuffer b = ByteBuffer.allocate(Math.min(len, COPY_BUF_SIZE));
         int read = 0;
         while (read < len) {
+            // Make sure we never read more than len bytes
+            b.limit(Math.min(len - read, b.capacity()));
             final int readNow = input.read(b);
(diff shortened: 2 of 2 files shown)


PeterAlfredLee:
Good catch!

This looks good to me.
Thanks for your contributions.

src/test/java/org/apache/commons/compress/utils/IOUtilsTest.java
```diff
@@ -148,6 +148,25 @@ public void readRangeFromChannelDoesntReadMoreThanAskedFor() throws IOException
         }
     }
 
+    @Test
+    public void readRangeFromChannelDoesntReadMoreThanAskedForWhenItGotLessInFirstReadCall() throws IOException {
+        try (ReadableByteChannel in = new SeekableInMemoryByteChannel(new byte[] { 1, 2, 3, 4, 5 }) {
+            @Override
+            public int read(ByteBuffer buf) throws IOException {
+                // Read max 3 bytes at a time
+                ByteBuffer temp = ByteBuffer.allocate(Math.min(3, buf.remaining()));
```
PeterAlfredLee, line 157: I think we can have a `final` here.
Even through this is just a test, it's always appreciated if `final` is always used if possible.

src/test/java/org/apache/commons/compress/utils/IOUtilsTest.java
```diff
@@ -148,6 +148,25 @@ public void readRangeFromChannelDoesntReadMoreThanAskedFor() throws IOException
...
+                int read = super.read(temp);
+                if (read > 0) {
+                    buf.put(temp.array(), 0, read);
+                }
+                return read;
+            }
+        }) {
+            byte[] read = IOUtils.readRange(in, 4);
```
PeterAlfredLee, line 165: Same as above

PeterAlfredLee (approved):

matthijsln:
Unfortunately my original fix had a bug when looping more than 2 times. Revisiting this method I came up with a much simpler implementation that should be better. I can squash the previous version and force-push if this looks better (or it can be squash-merged).

PeterAlfredLee:
The new implemention also has some problems :
1. The test `readRangeFromChannelStopsIfThereIsNothingToReadAnymore` is failed
2. When the `len` of bytes to read is bigger than `COPY_BUF_SIZE`, there are some problems. You can reproduce this by simply set the `COPY_BUF_SIZE` to a small value.

matthijsln:
> Unfortunately my original fix had a bug when looping more than 2 times. Revisiting this method I came up with a much simpler implementation that should be better. I can squash the previous version and force-push if this looks better (or it can be squash-merged).

Third time's a charm...

PeterAlfredLee:
Hi @matthijsln
Please have a look at my reply :-)

> When the len of bytes to read is bigger than COPY_BUF_SIZE, there are some problems. You can reproduce this by simply set the COPY_BUF_SIZE to a small value.

matthijsln:
> Hi @matthijsln
> Please have a look at my reply :-)
> 
> > When the len of bytes to read is bigger than COPY_BUF_SIZE, there are some problems. You can reproduce this by simply set the COPY_BUF_SIZE to a small value.

Indeed it did not handle a small COPY_BUF_SIZE correctly. If I set it to 1, the test cases fail because they do not expect it, but with the e50dedf version it never reads more than COPY_BUF_SIZE.

Note that in [COMPRESS-585](https://issues.apache.org/jira/browse/COMPRESS-585) I propose limiting the number of bytes read to COPY_BUF_SIZE altogether (the same as the `IOUtils.readRange(InputStream,int)` method) because the calling code does not expect the method to return less than `len` bytes unless EOF is encountered.

matthijsln:
This would work without a copy buffer:
```
    public static byte[] readRange(final ReadableByteChannel input, final int len) throws IOException {
        final ByteBuffer b = ByteBuffer.allocate(len);
        while (b.hasRemaining()) {
            if (input.read(b) <= 0) {
                break;
            }
        }
        return Arrays.copyOfRange(b.array(), 0, b.position());
    }
```

With a copy buffer:

```
    public static byte[] readRange(final ReadableByteChannel input, final int len) throws IOException {
        final ByteArrayOutputStream output = new ByteArrayOutputStream();
        final ByteBuffer b = ByteBuffer.allocate(Math.min(len, COPY_BUF_SIZE));
        int read = 0;
        while (read < len) {
            // Make sure we never read more than len bytes
            b.limit(Math.min(len - read, b.remaining()));
            final int readNow = input.read(b);
            if (readNow <= 0) {
                break;
            }
            output.write(b.array(), 0, readNow);
            b.rewind();
            read += readNow;
        }
        return output.toByteArray();
    }
```

Additional test case for reading more than the copy buffer length (of course COPY_BUF_SIZE can be made more accessible instead of using reflection):
```
    @Test
    public void readRangeMoreThanCopyBufferSize() throws Exception {
        final Field COPY_BUF_SIZE = IOUtils.class.getDeclaredField("COPY_BUF_SIZE");
        COPY_BUF_SIZE.setAccessible(true);
        final int copyBufSize = (int)COPY_BUF_SIZE.get(null);

        // Make an input that requires two read loops
        byte[] input = new byte[copyBufSize + 10];

        try (SeekableByteChannel in = new SeekableInMemoryByteChannel(input)) {
            // Ask for less than the input length, but more than the buffer size
            int toRead = copyBufSize + 1;
            byte[] read = IOUtils.readRange(in, toRead);
            assertEquals(toRead, read.length);
            assertEquals(toRead, in.position());
        }
    }
```

Hope this is correct :) What version is preferable? I don't think a copy buffer in itself is needed.

PeterAlfredLee:
I prefer the one with a copy buffer.

I'm not professional in IO but the implemention with buffer is pretty common in IO implementions.

matthijsln:
Wrapping a byte array in a `ByteBuffer` is the most efficient. In the following implementation an array copy is only needed when hitting EOF:

```
    public static byte[] readRange(final ReadableByteChannel input, final int len) throws IOException {
        final byte[] out = new byte[len];
        final ByteBuffer b = ByteBuffer.wrap(out);
        while (b.hasRemaining()) {
            if (input.read(b) <= 0) {
                break;
            }
        }
        if (b.hasRemaining()) {
            // We hit EOF and read less bytes, return a smaller array than len
            return Arrays.copyOf(b.array(), b.position());
        } else {
            return out;
        }
    }

PeterAlfredLee:
This current implemention looks good to me now.

For the "no copy buffer" version, it looks more efficient. But I'm not quite sure about this. I will try to make some benchmark tests if I got time. And I'd like to hear ideas from someone else.

PeterAlfredLee:
Hi @matthijsln 

I want to merge this PR first. The current implementation didn't change much from original implemention.

I will investigate into the "no copy buffer" implementation later when I can curve out some time.

Thank you for your contributions!

PR #199, opened 2021-05-23
Whitespace improvement, site nits and assorted typos

HelderMagalhaes:
Closed PR 197 and PR 198, as those changes were also included here by mistake.

PR #197, opened 2021-05-23
Whitespace improvement

HelderMagalhaes:
tabs to spaces; reduce padding (1 space on the longest member) for consistency with other structures.

_The tabs vs. spaces mix does lead to unexpected visual formatting (only OK in a properly configured text editor)_ (split from [PR 192](https://github.com/apache/commons-compress/pull/192))
image

PR #217, opened 2021-08-21
Documentation nits

HelderMagalhaes:
Redid #199 (and implicitly #198 and #197) from scratch:
 * Now "rebased" on master as hinted by @garydgregory (on #199)
 * More squashed than previous as advised by @PeterAlfredLee (on #199)
 * A separate commit for whitespace as suggested by @bodewig ([follow-up](https://github.com/apache/commons-compress/pull/192#issuecomment-846393334) from #192)


HelderMagalhaes:
Added a last-minute found batch of spelling fixes. Not squashed with previous "spelling and grammar check" but, given the measurable number of entries, I felt like it was better being shipped in this PR than creating a separate one. 🙂

src/site/site.xml
```diff
@@ -56,7 +57,6 @@
       <item name="1.8" href="javadocs/api-1.8/index.html"/>
```
garydgregory, line 57: Instead of keep this list, why not point to javadoc.io like we do here https://commons.apache.org/proper/commons-io/ ?

HelderMagalhaes: > Instead of keeping this list, why not point to javadoc.io like we do here https://commons.apache.org/proper/commons-io/ ?

Sounds like a good idea, but possibly a bit off-topic for specific changes as these ones: according to I've seen in _commons-io_, it seems that the (commons-compress) download page would also need some work in order to list previous versions, together with the javadoc links. Could I suggest creating a follow-up on this? 😉

src/site/xdoc/tar.xml
```diff
@@ -129,14 +129,14 @@
         <p>Prior to Commons Compress 1.20 <code>TarArchiveInputStream</code> would recognize sparse
         file entries stored using the "oldgnu" format
         (<code>-&#x2d;sparse-version=0.0</code> in GNU tar) but not
-        able to extract them correctly. Starting with Commons Compress
+        able to extract them correctly. Starting with Commons Compress 1.21
```
PeterAlfredLee, line 132: I think this should be 1.20. 
#87 introduced this feature and it's included in 1.20

PeterAlfredLee (approved):

PeterAlfredLee:
Looks good to me now.

src/site/site.xml
HelderMagalhaes: > Could I suggest creating a follow-up on this? 😉

I've created the follow-up: [COMPRESS-593](https://issues.apache.org/jira/projects/COMPRESS/issues/COMPRESS-593)
Please add any relevant suggestions there. 🙂

PR #223, opened 2021-09-22
COMPRESS-591: Fix decoding of 7z files containing LZMA streams with end

PeterAlfredLee:
Call `enableRelaxedEndCondition()` to fix decoding of 7z files containing LZMA streams with end marker. 
Thanks for `Lasse Collin's` reporting and patch.

See also [here](https://issues.apache.org/jira/browse/COMPRESS-591?page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel&focusedCommentId=17416947#comment-17416947)


src/test/java/org/apache/commons/compress/archivers/sevenz/SevenZFileTest.java
```diff
@@ -756,6 +756,11 @@ public void retrieveInputStreamForShuffledEntries() throws IOException {
         }
     }
 
+    @Test
+    public void testSevenZWithEOS() throws IOException {
+        new SevenZFile(getFile("lzma-with-eos.7z"));
```
garydgregory, line 761: Resource leak? Use try-with-resources block?

PeterAlfredLee: Fixed.

PR #242, opened 2022-02-01
bzip2: calculate median-of-3 on unsigned values

peterdettman:
The existing `BlockSort#med3`  function is comparing byte values, which are signed. This seems off to me, since the main calling loop appears to be handling these bytes as (non-negative) 8-bit ints. Also the original C code appears to use unsigned char values (e.g. https://github.com/centricular/bzip2/blob/35e7280360320787d09a2419e0f5a1863feb8a34/blocksort.c#L583).

This PR changes `BlockSort#med3` to accept int arguments and the caller to pass non-negative 8-bit values. I'm not really familiar enough with bzip2 to judge the impact, or to select a sample input for which this might matter. Hopefully someone more knowledgeable can weigh in.

### DIFF
--- a/src/main/java/org/apache/commons/compress/compressors/bzip2/BlockSort.java
+++ b/src/main/java/org/apache/commons/compress/compressors/bzip2/BlockSort.java
@@ -790,7 +790,7 @@ class BlockSort {
         }
     }
 
-    private static byte med3(final byte a, final byte b, final byte c) {
+    private static int med3(final int a, final int b, final int c) {
         return (a < b) ? (b < c ? b : a < c ? c : a) : (b > c ? b : a > c ? c
                                                         : a);
     }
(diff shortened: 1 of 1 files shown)


garydgregory:
Hi @peterdettman 
Thank you for the PR.
Do you see similar issues in the rest of the class, package or codebase? I approved the CI build, so let's make sure is passed. I wonder if this can be tested in any way. Because it will be too easy to change the code back without any test failure.

peterdettman:
I've only really looked at mainQSort3; we have a fork of an older version of this bzip2 code in the BouncyCastle codebase and I refactored the qsort code a bit, then noticed this discrepancy when referring back here.

I don't think median selection affects correctness; it's only a matter of improving performance if you can choose better pivots on average. Even then I'm not sure it affects overall performance that much, although perhaps there is some subclass of inputs where it does, or some worst-case that is avoided.

garydgregory:
@kinow , any thoughts here?

kinow (commented):
@peterdettman thanks for linking the original code. The change looks good to me. The method changed is `private` to that class. Doesn't look correct to use the negative bytes directly in the comparison, so I think it should be OK to make it positive with the bitwise and and that mask. Not sure if we can find some file or test data to verify/test it? Probably a little hard...

Thanks!

kinow (approved):

PR #247, opened 2022-02-08
Use Math.min and Math.max calculations.

arturobernalg:


src/main/java/org/apache/commons/compress/harmony/pack200/BHSDCodec.java
```diff
@@ -308,11 +308,8 @@ public boolean encodes(final long value) {
             }
         } else if (z < 0) {
             // Need to use integer overflow here to represent negatives.
-            if (cardinality < 4294967296L) {
-                z += cardinality;
-            } else {
-                z += 4294967296L; // this value is equal to (1 << 32).
-            }
+            // this value is equal to (1 << 32).
```
kinow, line 311: Wrong comment, I think. `1 << 32` is `4294967296L`.

image

After this change it could be interpreted as `z` being `1 << 32`. Maybe change the text to `4294967296 is 1 << 32` (which helps devs with the magic-numbers issue)

garydgregory: I think you'd want to use jshell here, not python...

kinow: True! I have to get used to jshell. Looks like the integer overflow resets the value to `1`. But using longs works the same way as in Python.

image

src/main/java/org/apache/commons/compress/compressors/bzip2/BlockSort.java
```diff
@@ -877,11 +876,9 @@ private void mainQSort3(final BZip2CompressorOutputStream.Data dataShadow,
                     stack_dd[sp] = d1;
                     sp++;
                 } else {
-                    int n = ((ltLo - lo) < (unLo - ltLo)) ? (ltLo - lo)
-                        : (unLo - ltLo);
+                    int n = Math.min((ltLo - lo), (unLo - ltLo));
```
garydgregory, line 879: No need for extra parens to wrap arguments to the min call.

src/main/java/org/apache/commons/compress/compressors/bzip2/BlockSort.java
```diff
@@ -877,11 +876,9 @@ private void mainQSort3(final BZip2CompressorOutputStream.Data dataShadow,
...
                     vswap(fmap, lo, unLo - n, n);
-                    int m = ((hi - gtHi) < (gtHi - unHi)) ? (hi - gtHi)
-                        : (gtHi - unHi);
+                    int m = Math.min((hi - gtHi), (gtHi - unHi));
```
garydgregory, line 881: No need for extra parens to wrap arguments to the min call.

src/main/java/org/apache/commons/compress/compressors/bzip2/BZip2CompressorOutputStream.java
```diff
@@ -244,10 +244,7 @@ private static void hbMakeCodeLengths(final byte[] len, final int[] freq,
                 final int weight_n2 = weight[n2];
                 weight[nNodes] = ((weight_n1 & 0xffffff00)
                                   + (weight_n2 & 0xffffff00))
-                    | (1 + (((weight_n1 & 0x000000ff)
-                             > (weight_n2 & 0x000000ff))
-                            ? (weight_n1 & 0x000000ff)
-                            : (weight_n2 & 0x000000ff)));
+                    | (1 + (Math.max((weight_n1 & 0x000000ff), (weight_n2 & 0x000000ff))));
```
garydgregory, line 247: No need for extra parens to wrap arguments to the min call.

src/main/java/org/apache/commons/compress/harmony/pack200/BHSDCodec.java
garydgregory: So... now what?

garydgregory (changes requested):
Please see my comments.

src/main/java/org/apache/commons/compress/harmony/pack200/BHSDCodec.java
kinow: I think the comment needs to be updated to say that `4294967296L` is the `1 << 32`, and not `z += Math.min(cardinality, 4294967296L);`.

arturobernalg:
> One comment from @garydgregory pending to be resolved, but rest looks good to me. Thanks @arturobernalg

hi @kinow done
TY

src/main/java/org/apache/commons/compress/archivers/sevenz/SevenZFile.java
```diff
@@ -1361,7 +1361,7 @@ private void sanityCheckFilesInfo(final ByteBuffer header, final ArchiveStatisti
                 }
             }
         }
-        stats.numberOfEntriesWithStream = stats.numberOfEntries - (emptyStreams > 0 ? emptyStreams : 0);
+        stats.numberOfEntriesWithStream = stats.numberOfEntries - (Math.max(emptyStreams, 0));
```
garydgregory, line 1364: Parens around the call to max are not needed.

arturobernalg: done.

PR #248, opened 2022-02-13
Fix thread safety issues when encoding 7z password

Glavo:
`CharsetEncoder` is not thread safe. Due to shared `PASSWORD_ENCODER`, problems may occur when encoding 7z file passwords in parallel. The UTF-16LE encoder rarely changes the state, so problems are rare, but the problem is real: When processing unicode surrogate pairs, the Encoder needs to store the state internally.

`Charset.encode` will cache a thread local encoder, so it is reasonable to use it directly.

### DIFF
--- a/src/main/java/org/apache/commons/compress/archivers/sevenz/SevenZFile.java
+++ b/src/main/java/org/apache/commons/compress/archivers/sevenz/SevenZFile.java
@@ -30,7 +30,6 @@ import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
 import java.nio.CharBuffer;
 import java.nio.channels.SeekableByteChannel;
-import java.nio.charset.CharsetEncoder;
 import java.nio.file.Files;
 import java.nio.file.StandardOpenOption;
 import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


kinow (approved):
Change looks good to me. The [javadocs] for `CharsetEncoder` indeed says it's not thread-safe “ Instances of this class are not safe for use by multiple concurrent threads.”. Thanks!

kinow (commented):
Said the above, looks like the Compress class itself is marked as not thread-safe as well @Glavo ?

See the `@NotThreadSafe` annotation. So we could have other issues besides the `CharsetEncoder` that would have to be fixed.

Glavo:
> Said the above, looks like the Compress class itself is marked as not thread-safe as well @Glavo ?
> 
> See the `@NotThreadSafe` annotation. So we could have other issues besides the `CharsetEncoder` that would have to be fixed.

I checked the 7z source and didn't see a similar problem. 

My goal is not to make the instance of `SevenZFile` thread safe. I have reason to believe that `@NotThreadSafe` indicates that a single instance cannot operate in parallel, and the problem I'm dealing with now is security in the global scope. 

 `utf16Decode` is used to initialize the instance of `SevenZFile`, this is reflected in the fact that errors may occur if multiple `SevenZFile` instances with Unicode extended flat characters in passwords are created in parallel. It doesn't appear on one instance, so I want to solve it.

PeterAlfredLee (approved):

PeterAlfredLee:
This change looks good to me.

PeterAlfredLee:
For the `@NotThreadSafe` annotation :

For SevenZ archives, many entries may be compressed in a single 'folder'(see also [7z format specification](https://7zip.bugaco.com/7zip/7zFormat.txt)). This means we can not read the Nth entry before the (N-1)th entry is read, and that's why we can not achieve a easy enough random access as Zip format did. In this case, we need to read the inputstream carefully: no much or no less bytes should be read, and we can only access the entries in the given order.

I have no idea how we can implement a multi-threading SevenZFile, so I'm not surprised about the `@NotThreadSafe` annotation.

PR #254, opened 2022-03-14
[COMPRESS-612] improve TAR support for file times

andrebrait:
R/W atime and ctime support for XSTAR/XUSTAR/POSIX
R/W high precision (100ns increments) time support for POSIX
Read support for atime and ctime for OLDGNU/GNU
Use FileTime instead of Date to allow for higher precision

Possible breaking change:
`ctime` and `atime` are not part of the _**extra**_ PAX headers anymore. 
This means that, if anyone is using those manually, they won't be able to do that anymore.
A possible solution would be to add them to the extra PAX headers anyway and update their values when setting the corresponding date fields. The other way around (updating the dates when setting the headers) is already working.

Looking for feedback on that so I know if you think that's a problem or not.

I'm als looking for feedback into possibly making BIGNUMBER_POSIX and possibly LONGNAME_POSIX the defaults, or at least having expand most errors related to big numbers to reference BIGNUMBER_POSIX and/or BIGNUMBER_GNU (although that's problematic for some TAR tools).


andrebrait:
I'm also looking into feedback on why I'm getting this breakage. The checker says there's some method that does not throw a checked exception anymore, but I can't seem to find what's causing that. I haven't changed the signature of any methods in that way as far as checked exceptions go.

PeterAlfredLee:
Hi @andrebrait 
Thank you for your contributing.

IIRC there's a japicmp report in generated target directory(/target/japicmp/japicmp.html). You can check it for detailed information.

andrebrait:
@PeterAlfredLee Thanks, I checked it and I found the reason. 
It's because I removed this:

```java
/** Convert millis to seconds */
public static final int MILLIS_PER_SECOND = 1000;
```

I put it back.

src/test/java/org/apache/commons/compress/archivers/tar/TarArchiveEntryTest.java
```diff
@@ -306,6 +308,164 @@ public void getOrderedSparseHeadersRejectsStructsPointingBeyondOutputEntry() thr
...
+        assertEquals("atime", FileTime.from(Instant.parse("2022-03-14T01:31:00.706927200Z")), entry.getLastAccessTime());
+        assertEquals("ctime", FileTime.from(Instant.parse("2022-03-14T01:28:59.700505300Z")), entry.getStatusChangeTime());
+        assertEquals("birthtime", FileTime.from(Instant.parse("2022-03-14T01:29:00.723509000Z")), entry.getCreationTime());
+    }
+
+    @Test
+    public void shouldNotWriteTimePaxHeadersByDefault() throws IOException {
+        TarArchiveEntry entry = createEntryForTimeTests();
+
+        final ByteArrayOutputStream bos = new ByteArrayOutputStream();
```
garydgregory, line 331: Use try-with-resources. I know this is will not cause a real resource leak on a thrown exception this avoids compiler warnings and it is a best practice we want to exemplify.

src/test/java/org/apache/commons/compress/archivers/tar/TarArchiveEntryTest.java
```diff
@@ -306,6 +308,164 @@ public void getOrderedSparseHeadersRejectsStructsPointingBeyondOutputEntry() thr
...
+
+        assertNull("should be at end of file", tis.getNextTarEntry());
+        tis.close();
+    }
+
+    @Test
+    public void shouldWriteTimesForStarMode() throws IOException {
+        TarArchiveEntry entry = createEntryForTimeTests();
+
+        final ByteArrayOutputStream bos = new ByteArrayOutputStream();
```
garydgregory, line 363: Use try-with-resources. I know this is will not cause a real resource leak on a thrown exception this avoids compiler warnings and it is a best practice we want to exemplify.

src/test/java/org/apache/commons/compress/archivers/tar/TarArchiveEntryTest.java
```diff
@@ -306,6 +308,164 @@ public void getOrderedSparseHeadersRejectsStructsPointingBeyondOutputEntry() thr
...
+
+        assertNull("should be at end of file", tis.getNextTarEntry());
+        tis.close();
+    }
+
+    @Test
+    public void shouldWriteTimesAsPaxHeadersForPosixMode() throws IOException {
+        TarArchiveEntry entry = createEntryForTimeTests();
+
+        final ByteArrayOutputStream bos = new ByteArrayOutputStream();
```
garydgregory, line 396: Use try-with-resources. I know this is will not cause a real resource leak on a thrown exception this avoids compiler warnings and it is a best practice we want to exemplify.

src/main/java/org/apache/commons/compress/archivers/tar/TarArchiveOutputStream.java
```diff
@@ -624,11 +635,48 @@ private void addPaxHeaderForBigNumber(final Map<String, String> paxHeaders,
...
+            } else {
+                addInstantPaxHeader(paxHeaders, header, seconds, nanos);
+            }
+        }
+    }
+
+    private void addFileTimePaxHeader(final Map<String, String> paxHeaders,
+        final String header, final FileTime value) {
+        if (value != null) {
+            Instant instant = value.toInstant();
```
garydgregory, line 656: Use final where possible.

src/main/java/org/apache/commons/compress/archivers/tar/TarArchiveOutputStream.java
```diff
@@ -624,11 +635,48 @@ private void addPaxHeaderForBigNumber(final Map<String, String> paxHeaders,
         }
     }
 
+    private void addFileTimePaxHeaderForBigNumber(final Map<String, String> paxHeaders,
+        final String header, final FileTime value,
+        final long maxValue) {
+        if (value != null) {
+            Instant instant = value.toInstant();
```
garydgregory, line 642: Use final where possible.

src/main/java/org/apache/commons/compress/archivers/tar/TarConstants.java
```diff
@@ -227,6 +227,13 @@
      */
     byte   LF_OLDNORM = 0;
 
+    /**
+     * Offset inside the header for the "link flag" field.
+     *
+     * @since 1.22
+     */
+    int    LF_OFFSET = 156;
```
garydgregory, line 235: No need for whitespace IMO.

src/main/java/org/apache/commons/compress/archivers/tar/TarConstants.java
```diff
@@ -347,6 +361,13 @@
      */
     String MAGIC_XSTAR = "tar\0";
 
+    /**
+     * Offset inside the header for the xtar multivolume data
+     *
```
garydgregory, line 366: Is there an RFC or resource for which we can provide a URL for these offsets?

garydgregory (changes requested):
Hi @andrebrait 
Thank you for your PR.
Please see my scattered comments.

src/main/java/org/apache/commons/compress/archivers/tar/TarConstants.java
andrebrait: I'm just using the same style of the surrounding code. It looks better, unless I also remove spaces from everything else.

src/main/java/org/apache/commons/compress/archivers/tar/TarArchiveOutputStream.java
andrebrait: Done

src/main/java/org/apache/commons/compress/archivers/tar/TarArchiveOutputStream.java
andrebrait: Done

src/test/java/org/apache/commons/compress/archivers/tar/TarArchiveEntryTest.java
andrebrait: Done

andrebrait: Btw, I can maybe add try-with-resouces on this ByteArrayOutputStream, but the TarArchiveOutputStream's close methd already closes the BAOS as well. If I do add the BAOS to the try-with-resources, I'll have to create a byte array to save its internal byte array after closing the TarArchiveOutputStream, meaning a nested try-with-resources. Let me know what you prefer, as both are ok examples because of how TarArchiveOutputStream works.

src/test/java/org/apache/commons/compress/archivers/tar/TarArchiveEntryTest.java
andrebrait: Done

src/test/java/org/apache/commons/compress/archivers/tar/TarArchiveEntryTest.java
andrebrait: Done

src/main/java/org/apache/commons/compress/archivers/tar/TarConstants.java
andrebrait: I derived it from [star's source code](https://sourceforge.net/projects/s-tar/files/star-1.6.tar.bz2). Specifically, `star.h`, line 400, which you can also view [here](https://fossies.org/linux/misc/star-1.6.tar.bz2/star-1.6/star/star.h).

I also added the struct and the offsets at the javadoc for `TarArchiveEntry`. Would you like me to find a way to incorporate a link or pointer to the source code?

garydgregory: If you've already documented well on one place you can just use "@see" to that place.

src/test/java/org/apache/commons/compress/archivers/tar/TarArchiveEntryTest.java
garydgregory: Leave it as is then. I don't think you'd need need nested try-w-res since you can have >1 statement in a try. But it's a detail we do not need to deal with as a requirement for this PR to make it. I'll leave it up to you.

andrebrait: I'd think that would be an antipattern in a way, because I'd have to get the data from the BAOS before both it and the TarArchiveOutputStream are closed if both are on the same try-with-resources, which is generally not a good idea with some OutputStream implementations.

I'd rather have them nested to ensure the TarArchiveOutputStream is closed before getting the data from the BAOS as that's the safest thing to do in most scenarios.

src/main/java/org/apache/commons/compress/archivers/tar/TarConstants.java
andrebrait: I see there's no support for multivolume TAR archives yet. 

I decided to only add the `@see` bits in the offset constants I added and to add a note on the multivolume check for xstar for clarity. The whole section was taken from star anyway, so I trust it is correct. See 07d34a2.

Regardless, the `M` link flag is a GNU/xstar-specific extension that's only used in multivolume archives, which compress currently doesn't support, so that code is there more for documentation/future-proofing than anything else.

andrebrait:
Hi @garydgregory. Let me know if there's anything else you'd like me to change here.

asthagarg2428:
@andrebrait @garydgregory Does apache.compress support creating password protected zip/7zip files ?
I explored https://commons.apache.org/proper/commons-compress/index.html but couldn't find any ,
also is there any limit of 2GB with apache in extraction?

andrebrait:
@asthagarg2428 I don't think posting questions like that on unrelated PRs is a good way to go about things, but let me try to answer as best as I can.

I don't think you can create password-protected zip or 7z archives. I could not find any code for that. Only for _decompressing_ password-protected 7z archives.

There are some examples for 7z here: https://commons.apache.org/proper/commons-compress/examples.html but as you can see, nothing about creating them.

You can see here that it does not support encrypted zip files: https://github.com/apache/commons-compress/blob/82ac0e1f2811aa39ceec0367aac031897947f1ad/src/test/java/org/apache/commons/compress/archivers/zip/EncryptedArchiveTest.java

As for the file size limit, no, there should not be any limit of that kind, as long as you don't try to extract the whole file to memory but use the InputStream methods to read it little by little. I've extracted/read large files with commons-compress using my tools.

andrebrait:
@asthagarg2428 mind you, the 2GB limit only applies to old TAR formats when _creating_ them. If you set the big number mode to either XSTAR or POSIX, it's basically unlimited.

asthagarg2428:
@andrebrait Thanks for your reply.
Actually I was able to decompress a password-project **zip** file using apache-compress library in my Android project although it's mentioned in the ([apache-compress-limitaion](https://commons.apache.org/proper/commons-compress/limitations.html)) that they do not support it So, thought may be apache would have implemented password protection support in archiving also but the documentation is not updated.

But seems, this feature is still not there, have to implement our own logic for supporting password protected compression.

And the Issues tab seems disabled for this project, so had to ask here..apologies for that.

andrebrait:
@asthagarg2428 oh, weird. I can maybe look into that side of documentation, but not sure I'll have the time for that. 

As for compression, I think it's really not implemented.

asthagarg2428:
@andrebrait Yes, even I found it weird when I was able to extract a password-protected zip file using this library which doesn't claim to.

Here's the limitation they have written:-
```
_ZIP
no support for encryption_
```
Anyway, thanks! I will check if I can support password-protected compression also.

PR #284, opened 2022-04-15
Use StringBuilder instead of StringBuffer.

arturobernalg:


XenoAmess:
instead of StringBuffer, not instead of StringBuilder.

arturobernalg:
> instead of StringBuffer, not instead of StringBuilder.

yep. TY @XenoAmess

PR #289, opened 2022-05-06
Fix TarFileTest

darkma773r:
fix TarFileTest.testDirectoryWithLongNameEndsWithSlash


darkma773r:
@garydgregory, this fixes `TarFileTest` on Linux but it's not clear to me if this is the correct fix or if a deeper issue needs to be resolved. The test started failing with the PR for COMPRESS-612 and the extended logic for detecting the XSTAR format [here](https://github.com/apache/commons-compress/blob/afa14ea0fe5c65ab923080ccded72678979563fa/src/main/java/org/apache/commons/compress/archivers/tar/TarArchiveEntry.java#L741). The logic previously evaluated the test tar files as being in POSIX_FORMAT and so appended the '/' character to the end of directory names. It now evaluates to XSTAR and so does not append the '/' character. In order to make the test pass, I changed the requested entry format in the test from BIGNUMBER_STAR to BIGNUMBER_POSIX.

PR #332, opened 2022-11-28
feat: Encyrption support for Seven7

Dougniel:
👉🏼 The purpose is to provide password based encryption for 7z compression in the same way of decryption already supported, so go forward of one know limitation

☝️In this way, I would like to submit my contribution based on the existing implementation of decryption and the C++ implementation of 7z 

✅ I added one unit test
Implementation of password-based encryption for 7z compressor

Let me know what do you think about this implementation 👍

https://issues.apache.org/jira/browse/COMPRESS-633


Dougniel:
I push an alternate implementation to throw out use of `AES/CBC/PKCS5Padding` raised by _Code scanning
/ CodeQL_ as weak

src/main/java/org/apache/commons/compress/utils/ByteUtils.java
```diff
@@ -266,4 +270,17 @@ private static void checkReadLength(final int length) {
             throw new IllegalArgumentException("Can't read more than eight bytes into a long value");
         }
     }
+
+    public static byte[] utf16Decode(final char[] chars) {
```
kinow, line 274: Missing javadoc for this public method. See other public methods for examples.

src/main/java/org/apache/commons/compress/archivers/sevenz/SevenZOutputFile.java
```diff
@@ -93,8 +114,27 @@ public SevenZOutputFile(final File fileName) throws IOException {
...
+     * Prepares channel to write a 7z archive to.
+     *
+     * <p>{@link
+     * org.apache.commons.compress.utils.SeekableInMemoryByteChannel}
+     * allows you to write to an in-memory archive.</p>
+     *
+     * @param channel the channel to write to
+     * @param password optional password if the archive have to be encrypted
+     * @throws IOException if the channel cannot be positioned properly
+     * @since 1.13
```
kinow, line 130: s/1.13/1.23 (see `pom.xml` for the current snapshot, i.e. next release)

src/main/java/org/apache/commons/compress/archivers/sevenz/SevenZOutputFile.java
```diff
@@ -68,17 +72,34 @@ public class SevenZOutputFile implements Closeable {
     private Iterable<? extends SevenZMethodConfiguration> contentMethods =
             Collections.singletonList(new SevenZMethodConfiguration(SevenZMethod.LZMA2));
     private final Map<SevenZArchiveEntry, long[]> additionalSizes = new HashMap<>();
+    private byte[] password;
 
     /**
      * Opens file to write a 7z archive to.
      *
      * @param fileName the file to write to
+     * @param password optional password if the archive have to be encrypted
```
kinow, line 81: s/if the archive have/if the archive has

src/test/java/org/apache/commons/compress/archivers/sevenz/SevenZOutputFileTest.java
```diff
@@ -468,6 +473,40 @@ public void testArchiveWithMixedMethods() throws Exception {
         }
     }
 
+    /**
+     * Test password-baed encryption
+     * 
+     * As AES/CBC Cipher requires a minimum of 16 bytes file data to be encrypted, some padding logic have been implemented.
```
kinow, line 479: I think it needs a `<p>` wrapping it to be a separate paragraph.

kinow: And s/some padding logic have/some padding logic has

src/main/java/org/apache/commons/compress/archivers/sevenz/SevenZFile.java
```diff
@@ -2054,19 +2054,6 @@ public String getDefaultName() {
...
-        }
-        final ByteBuffer encoded = UTF_16LE.encode(CharBuffer.wrap(chars));
-        if (encoded.hasArray()) {
-            return encoded.array();
-        }
-        final byte[] e = new byte[encoded.remaining()];
-        encoded.get(e);
-        return e;
-    }
-
```
kinow, line 2069: :+1: I think it makes sense to move this method to `ByteUtils`, thanks!

src/main/java/org/apache/commons/compress/archivers/sevenz/AES256Options.java
```diff
@@ -0,0 +1,34 @@
...
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ */
+package org.apache.commons.compress.archivers.sevenz;
+
+import java.util.Arrays;
+import java.util.Random;
+
+public class AES256Options {
```
kinow, line 23: Missing javadoc for public class. Worth documenting its members (properties, methods) too, whenever possible (something short is fine).

src/main/java/org/apache/commons/compress/archivers/sevenz/AES256Options.java
```diff
@@ -0,0 +1,34 @@
...
+    byte[] password;
+    byte[] salt = new byte[0];
+    byte[] iv = new byte[16];
+    int numCyclesPower = 19;
+
+    public AES256Options(byte[] password) {
+        this.password = password;
```
kinow, line 30: I don't know the Commons Compress API well enough, but I guess it'd be better if the password were never stored in an object in memory or serialized, if possible. This means instead of storing the object in the class, instead trying to design a method `public void decompress(byte[] data, byte[] password) {...}` that never stores the password.

But if other classes are already doing that, then I guess that should be fine. (Didn't have time to check, sorry)

src/main/java/org/apache/commons/compress/archivers/sevenz/AES256Options.java
```diff
@@ -0,0 +1,34 @@
...
+        new Random(Arrays.hashCode(password)).nextBytes(salt);
+        new Random(Arrays.hashCode(password)).nextBytes(iv);
```
kinow, line 32: Do we need to worry about the `Random` instead of `SecureRandom` or some other random class from Commons Rng/etc?

src/main/java/org/apache/commons/compress/archivers/sevenz/AES256SHA256Decoder.java
```diff
@@ -126,4 +115,121 @@ public void close() throws IOException {
...
+        final byte[] aesKeyBytes = sha256Password(opts.password, opts.numCyclesPower, opts.salt);
+        final SecretKey aesKey = new SecretKeySpec(aesKeyBytes, "AES");
+
+        final Cipher cipher;
+        try {
+            cipher = Cipher.getInstance("AES/CBC/NoPadding");
+            cipher.init(Cipher.ENCRYPT_MODE, aesKey, new IvParameterSpec(opts.iv));
+        } catch (final GeneralSecurityException generalSecurityException) {
+            throw new IOException(
+                "Encryption error " + "(do you have the JCE Unlimited Strength Jurisdiction Policy Files installed?)",
```
kinow, line 131: Unnecessary string concatenation?

kinow (commented):
Hi @tcurdt 

Thanks a lot for your contribution.

I normally don't contribute much to Commons Compress, so I just did a quick review, and left some comments. But some of these are not necessarily must-haves (e.g. questions about storing password, random). Take a look at the other ones for misspellings, and javadocs :+1: 

I stopped before taking a look at `AES256SHA256Decoder` as I realized it would take a bit longer to understand the change there.

Cheers
-Bruno

kinow:
(Also needs a JIRA issue for the changelog)

tcurdt:
@kinow It seems like you have tagged the wrong person.

kinow:
> @kinow It seems like you have tagged the wrong person.

Oh, apologies @tcurdt . No idea why GitHub suggested your user. it's normally pretty clever to show the handle of the opener of the issue first at the list. Fixed!

Dougniel:
> (Also needs a JIRA issue for the changelog)

Hi @kinow,

Thank you for your rigorous and interesting review, I will correct these and respond 

The JIRA in description is not right for the purpose of changelog ? (https://issues.apache.org/jira/browse/COMPRESS-633)

Cheers,
Daniel

src/main/java/org/apache/commons/compress/archivers/sevenz/AES256Options.java
Dougniel: Good suggestion, as `SecureRandom` doc says : 
> This class provides a cryptographically strong random number generator (RNG)

👉🏼 So I will switch to it

src/main/java/org/apache/commons/compress/archivers/sevenz/AES256SHA256Decoder.java
Dougniel: Right, a stupid copy/paste that I made. Maybe it was done that way for formatting purposes

src/main/java/org/apache/commons/compress/archivers/sevenz/AES256Options.java
Dougniel: I consider myself not enought skilled in security concerns, but I asked myself about that too when I saw this mechanism that [clears password as soon as possible](https://github.com/apache/commons-compress/blob/0cbe431fe13e40735ccba054bab8fe4a83062b37/src/main/java/org/apache/commons/compress/archivers/sevenz/SevenZFile.java#L386) 🤔.
 
But I have chosen to implement the encryption like decryption [`SevenZFile.java`](https://github.com/apache/commons-compress/blob/0cbe431fe13e40735ccba054bab8fe4a83062b37/src/main/java/org/apache/commons/compress/archivers/sevenz/SevenZFile.java#L102) (that stores password too) for homogeneous reasons 🤷.

By the way, as the password is required to decrypt each file independently (the sevenZ archive is not entirely encrypted - we can access metadata like file names without password - only files are encrypted)_. I don't know how to manage that without storing it somewhere in memory - asking user to fulfill it for each file to compress/decompress ? Even in this way it will be stored inside the Input/OutputStream used by the library user.

👉🏼 What I could try is to store it in less places (I stored it in `AES256Options` as in `SevenZOutputFile`

src/test/java/org/apache/commons/compress/archivers/sevenz/SevenZOutputFileTest.java
```diff
@@ -546,8 +603,16 @@ private Boolean verifyFile(final SevenZFile archive, final int index,
         if (entry.getSize() == 0) {
             return Boolean.FALSE;
         }
-        assertEquals(1, entry.getSize());
-        assertEquals('A', archive.read());
+        assertEquals(size, entry.getSize());
+        
+        byte[] actual = new byte[size];
+        int count = 0;
+        while(count < size) {
```
garydgregory, line 610: Fix formatting `while (`

src/main/java/org/apache/commons/compress/archivers/sevenz/AES256SHA256Decoder.java
```diff
@@ -126,4 +115,121 @@ public void close() throws IOException {
...
+            }
+        };
+    }
+
+    private byte[] sha256Password(final byte[] password, final int numCyclesPower, final byte[] salt) throws IOException {
+        final MessageDigest digest;
+        try {
+            digest = MessageDigest.getInstance("SHA-256");
+        } catch (final NoSuchAlgorithmException noSuchAlgorithmException) {
+            throw new IOException("SHA-256 is unsupported by your Java implementation", noSuchAlgorithmException);
```
garydgregory, line 199: This is not possible since SHA-256 is required by the JRE specification, see https://docs.oracle.com/javase/8/docs/api/java/security/MessageDigest.html, therefore this could be an `IllegalStateException`.

src/main/java/org/apache/commons/compress/utils/ByteUtils.java
```diff
@@ -266,4 +270,17 @@ private static void checkReadLength(final int length) {
...
+        if (chars == null) {
+            return null;
+        }
+        final ByteBuffer encoded = UTF_16LE.encode(CharBuffer.wrap(chars));
```
garydgregory, line 278: Hm, why little-endian?

Dougniel: Hi @garydgregory,

I don't have the answer, but the existing implementation of decryption use this 🤷

Dougniel:
@kinow, @garydgregory,

Thanks for your reviews, I just pushed corrections that you suggested and I changed the place where `Cipher` is initialized in order to not store password in a clear way (following @kinow [advice](https://github.com/apache/commons-compress/pull/332#discussion_r1037484243)), let me know what do you think about that.

With these changes, the implementation between decryption and encryption is less homogeneous. In my opinion the decryption could be aligned in the encryption way for the same concern about password storage, but also to reduce the complexity of [decode](https://github.com/apache/commons-compress/blob/0cbe431fe13e40735ccba054bab8fe4a83062b37/src/main/java/org/apache/commons/compress/archivers/sevenz/AES256SHA256Decoder.java#L36) method _(ex: replace `coder` and `password` parameters with `options` as done in `encode` method). But as theses changes will introduce a lot of changes, maybe it could be easier to do it in another PR

Cheers,
Daniel

src/main/java/org/apache/commons/compress/archivers/sevenz/AES256SHA256Decoder.java
Dougniel: I rework the implementation and now I only use SecureRandom one time

As in Java 8 it's not clear that [SecureRandom is thread safe or not](https://docs.oracle.com/javase/8/docs/api/java/security/SecureRandom.html), I didn't put In a static member _(even if the JavaDocs 9 seems to tell that [it could be ThreadSafe](https://docs.oracle.com/javase/9/docs/api/java/security/SecureRandom.html))_

What do you think ?

kinow:
> > (Also needs a JIRA issue for the changelog)
> 
> Hi @kinow,
> 
> Thank you for your rigorous and interesting review, I will correct these and respond
> 
> The JIRA in description is not right for the purpose of changelog ? (https://issues.apache.org/jira/browse/COMPRESS-633)
> 
> Cheers, Daniel

Hi Daniel,\\My bad, I hadn't seen that there was already a JIRA created for this issue. Thanks for updating the PR!

src/main/java/org/apache/commons/compress/archivers/sevenz/AES256SHA256Decoder.java
```diff
@@ -126,4 +115,109 @@ public void close() throws IOException {
...
+                    break;
+                }
+            }
+        }
+        return digest.digest();
+    }
+
+    @Override
+    byte[] getOptionsAsProperties(Object options) throws IOException {
+        AES256Options opts = (AES256Options) options;
```
garydgregory, line 206: Use final where you can.

src/main/java/org/apache/commons/compress/archivers/sevenz/SevenZFile.java
```diff
@@ -18,6 +18,7 @@
 package org.apache.commons.compress.archivers.sevenz;
 
 import static java.nio.charset.StandardCharsets.UTF_16LE;
+import static org.apache.commons.compress.utils.ByteUtils.utf16Decode;
```
garydgregory, line 21: I think we only use this style of method references for assertions in tests.

src/main/java/org/apache/commons/compress/utils/ByteUtils.java
```diff
@@ -266,4 +270,24 @@ private static void checkReadLength(final int length) {
             throw new IllegalArgumentException("Can't read more than eight bytes into a long value");
         }
     }
+
```
garydgregory, line 273: It looks like this method is only used from the `sevenz` package so let's keep it package-private there and avoid increasing the public API surface.

Dougniel: Right, I move it on `AES256SHA256Decoder` because it is used to convert password from char to byte for encryption/decryption

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "java.configuration.updateBuildConfiguration": "automatic"
+}
```
kinow, line 3: This file does not need to be added. I believe the repository does not include anything specific to Eclipse/IntelliJ/etc. If you use this file locally, you can add a global `.gitignore` to prevent Git from including it with your changes (or just ignore it when committing your changes).

kinow (changes requested):
Just one comment about the VSCode files. But everything else looks OK to me. I would need to find some spare time now to sit down, read some of the surrounding API in Compress, and then really understand what's going on in order to do a proper review :+1: Thanks!

.vscode/settings.json
Dougniel: Right, It was a mistake 👍

BTW, there is a `.idea` directory pushed before, so I will remove it too as it seems a mistake because there is a `.idea` on `.gitignore`

garydgregory:
Just rebase on git master to test with the latest, that will help us get closer ;-)

src/main/java/org/apache/commons/compress/archivers/sevenz/AES256Options.java
```diff
@@ -0,0 +1,100 @@
...
+ * @see AES256SHA256Decoder
+ */
+public class AES256Options {
+
+    private final byte[] salt;
+    private final byte[] iv;
+    private final int numCyclesPower;
+    private final Cipher cipher;
+
+    /**
```
garydgregory, line 41: Please complete the Javadoc comments. You need a starting sentence.

garydgregory:
@Dougniel 
Thank you for your work and patience. Please verify.

Dougniel:
> @Dougniel
> Merged. Thank you for your work and patience. Please verify.

Thank you for your rigor, all is ok 👍

I noticed that you switched `AES256Options` to package-private, I hesitated because, as public, it offers more possibilities to advanced uses (per file Initialization Vector, per file password.. when setting it at `SevenZArchiveEntry` level). But this increases the public API surface and is not documented

garydgregory:
@Dougniel 
Right. We can always document it and make it public later. But also then make sure the API is right.

Dougniel:
Hi @garydgregory, an idea of next release date ?

garydgregory:
Nope, we have other components to work through and Commons Net just had a release.

PR #334, opened 2022-12-07
COMPRESS-621: Fix calculation the offset of the first zip central directory entry

Glavo:
Since the zip central directory is at the file end, we can splice other content before the zip file.

As an example, in Java 9 and above, the `$JAVA_HOME/jmods` folder contains some files with the extension `jmod`.
The jmod file is a normal zip file, but the file starts with the four-byte magic number `0x4A 0x4D 0x01 0x00`.

Another example, [Ammonite Shell](http://ammonite.io/) splices a bash shell script before the jar file, looks for Java through the bash shell script and starts itself.

`java.util.zip.ZipFile` reads these files correctly, but `org.apache.commons.compress.archivers.zip.ZipFile` recognizes them as empty zip files.

This PR fixes this issue and allows users to read content before the zip body.

I didn't implement it for zip64 because `java.util.zip.ZipFile` didn't implement it either, and I'm not sure how it should be implemented correctly.

(I seem to have missed the release date of commons-compress 1.22. This problem has a wider impact, and I hope to fix it as soon as possible.)


Glavo:
@PeterAlfredLee Can you review this PR?

src/main/java/org/apache/commons/compress/archivers/zip/ZipFile.java
```diff
@@ -163,6 +163,7 @@ public class ZipFile implements Closeable {
 
     private long centralDirectoryStartDiskNumber, centralDirectoryStartRelativeOffset;
     private long centralDirectoryStartOffset;
+    private long firstLocalFileHeaderOffset = 0L;
```
garydgregory, line 166: No need to initialize to default value.

src/main/java/org/apache/commons/compress/archivers/zip/ZipFile.java
```diff
@@ -720,6 +721,28 @@ public String getUnixSymlink(final ZipArchiveEntry entry) throws IOException {
         return null;
     }
 
+    /**
+     * Offset of the first local file header in the file.
```
garydgregory, line 725: Javadoc for getters should start with "Gets..."

src/main/java/org/apache/commons/compress/archivers/zip/ZipFile.java
```diff
@@ -720,6 +721,28 @@ public String getUnixSymlink(final ZipArchiveEntry entry) throws IOException {
...
+    }
+
+    /**
+     * Returns an InputStream for reading the content before the first local file header.
+     *
+     * @return null if there is no content before the first local file header.
+     * Otherwise returns a stream to read the content before the first local file header.
+     * @since 1.23
+     */
+    public InputStream getContentBeforeFirstLocalFileHeader() {
```
garydgregory, line 741: See above.

garydgregory (commented):
See comments for future PRs.

PR #344, opened 2022-12-26
Replace JUnit 3 with JUnit 5

theobisproject:
Most tests are replaced with simple JUnit 5 tests. A small number of tests were migrated to parameterized tests.

### DIFF
--- a/src/test/java/org/apache/commons/compress/archivers/LongPathTest.java
+++ b/src/test/java/org/apache/commons/compress/archivers/LongPathTest.java
@@ -41,8 +41,6 @@ import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.Arguments;
 import org.junit.jupiter.params.provider.MethodSource;
 
-import junit.framework.AssertionFailedError;
-
 /**
  * Test that can read various tar file examples.
  *
@@ -142,7 +140,7 @@ public class LongPathTest extends AbstractTestCase {
         }
         try {
             checkArchiveContent(ais, expected);
-        } catch (final AssertionFailedError e) {
+        } catch (final Exception e) {
             fail("Error processing "+file.getName()+" "+e);
         } finally {
             ais.close();
--- a/src/test/java/org/apache/commons/compress/archivers/LongPathTest.java
+++ b/src/test/java/org/apache/commons/compress/archivers/LongPathTest.java
@@ -20,6 +20,7 @@ package org.apache.commons.compress.archivers;
 
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
 
 import java.io.BufferedInputStream;
 import java.io.BufferedReader;
@@ -139,9 +140,7 @@ public class LongPathTest extends AbstractTestCase {
             fail("Unexpected file type: "+name);
         }
         try {
-            checkArchiveContent(ais, expected);
-        } catch (final Exception e) {
-            fail("Error processing "+file.getName()+" "+e);
+            assertDoesNotThrow(() -> checkArchiveContent(ais, expected), "Error processing " + file.getName());
         } finally {
             ais.close();
         }
--- a/src/test/java/org/apache/commons/compress/harmony/pack200/tests/ArchiveTest.java
+++ b/src/test/java/org/apache/commons/compress/harmony/pack200/tests/ArchiveTest.java
@@ -16,6 +16,9 @@
  */
 package org.apache.commons.compress.harmony.pack200.tests;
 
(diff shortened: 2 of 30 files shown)


src/test/java/org/apache/commons/compress/archivers/LongPathTest.java
```diff
@@ -142,7 +140,7 @@ public void testArchive(final File file) throws Exception {
         }
         try {
             checkArchiveContent(ais, expected);
-        } catch (final AssertionFailedError e) {
+        } catch (final Exception e) {
```
garydgregory, line 143: -1: Porting to JUnit 5 means using `Assert.assertThrows()`, not whatever this is.

theobisproject: I'm with you on this one (and the one in the LongSymLinkTest). Just did not want to do too much in one go.

If I understand the test correctly we can get rid of the try-catch block completely since we do not want the check to throw an exception.

garydgregory:
Hi @theobisproject 
Thank you for pitching in!
No matter which way you slice it, it's going to be a large PR. The only way I can think of making it smaller is by doing it one package at a time. Whichever way you go, I'd rather see it all done the "JUnit 5 way" rather than something in between.

theobisproject:
@garydgregory I will go through the classes and do some more migrations. Is there anything else you would like to have changed (like using try-with-resources, ...)?

garydgregory:
Keep it focused on the one task please: Junit 5 😉

theobisproject:
I have replaced all the try-catch-assert patterns in the tests I migrated. Also more tests do now use the parameterized tests for execution.

src/test/java/org/apache/commons/compress/harmony/pack200/tests/ArchiveTest.java
```diff
@@ -25,20 +28,27 @@
...
 import org.apache.commons.compress.harmony.unpack200.Segment;
 
-import junit.framework.TestCase;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
 
-public class ArchiveTest extends TestCase {
+class ArchiveTest {
```
garydgregory, line 51: There is no need to change the visibility of elements, it just overloads the PR with noise for no gain. Yes, I know that JUnit 5 like this style, but that is not the style we use in most of not all of Conmons components. It just make the PR harder and longer to review. What matters is the use of the Junit API, not the style IMO. For my money, the fact that a test is public tells me it's important, not some internal gadget, so let's not change access unless required (which should not happen).

src/test/java/org/apache/commons/compress/harmony/pack200/tests/ArchiveTest.java
```diff
@@ -95,11 +104,12 @@ private void compareJarEntries(final JarFile jarFile, final JarFile jarFile2) {
...
             String name = entry.getName();
             JarEntry entry2 = jarFile2.getJarEntry(name);
-            assertNotNull("Missing Entry: " + name, entry2);
+            assertNotNull(entry2, "Missing Entry: " + name);
         }
     }
 
-    public void testAlternativeConstructor() throws IOException, URISyntaxException, Pack200Exception {
+    @Test
+    void testAlternativeConstructor() throws IOException, URISyntaxException, Pack200Exception {
```
garydgregory, line 112: See above.

src/test/java/org/apache/commons/compress/harmony/pack200/tests/ArchiveTest.java
```diff
@@ -111,7 +121,8 @@ public void testAlternativeConstructor() throws IOException, URISyntaxException,
         out.close();
     }
 
-    public void testAnnotations() throws IOException, Pack200Exception,
+    @Test
+    void testAnnotations() throws IOException, Pack200Exception,
```
garydgregory, line 125: See above.

src/test/java/org/apache/commons/compress/harmony/pack200/tests/ArchiveTest.java
```diff
@@ -140,7 +151,8 @@ public void testAnnotations() throws IOException, Pack200Exception,
         compareFiles(jarFile, jarFile2);
     }
 
-    public void testAnnotations2() throws IOException, Pack200Exception,
+    @Test
+    void testAnnotations2() throws IOException, Pack200Exception,
```
garydgregory, line 155: And so on, so I'll stop here.

garydgregory (changes requested):
See comments. TY!

theobisproject:
I have removed the visibility change of the classes and methods from the commits.

PR #345, opened 2022-12-28
COMPRESS-613: Support for extra time data in Zip archives

andrebrait:
https://issues.apache.org/jira/browse/COMPRESS-613

This adds support for extra time data (InfoZip and NTFS Extra fields) to Zip archives.
This supports both reading and writing to these fields automatically, including when creating a ZipArchiveEntry from an existing file on the disk, with proper fall backs.

Additionally:

1. Works around a bug involving Integer Overflow in Java 8 and Zip archives: https://bugs.openjdk.org/browse/JDK-8130914
2. Consolidates a few more time conversions into TimeUtils
3. Works around an oversight in Java 8 where the NTFS fields on Zip archives are only read to the precision of microseconds instead of their maximum of 100ns.

Currently, it's missing a few test cases for reading/writing, though I have tested them myself. Submitting for this review because the code itself is likely final (and there are enough tests to cover most of them, and I'll submit the rest of the test cases in the coming days).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -200,7 +200,13 @@ Brotli, Zstandard and ar, cpio, jar, tar, zip, dump, 7z, arj.
       <artifactId>slf4j-api</artifactId>
       <version>${slf4j.version}</version>
       <scope>test</scope>
-    </dependency>    
+    </dependency>
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-simple</artifactId>
+      <version>${slf4j.version}</version>
+      <scope>test</scope>
+    </dependency>
     <dependency>
       <groupId>commons-io</groupId>
       <artifactId>commons-io</artifactId>
--- a/src/main/java/org/apache/commons/compress/archivers/zip/X000A_NTFS.java
+++ b/src/main/java/org/apache/commons/compress/archivers/zip/X000A_NTFS.java
@@ -91,7 +91,6 @@ public class X000A_NTFS implements ZipExtraField {
         }
         return new ZipEightByteInteger(TimeUtils.toNtfsTime(time));
     }
-
     private static Date zipToDate(final ZipEightByteInteger z) {
         if (z == null || ZipEightByteInteger.ZERO.equals(z)) {
             return null;
@@ -105,12 +104,12 @@ public class X000A_NTFS implements ZipExtraField {
         }
         return TimeUtils.ntfsTimeToFileTime(z.getLongValue());
     }
+
     private ZipEightByteInteger modifyTime = ZipEightByteInteger.ZERO;
 
     private ZipEightByteInteger accessTime = ZipEightByteInteger.ZERO;
 
     private ZipEightByteInteger createTime = ZipEightByteInteger.ZERO;
-
 
     @Override
     public boolean equals(final Object o) {
--- a/src/main/java/org/apache/commons/compress/archivers/zip/X000A_NTFS.java
+++ b/src/main/java/org/apache/commons/compress/archivers/zip/X000A_NTFS.java
@@ -68,6 +68,14 @@ import org.apache.commons.compress.utils.TimeUtils;
  * @NotThreadSafe
  */
 public class X000A_NTFS implements ZipExtraField {
+
+    /**
+     * The header ID for this extra field.
+     *
+     * @since 1.23
+     */
+    public static final ZipShort HEADER_ID = new ZipShort(0x000a);
+
     private static final ZipShort TIME_ATTR_TAG = new ZipShort(0x0001);
     private static final ZipShort TIME_ATTR_SIZE = new ZipShort(3 * 8);
 
@@ -83,6 +91,7 @@ public class X000A_NTFS implements ZipExtraField {
         }
         return new ZipEightByteInteger(TimeUtils.toNtfsTime(time));
     }
+
     private static Date zipToDate(final ZipEightByteInteger z) {
         if (z == null || ZipEightByteInteger.ZERO.equals(z)) {
             return null;
@@ -96,12 +105,12 @@ public class X000A_NTFS implements ZipExtraField {
         }
         return TimeUtils.ntfsTimeToFileTime(z.getLongValue());
     }
-
     private ZipEightByteInteger modifyTime = ZipEightByteInteger.ZERO;
 
     private ZipEightByteInteger accessTime = ZipEightByteInteger.ZERO;
 
     private ZipEightByteInteger createTime = ZipEightByteInteger.ZERO;
+
 
     @Override
     public boolean equals(final Object o) {
@@ -114,13 +123,6 @@ public class X000A_NTFS implements ZipExtraField {
         }
         return false;
     }
-
-    /**
-     * The header ID for this extra field.
-     *
-     * @since 1.23
-     */
-    public static final ZipShort HEADER_ID = new ZipShort(0x000a);
 
     /**
      * Gets the access time as a {@link FileTime}
--- a/src/main/java/org/apache/commons/compress/archivers/zip/X000A_NTFS.java
+++ b/src/main/java/org/apache/commons/compress/archivers/zip/X000A_NTFS.java
@@ -68,7 +68,6 @@ import org.apache.commons.compress.utils.TimeUtils;
  * @NotThreadSafe
  */
 public class X000A_NTFS implements ZipExtraField {
-    private static final ZipShort HEADER_ID = new ZipShort(0x000a);
     private static final ZipShort TIME_ATTR_TAG = new ZipShort(0x0001);
     private static final ZipShort TIME_ATTR_SIZE = new ZipShort(3 * 8);
 
@@ -115,6 +114,13 @@ public class X000A_NTFS implements ZipExtraField {
         }
         return false;
     }
+
+    /**
+     * The header ID for this extra field.
+     *
+     * @since 1.23
+     */
+    public static final ZipShort HEADER_ID = new ZipShort(0x000a);
 
     /**
      * Gets the access time as a {@link FileTime}
--- a/src/main/java/org/apache/commons/compress/archivers/zip/X5455_ExtendedTimestamp.java
+++ b/src/main/java/org/apache/commons/compress/archivers/zip/X5455_ExtendedTimestamp.java
@@ -83,8 +83,14 @@ import java.util.zip.ZipException;
  * @since 1.5
  */
 public class X5455_ExtendedTimestamp implements ZipExtraField, Cloneable, Serializable {
-    private static final ZipShort HEADER_ID = new ZipShort(0x5455);
     private static final long serialVersionUID = 1L;
+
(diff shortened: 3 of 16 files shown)


src/main/java/org/apache/commons/compress/archivers/zip/X5455_ExtendedTimestamp.java
```diff
@@ -242,6 +285,20 @@ public Date getCreateJavaTime() {
...
     }
 
+    /**
+     * Returns the create time as a {@link FileTime}
+     * of this zip entry, or null if no such timestamp exists in the zip entry.
+     * The milliseconds are always zeroed out, since the underlying data
+     * offers only per-second precision.
+     *
+     * @return modify time as {@link FileTime} or null.
+     *
```
garydgregory, line 295: Remove extra line b/w Javadoc tags return and since.

src/main/java/org/apache/commons/compress/archivers/zip/X000A_NTFS.java
```diff
@@ -97,13 +105,13 @@ private static FileTime zipToFileTime(final ZipEightByteInteger z) {
         }
         return TimeUtils.ntfsTimeToFileTime(z.getLongValue());
     }
-
```
garydgregory, line 100: Keep this empty line.

src/main/java/org/apache/commons/compress/archivers/zip/X000A_NTFS.java
```diff
@@ -97,13 +105,13 @@ private static FileTime zipToFileTime(final ZipEightByteInteger z) {
...
     private ZipEightByteInteger modifyTime = ZipEightByteInteger.ZERO;
 
     private ZipEightByteInteger accessTime = ZipEightByteInteger.ZERO;
 
     private ZipEightByteInteger createTime = ZipEightByteInteger.ZERO;
 
+
```
garydgregory, line 114: Remove extra empty line.

src/main/java/org/apache/commons/compress/archivers/zip/X5455_ExtendedTimestamp.java
```diff
@@ -130,16 +155,20 @@ private static Date zipLongToDate(final ZipLong unixTime) {
         return unixTime != null ? new Date(unixTime.getIntValue() * 1000L) : null;
     }
 
+    private static FileTime zipLongToFileTime(final ZipLong unixTime) {
```
garydgregory, line 158: The "zipLong" method name prefix is redundant with the argument type `zipLongToFileTime` -> `toFileTime` or probably better `zipLongToFileTime` -> `unixToFileTime`

src/main/java/org/apache/commons/compress/utils/TimeUtils.java
```diff
@@ -46,6 +46,62 @@ public final class TimeUtils {
...
+        return FileTime.from(time, TimeUnit.SECONDS);
+    }
+
+    /**
+     * Converts {@link FileTime} to "standard UNIX time".
+     *
+     * @param time the original FileTime
+     * @return the UNIX timestamp
+     */
+    public static long fileTimeToUnixTime(final FileTime time) {
```
garydgregory, line 65: The method prefix "fileTime" is redundant with the argument type, `fileTimeToUnixTime` -> `toUnixTime`. Since Unix time is always in seconds (as a primitive long here), we should not need to include the scale in the name.

src/main/java/org/apache/commons/compress/utils/TimeUtils.java
```diff
@@ -46,6 +46,62 @@ public final class TimeUtils {
...
+        return time.to(TimeUnit.SECONDS);
+    }
+
+    /**
+     * Converts Java time (milliseconds since Epoch) to "standard UNIX time".
+     *
+     * @param time the original Java time
+     * @return the UNIX timestamp
+     */
+    public static long javaTimeToUnixTime(final long time) {
```
garydgregory, line 75: The method name is good here because the parameter and return type are long.

garydgregory (changes requested):
Hi @andrebrait 
Thank you for the PR. Please see my scattered comments. Rebase on Git master as well SVP.

pom.xml
```diff
@@ -200,7 +200,13 @@ Brotli, Zstandard and ar, cpio, jar, tar, zip, dump, 7z, arj.
       <artifactId>slf4j-api</artifactId>
       <version>${slf4j.version}</version>
       <scope>test</scope>
-    </dependency>    
+    </dependency>
```
garydgregory, line 203: Remove this dep unless required.

andrebrait: Not required, but it's only in test scope and just so we can see any logs produced by the application while running the tests. Without this, they get swallowed.

src/main/java/org/apache/commons/compress/utils/TimeUtils.java
```diff
@@ -46,6 +46,62 @@ public final class TimeUtils {
      */
     static final long WINDOWS_EPOCH_OFFSET = -116444736000000000L;
 
+    /**
+     * Converts "standard UNIX time" (in seconds, UTC/GMT) to {@link FileTime}.
```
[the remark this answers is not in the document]
andrebrait: No idea, TBH. Will change it.

src/main/java/org/apache/commons/compress/utils/TimeUtils.java
```diff
@@ -46,6 +46,62 @@ public final class TimeUtils {
...
+    }
+
+    /**
+     * Checks whether a FileTime exceeds the minimum or maximum for the "standard UNIX time".
+     * If the FileTime is null, this method always returns false.
+     *
+     * @param time the FileTime to evaluate, can be null
+     * @return true if the time exceeds the minimum or maximum UNIX time, false otherwise
+     */
+    public static boolean exceedsUnixTime(final FileTime time) {
```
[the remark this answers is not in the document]
andrebrait: I think it could be negated and called "fitsInUnixTime" or something like that? Or maybe "isUnixTime" (but also negated)

src/main/java/org/apache/commons/compress/archivers/zip/X000A_NTFS.java
```diff
@@ -68,7 +68,14 @@
...
  */
 public class X000A_NTFS implements ZipExtraField {
-    private static final ZipShort HEADER_ID = new ZipShort(0x000a);
+
+    /**
+     * The header ID for this extra field.
+     *
+     * @since 1.23
+     */
+    public static final ZipShort HEADER_ID = new ZipShort(0x000a);
```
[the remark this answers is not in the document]
andrebrait: Will do if it works properly.

The reason for making it public was that I was testing it before with my project that uses commons-compress and I had it public to be able to use it there. I was getting the extra field directly.

But tbh, it makes perfect sense to let it be public if we also have a public method kn ZipArchiveEntry to fetch an ExtraField based on a ZipLong containing its identifier. I know having a small API is important, but should anyone need to fetch that extra field (which is allowed using ZipArchiveEntry) it's quite frustrating, as the user, that I need to maintain a bunch of my own constants, copied from this class, instead of having this class expose this itself.

src/main/java/org/apache/commons/compress/archivers/zip/X000A_NTFS.java
andrebrait: Removed by mistake. I'll take a look at the whole changelog again.

src/main/java/org/apache/commons/compress/archivers/zip/X5455_ExtendedTimestamp.java
andrebrait: I get it, but it became somewhat confusing having all method be overloaded like that. I can change, though. It's up to you.

src/main/java/org/apache/commons/compress/utils/TimeUtils.java
andrebrait: The scale isn't included in the name, is it?

src/main/java/org/apache/commons/compress/archivers/zip/ZipUtil.java
```diff
@@ -30,10 +33,85 @@
...
+     * <li>Seconds: 00000</li>
+     * </ul>
+     *
+     * <p>
+     * This was copied from {@link ZipEntry}.
+     * </p>
+     *
+     * @since 1.23
      */
-    private static final byte[] DOS_TIME_MIN = ZipLong.getBytes(0x00002100L);
```
andrebrait, line 36: @garydgregory note: this was previously the wrong value. We were lucky we never stumbled upon this as a bug.

theobisproject:
I just want to raise awareness that a similar change I have done for the tar format in [COMPRESS-404](https://issues.apache.org/jira/browse/COMPRESS-404) has lead to problems with reproducible builds (see [COMPRESS-583](https://issues.apache.org/jira/browse/COMPRESS-583)). I don't know if something similar could happen here too.

andrebrait:
> I just want to raise awareness that a similar change I have done for the tar format in [COMPRESS-404](https://issues.apache.org/jira/browse/COMPRESS-404) has lead to problems with reproducible builds (see [COMPRESS-583](https://issues.apache.org/jira/browse/COMPRESS-583)). I don't know if something similar could happen here too.

Oh, you mean files built using these? Yes, I know this could do that, but it was already setting modification dates. Access date would change on each access, (provided the filesystem supports it and isn't mounted with e.g. noatime), though.

I guess I could make setting those things optional with a parameter on the constructor or a static constructor for them, but other than that, there isn't that much that can be done if the filesystem gives something different all the time.

andrebrait:
I also made some changes to TAR and 7z lately, around those dates.

Maybe I can revisit all of them and fix COMPRESS-583 as well.

theobisproject:
The "fix" for COMPRESS-583 was to document that the behaviour changed when setting the entry from a file. For me this would be ok but at the same time reproduciblity is no concern for my use cases. But maybe someone else has a different opinion.

andrebrait:
> The "fix" for COMPRESS-583 was to document that the behaviour changed when setting the entry from a file. For me this would be ok but at the same time reproduciblity is no concern for my use cases. But maybe someone else has a different opinion.

Creating the entry from a file is still very convenient. I guess we could come up with a different constructor for when that's desired without setting any of the optional attributes (i.e. the original behavior).

garydgregory:
> > The "fix" for COMPRESS-583 was to document that the behaviour changed when setting the entry from a file. For me this would be ok but at the same time reproduciblity is no concern for my use cases. But maybe someone else has a different opinion.
> 
> Creating the entry from a file is still very convenient. I guess we could come up with a different constructor for when that's desired without setting any of the optional attributes (i.e. the original behavior).

I think we can work on build reproducibility issues in subsequent PRs.

garydgregory:
@andrebrait 
Thank you for your updates, merged.

PR #346, opened 2022-12-29
Use JUnit 5 assertions in JUnit 5 tests instead of JUnit 4 classes

theobisproject:
Only the `OsgiITest` still uses the JUnit 4 test annotation. Migrate all other tests to make use of the JUnit 5 api only.


src/test/java/org/apache/commons/compress/archivers/tar/SparseFilesTest.java
```diff
@@ -119,9 +120,8 @@ private String getTarBinaryHelp() throws IOException {
     }
 
     @Test
+    @DisabledOnOs(OS.WINDOWS)
```
kinow, line 123: :ok_man:

src/test/java/org/apache/commons/compress/compressors/bzip2/PythonTruncatedBzip2Test.java
```diff
@@ -105,7 +105,7 @@ public void testPartialReadTruncatedData() throws IOException {
         buffer = ByteBuffer.allocate(1);
         try {
             bz2Channel.read(buffer);
-            Assert.fail("The read should have thrown.");
+            fail("The read should have thrown.");
```
kinow, line 108: assertThrows here too, I think.

src/test/java/org/apache/commons/compress/compressors/snappy/FramedSnappyCompressorInputStreamTest.java
```diff
@@ -184,14 +184,11 @@ public void testUnskippableChunk() {
...
         };
-        try {
-            final FramedSnappyCompressorInputStream in =
-                new FramedSnappyCompressorInputStream(new ByteArrayInputStream(input));
-            in.read();
-            fail("expected an exception");
-            in.close();
+        try (final FramedSnappyCompressorInputStream in =
+                     new FramedSnappyCompressorInputStream(new ByteArrayInputStream(input))) {
+            IOException exception = assertThrows(IOException.class, () -> in.read());
```
kinow, line 189: :+1: you've already replaced that `fail` by an `assertThrows` here. We should probably change it in the other places too.

src/test/java/org/apache/commons/compress/utils/FixedLengthBlockOutputStreamTest.java
```diff
@@ -208,8 +208,7 @@ public void testPartialWritingThrowsException() {
             fail("Exception for partial write not thrown");
         } catch (final IOException e) {
             final String msg = e.getMessage();
-            assertEquals("exception message",
-                "Failed to write 512 bytes atomically. Only wrote  511", msg);
+            assertEquals("Failed to write 512 bytes atomically. Only wrote  511", msg, "exception message");
```
kinow, line 211: I think this can be rewritten for an assertThrows here too

src/test/java/org/apache/commons/compress/utils/MultiReadOnlySeekableByteChannelTest.java
```diff
@@ -139,44 +143,44 @@ private void check(final byte[] expected, final SeekableByteChannel channel, fin
...
         try {
             channel.read(buf);
-            Assert.fail("expected a ClosedChannelException");
+            fail("expected a ClosedChannelException");
         } catch (final ClosedChannelException expected) {
         }
         try {
             channel.position(100);
-            Assert.fail("expected a ClosedChannelException");
+            fail("expected a ClosedChannelException");
```
kinow, line 183: assertThrows :point_up:

kinow (commented):
Everything looks in order. Thanks for disabling tests on Windows with the annotation.

Left some comments about `assertThrows`. I see you've used it in at least one of the tests updated, but in many others it's still using try/catch/fail, or a similar approach that could be simplified by assertThrows.

Thanks!

src/test/java/org/apache/commons/compress/archivers/tar/SparseFilesTest.java
garydgregory: Why would we disable a test that is passing?

kinow: It's disabled on Windows. The old test had this line removed:

```java
assumeFalse("This test should be ignored on Windows", isOnWindows);
```

src/test/java/org/apache/commons/compress/archivers/TarTestCase.java
```diff
@@ -400,7 +400,7 @@ public void testTarFileCOMPRESS178() throws Exception {
             fail("Expected IOException");
         } catch (final IOException e) {
             final Throwable t = e.getCause();
-            assertTrue("Expected cause = IllegalArgumentException", t instanceof IllegalArgumentException);
+            assertTrue(t instanceof IllegalArgumentException, "Expected cause = IllegalArgumentException");
```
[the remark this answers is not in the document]
theobisproject: I've gone through all tests and replaced the try/catch/fail pattern with assertThrows

src/test/java/org/apache/commons/compress/archivers/tar/SparseFilesTest.java
theobisproject: As @kinow said those tests where already disabled by the assume statements in the tests. Would be nice if you can double check that I didn't mess up the logic 😉

kinow (approved):
Thanks!

kinow:
Leaving unmerged for a while so others can review it too. We just need to squash and merge it if there's nothing wrong. Maybe a JIRA to credit in the changelog too, such a big task :+1: But optional. Thanks!

garydgregory:
Patch applied locally to resolve issues. TY.

PR #357, opened 2023-02-03
Removed dependency on the java.desktop module

axelfontaine:
This PR addresses https://issues.apache.org/jira/browse/COMPRESS-610

It removes commons-compress' dependency on the java.desktop module, allowing for much smaller JVM runtime images for jlink users.

### DIFF
--- a/src/main/java/org/apache/commons/compress/harmony/pack200/Pack200Adapter.java
+++ b/src/main/java/org/apache/commons/compress/harmony/pack200/Pack200Adapter.java
@@ -16,44 +16,16 @@
  */
 package org.apache.commons.compress.harmony.pack200;
 
-import java.beans.PropertyChangeListener;
-import java.beans.PropertyChangeSupport;
 import java.util.SortedMap;
 import java.util.TreeMap;
 
-/**
- * Provides generic JavaBeans support for the Pack/UnpackAdapters
- */
 public abstract class Pack200Adapter {
 
     protected static final int DEFAULT_BUFFER_SIZE = 8192;
 
-    private final PropertyChangeSupport support = new PropertyChangeSupport(this);
-
     private final SortedMap<String, String> properties = new TreeMap<>();
-
-    public void addPropertyChangeListener(final PropertyChangeListener listener) {
-        support.addPropertyChangeListener(listener);
-    }
-
-    /**
-     * Completion between 0..1.
-     *
-     * @param value Completion between 0..1.
-     */
-    protected void completed(final double value) {
-        firePropertyChange("pack.progress", null, String.valueOf((int) (100 * value))); //$NON-NLS-1$
-    }
-
-    protected void firePropertyChange(final String propertyName, final Object oldValue, final Object newValue) {
-        support.firePropertyChange(propertyName, oldValue, newValue);
-    }
 
     public SortedMap<String, String> properties() {
         return properties;
-    }
-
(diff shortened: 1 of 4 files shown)


axelfontaine:
@garydgregory It builds fine. The issue is the removal of the methods being flagged. See https://github.com/apache/commons-compress/actions/runs/4085125229/jobs/7042726675#step:5:782

These methods probably hardly saw any use in the wild and PropertyChangeListener isn't used anywhere else in the codebase. Removing them is essential for being able to remove the overall dependency on the java.desktop module.

rmannibucau:
Dropping them sounds ok but also implies:

* configuring the plugin to ignore this breaking change
* move the version to 2.0-SNAPSHOT (due to the breaking change and related commons policy) and package to `org.apache.commons.compress2`


What can maybe be done is to have the implementation you propose in new classes and keep the existing ones as this and deprecate them. Ideally the existing classes would delegate the impl to the new ones to avoid duplication.
This avoids the breaking change and enables you to cover a desktop-less packaging.

axelfontaine:
I tried to keep the changes to a minimum to make them easy for you to assess.

I'm happy to adjust the PR to configure the plugin to ignore this breaking change and move the version to 2.0-SNAPSHOT.

@rmannibucau Are you sure a tiny change like this requires a namespace change to `org.apache.commons.compress2` ? This sounds like a much bigger impact on users.

rmannibucau:
@axelfontaine not my rule but it is a breaking change so yes

garydgregory:
I am not sure we are ready to move to 2.0 for only this change. A major version is change is a rare opportunity when we can consider what else we want to clean up, change, or add that also needs to break BC.

axelfontaine:
@sebbASF Yes they can. But for them to do so, they must ignore the results of `jdeps`, do the exact same investigation I did and gain enough confidence in their understanding of the issue to come to the same conclusion. Not a great user journey to say the least.

That being said, I see this has no chance of being merged at the moment, so I am going ahead and closing this. The investigation, the changes and their implications are now clear to all involved. Feel free to pick up the changes again in the future when you are ready to move forward with this.

garydgregory:
We could break up this component into a multi maven module maven project, one per compression/archive type.

axelfontaine:
@garydgregory This would be perfect!

Glavo:
> We could break up this component into a multi maven module maven project, one per compression/archive type.

@garydgregory I'm very interested in the proposal of Commons Compress 2. There are some historical legacies in Commons Compress, and we can make it more modern through Commons Compress 2.

I created a fork for Commons Compress: https://github.com/Glavo/kala-compress

In this fork, I made some modifications to make it more modern (there is a more detailed description in the README of this fork):

* Modular support, splitting all compressors and archives into separate JPMS modules;
* Better NIO2 support, migrating all existing `java.io.File` based APIs to `java.nio.file.Path` to fully support non default file systems;
* Use `java.nio.charset.Charset` everywhere instead of using strings to represent encoding. `ZipEncoding` and `CharsetNames` have also been deleted;
* UTF-8 is used as the default encoding in most places.

There are also some minor changes that can be viewed in README for more details.

In addition, I think there are some things worth considering:

* I think the naming of `XxxFile` (such as `ZipFile`) is not appropriate enough. I changed them all to `XxxReader` in fork because I think `XxxFile` should have more functions, such as adding or deleting entries;
* Provide `FileSystem` implementation for each archive format.

I will continue to experiment on this fork, which may help with the future of Commons Compress 2.

PR #360, opened 2023-02-12
pack200: Fix FileBands misusing InputStream#read(byte[])

unascribed:
The Pack200 decompressor expects `read` to always return the amount of bytes it requested. This is *usually* true for raw input streams, such as File or ByteArray, but breaks down when combined with other block-based decompressors. For example, combining an LZMAInputStream with a Pack200 decompressor results in an error similar to the following if a file falls between block boundaries:

```
Exception in thread "main" org.apache.commons.compress.harmony.pack200.Pack200Exception: Expected to read 48873 bytes but read 3274
	at org.apache.commons.compress.harmony.unpack200.FileBands.processFileBits(FileBands.java:87)
	at org.apache.commons.compress.harmony.unpack200.Segment.readSegment(Segment.java:460)
	at org.apache.commons.compress.harmony.unpack200.Segment.unpackRead(Segment.java:514)
	at org.apache.commons.compress.harmony.unpack200.Segment.unpack(Segment.java:484)
	at org.apache.commons.compress.harmony.unpack200.Archive.unpack(Archive.java:198)
	at com.unascribed.unpacktest.UnpackTest.main(UnpackTest.java:20)
```

Pack200 is *intended* to be combined with other forms of compression, so this is a major issue in some cases.

This reimplements the portion of the method to use `read` correctly, keeping track of how much has been read. I did not address the TODO as I don't believe there's a good way to implement it with the current architecture of the decompressor.

The file I'm having issues with is a Pack200+LZMA file that does not contain any class files. Kind of a weird case, but it's what I'm stuck with, and Commons Compress is the only real extant Pack200 decompressor for modern Java.

### DIFF
--- a/src/main/java/org/apache/commons/compress/harmony/unpack200/FileBands.java
+++ b/src/main/java/org/apache/commons/compress/harmony/unpack200/FileBands.java
@@ -21,6 +21,7 @@ import java.io.InputStream;
 
 import org.apache.commons.compress.harmony.pack200.Codec;
 import org.apache.commons.compress.harmony.pack200.Pack200Exception;
+import org.apache.commons.compress.utils.IOUtils;
 
 /**
  * Parses the file band headers (not including the actual bits themselves). At the end of this parse call, the input
@@ -82,7 +83,7 @@ public class FileBands extends BandSet {
             // TODO This breaks if file_size > 2^32. Probably an array is
             // not the right choice, and we should just serialize it here?
             fileBits[i] = new byte[size];
-            final int read = in.read(fileBits[i]);
+            final int read = IOUtils.readFully(in, fileBits[i]);
             if (size != 0 && read < size) {
                 throw new Pack200Exception("Expected to read " + size + " bytes but read " + read);
             }


garydgregory:
@unascribed 
Thank you for the PR. You'll need to update this PR with a unit test. Ty.

unascribed:
The file I encountered this issue with is not redistributable, and I'm unsure of the best way to synthesize an equivalent test. I'll mess around when I have time.

----

For passerby with the same issue, in the meantime this is published to [my Maven](https://repo.sleeping.town) with coordinate `com.unascribed:commons-compress:1.23-pack200fix`.

quat1024:
A little bit more code digging, just for fun:

* If the provided `InputStream` does not support marking, [`unpack200.Segment` will wrap it in a `BufferedInputStream`](https://github.com/apache/commons-compress/blob/e6930d06cfebbdbee586b863481779b2c4b9b202/src/main/java/org/apache/commons/compress/harmony/unpack200/Segment.java#L499-L502), which does.
* `BufferedInputStream`, "as a convenience", repeatedly calls `read` until it fills as many bytes of the user-provided input array as possible, using a driver loop not unlike the one in this PR. However, if the wrapped stream's `available` method returns `0` (signaling that further `read` calls will block), `BufferedInputStream` will stop early before filling the entire array (because it doesn't want to block).
* `LZMACompressorInputStream` [delegates the `available` method](https://github.com/apache/commons-compress/blob/e6930d06cfebbdbee586b863481779b2c4b9b202/src/main/java/org/apache/commons/compress/compressors/lzma/LZMACompressorInputStream.java#L102-L106) through to `xz`'s `LZMAInputStream`.
* [`LZMAInputStream` doesn't override `available`](https://git.tukaani.org/?p=xz-java.git;a=blob;f=src/org/tukaani/xz/LZMAInputStream.java;h=1432eb1037a513dcd997fb0a946cc978032565aa;hb=HEAD), so the default implementation is used.
* The default implementation always returns 0,
* -> causing `BufferedInputStream` to be okay with returning a half-filled buffer,
* -> triggering the correctness error in the code that assumes `read(byte[])` fills the whole byte array,
* -> breaking the file parse.

A cheeky workaround is to wrap the stream provided to `Pack200CompressorInputStream` in one that returns `false` in `markSupported` and `Integer.MAX_VALUE` in `available`, this makes `BufferedInputStream` do your work for you.

garydgregory:
Could one the method be reused from https://github.com/apache/commons-compress/blob/master/src/main/java/org/apache/commons/compress/utils/IOUtils.java ?

jochenw:
+1 for pulling in, even without a test, The user has obviously spend considerable work on this, and manual inspection looks fine.

unascribed:
> Could one the method be reused from IOUtils?

Yes, that readFully method does the same thing. I'm not familiar with this project's internals so I wasn't aware it existed — nowadays I use InputStream#readNBytes, but that's Java 11.

kinow (approved):
Hmmm, tricky to write a unit test for this, especially hard to find or craft a test file for this. But the explanation and fix both look good IMO, so +1 too. Thanks @unascribed !

garydgregory:
Merged. TY.  Now, where else should we do the same kind of change?

unascribed:
I sure hope this isn't a common pattern throughout the codebase.

I'd imagine this is an isolated incident due to this being a very seldom used feature of the library, and a case of inheriting poorly tested code from Harmony, which itself was written in something of a hurry and without thorough testing be a priority — just "get API compatibility with pre-open Java ASAP".

A quick random look around the codebase guided by Open Call Hierarchy in Eclipse suggests this is done correctly elsewhere, but it was only a random sample.

PR #365, opened 2023-03-06
COMPRESS-641: add getLinkFlag for TarArchiveEntry

XenoAmess:
Hi.
I'm making a tar compressing tool.
And I need to know what the original flag of a tar entry I read, otherwise I would make it changed in the decompressed result...

### DIFF
--- a/src/main/java/org/apache/commons/compress/archivers/tar/TarArchiveEntry.java
+++ b/src/main/java/org/apache/commons/compress/archivers/tar/TarArchiveEntry.java
@@ -2129,5 +2129,15 @@ public class TarArchiveEntry implements ArchiveEntry, TarConstants, EntryStreamO
         }
         return offset;
     }
+
+    /**
+     * Get this entry's link flag.
+     *
+     * @since 1.23
+     */
+    public byte getLinkFlag() {
+        return this.linkFlag;
+    }
+
 }
 
--- a/src/test/java/org/apache/commons/compress/ChainingTestCase.java
+++ b/src/test/java/org/apache/commons/compress/ChainingTestCase.java
@@ -23,6 +23,7 @@ import static org.junit.jupiter.api.Assertions.assertNotNull;
 
 import org.apache.commons.compress.archivers.tar.TarArchiveEntry;
 import org.apache.commons.compress.archivers.tar.TarArchiveInputStream;
+import org.apache.commons.compress.archivers.tar.TarConstants;
 import org.apache.commons.compress.compressors.bzip2.BZip2CompressorInputStream;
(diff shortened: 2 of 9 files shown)


garydgregory (changes requested):
-1:  Javadoc is completely wrong; no unit test.

kinow (commented):
@XenoAmess are you going to retrieve the `linkFlag` byte value, and then compare doing something like `if (linkFlag == SOME_CONSTANT) {} else if (linkFlag == ANOTHER_CONSTANT) { ... }` ?

If so, would it be possible to instead use the public methods like `if (obj.isBlockDevice())`, `else if (obj.isCharacterDevice())`, etc., instead?

Just checking if there's no alternative to making this property public.

Cheers,

-Bruno

XenoAmess:
> @XenoAmess are you going to retrieve the `linkFlag` byte value, and then compare doing something like `if (linkFlag == SOME_CONSTANT) {} else if (linkFlag == ANOTHER_CONSTANT) { ... }` ?
> 
> If so, would it be possible to instead use the public methods like `if (obj.isBlockDevice())`, `else if (obj.isCharacterDevice())`, etc., instead?
> 
> Just checking if there's no alternative to making this property public.
> 
> Cheers,
> 
> -Bruno

no, I would put it directly into the output tar

the mechenism of my compression is only compress entries with some linkFlag types.

for others, just make them same and clone them directly into the output compressed tar.

XenoAmess:
> -1: Javadoc is completely wrong; no unit test.

javadoc fixed.

Would add unit test, but I want to know about your opinons about exposing such a field first.

If made sure you don't mind we expose it, then I would add the unit tests.

garydgregory:
Adding the getter seems reasonable to me as it is documented here if I am reading correctly: https://www.gnu.org/software/tar/manual/html_node/Standard.html

kinow:
> Adding the getter seems reasonable to me as it is documented here if I am reading correctly: https://www.gnu.org/software/tar/manual/html_node/Standard.html

Yeah, I was trying to think in other ways to solve what @XenoAmess needed without adding that getter (as there are other properties that are private only, without getters), but I can't think of an alternative. So +1 to the getter too.

XenoAmess:
@garydgregory 
unit tests added.

kinow (approved):
Just needs to squash commits. Otherwise, looks good to me! Thanks @XenoAmess !

XenoAmess:
> Just needs to squash commits. Otherwise, looks good to me! Thanks @XenoAmess !

@kinow squashed

kinow (approved):

PR #378, opened 2023-04-10
Optimize the nameMap of ZipFile

Glavo:
For most regular zip files, all of their entries will have different names.

The current implementation of ZipFile maps each name to a LinkedList in the nameMap in order to deal with the case of duplicate entry names. Normally (64-bit platform, compressed oops enabled), these LinkedLists occupy an additional 56 bytes of memory.

This PR optimizes the implementation of `ZipFile`, thereby:

* For most zip files, this will save memory;
* For most zip files, the speed of filling in nameMap becomes faster when opening `ZipFile`;
* `ZipFile::getEntry(String)` will be faster.

For extremely rare situations, this PR may have slight negative effects, but I think this should not be a problem.


kinow (commented):
If I understand it right, instead of having a reverse map with the names and list of duplicated entries, here we would have a new separate map to keep the duplicates. In case there are no entries with the same name, that data structure would not be used, and it would perform a little better.

`getEntry` will be definitely faster.

“For most zip files...” I think that might be correct, for most cases. +0 as I am not using the component with a large-ish project where I could test it. 

Thanks!

garydgregory:
No tests. Code coverage is down. No proof of performance change one way or another. Non-trivial additional complexity. So -1 as it stands. 

"For extremely rare situations, this PR may have slight negative effects,"

Which are?

"But I think this should not be a problem."

Because?

Glavo:
> No tests. Code coverage is down. No proof of performance change one way or another. Non-trivial additional complexity.

The core intention of this PR is to reduce memory allocation.

As mentioned earlier, currently we allocate a LinkedList for each entry in ZipFile, resulting in an additional 32 (LinkedList) + 24 (LinkedList$Node) bytes of memory overhead. This PR no longer assigns them.

As for the side effect - performance improvement, this is the result of a simple JMH benchmark:

<details><summary>Benchmark source code</summary>
<p>

```java
public class Compress {

    private static final Path jarPath = Path.of(/* path of commons-compress-1.23.0.jar */);

    private static final ZipFile globalFile ;
    private static final String[] entryNames;

    static {
        try {
            globalFile = new ZipFile(jarPath);
        } catch (IOException e) {
            throw new RuntimeException(e);
        }

        entryNames = Collections.list(globalFile.getEntries())
                .stream()
                .map(ZipArchiveEntry::getName)
                .toArray(String[]::new);
    }

    @Benchmark
    public void testOpen() throws IOException {
        try (ZipFile file = new ZipFile(jarPath)) {
        }
    }

    @Benchmark
    public void testGetEntry(Blackhole blackhole) throws IOException {
        for (String entryName : entryNames) {
            blackhole.consume(globalFile.getEntry(entryName));
        }
    }

    @Benchmark
    public void testGetEntries(Blackhole blackhole) throws IOException {
        for (String entryName : entryNames) {
            blackhole.consume(globalFile.getEntries(entryName));
        }
    }

    @Benchmark
    public void testGetEntriesInPhysicalOrder(Blackhole blackhole) throws IOException {
        for (String entryName : entryNames) {
            blackhole.consume(globalFile.getEntriesInPhysicalOrder(entryName));
        }
    }
}

```

</p>
</details>

commons-compress 1.23.0:

```
Benchmark                                Mode  Cnt       Score      Error  Units
Compress.testGetEntries                 thrpt    3  657266.611 ± 1690.346  ops/s
Compress.testGetEntriesInPhysicalOrder  thrpt    3  121514.308 ± 1001.179  ops/s
Compress.testGetEntry                   thrpt    3  504127.618 ± 1988.320  ops/s
Compress.testOpen                       thrpt    3     663.604 ±   24.088  ops/s
```

This PR:

```
Benchmark                                Mode  Cnt       Score       Error  Units
Compress.testGetEntries                 thrpt    3  465237.072 ± 37283.323  ops/s
Compress.testGetEntriesInPhysicalOrder  thrpt    3  459856.583 ± 13165.000  ops/s
Compress.testGetEntry                   thrpt    3  661743.860 ±  1055.489  ops/s
Compress.testOpen                       thrpt    3     665.040 ±    23.971  ops/s
```

For the same jar file (commons-compress-2.13.0.jar):

*  Little change in performance opening zip files;
* `getEntry` is 30% faster;
* `getEntriesInPhysicalOrder(String)` is 280% faster.

The only problem is `getEntries(String)`, which is 30% slower, because each call needs to create a new singleton list. But the old implementation was unsafe because it would return the internal mutable list that could be modified by the user simply by casting, thus breaking the `ZipFile`, the implementation in this PR fixes this issue.

> "For extremely rare situations, this PR may have slight negative effects,"
> 
> Which are?
>
> "But I think this should not be a problem."
> 
> Because?

Because this situation is too rare.

As far as I know, most tools (7zip, jar commands, etc.) don't create zip files with entries with duplicate names. Such unusual zip files are generally only possible to create using some low-level api, so such files themselves are very rare, I can't even easily find examples in the real world.

And, even for this unusual file, it's debatable whether this PR has a positive or negative effect. 

If the majority of items do not have the same name, this PR still has a positive effect. 

And if there are a large number of entries with the same name, the effect of this PR may be different under different conditions. But I don't think it's a good case to look into, as I can't imagine a real world use case for it.

If you guys know of a real use case for duplicate entries, I'll look into that further. But for now, I don't see much point in caring too much about this situation, since regular zip files are much, much more numerous.

src/main/java/org/apache/commons/compress/archivers/zip/ZipFile.java
```diff
@@ -371,9 +372,14 @@ public static void closeQuietly(final ZipFile zipFile) {
...
-     * Maps String to list of ZipArchiveEntrys, name -> actual entries.
+     * Maps a string to the first entry named it.
      */
-    private final Map<String, LinkedList<ZipArchiveEntry>> nameMap = new HashMap<>(HASH_SIZE);
+    private final Map<String, ZipArchiveEntry> nameMap = new HashMap<>(HASH_SIZE);
+
+    /**
+     * If multiple entries have the same name, maps the name to entries named it.
+     */
+    private Map<String, List<ZipArchiveEntry>> duplicateNameMap = null;
```
garydgregory, line 382: Don't initialize to the default value.

src/main/java/org/apache/commons/compress/archivers/zip/ZipFile.java
```diff
@@ -792,11 +798,24 @@ private BoundedArchiveInputStream createBoundedInputStream(final long start, fin
 
     private void fillNameMap() {
         entries.forEach(ze -> {
-            // entries is filled in populateFromCentralDirectory and
+            // entries are filled in populateFromCentralDirectory and
             // never modified
             final String name = ze.getName();
-            final LinkedList<ZipArchiveEntry> entriesOfThatName = nameMap.computeIfAbsent(name, k -> new LinkedList<>());
-            entriesOfThatName.addLast(ze);
+            ZipArchiveEntry firstEntry = nameMap.putIfAbsent(name, ze);
```
garydgregory, line 804: Use final.

src/main/java/org/apache/commons/compress/archivers/zip/ZipFile.java
```diff
@@ -868,9 +887,12 @@ public Enumeration<ZipArchiveEntry> getEntries() {
      * @since 1.6
      */
     public Iterable<ZipArchiveEntry> getEntries(final String name) {
-        final List<ZipArchiveEntry> entriesOfThatName = nameMap.get(name);
-        return entriesOfThatName != null ? entriesOfThatName
-            : Collections.emptyList();
+        List<ZipArchiveEntry> entriesOfThatName = duplicateNameMap == null ? null : duplicateNameMap.get(name);
```
garydgregory, line 890: Use final.

src/main/java/org/apache/commons/compress/archivers/zip/ZipFile.java
```diff
@@ -899,13 +921,17 @@ public Enumeration<ZipArchiveEntry> getEntriesInPhysicalOrder() {
...
     public Iterable<ZipArchiveEntry> getEntriesInPhysicalOrder(final String name) {
-        ZipArchiveEntry[] entriesOfThatName = ZipArchiveEntry.EMPTY_ARRAY;
-        final LinkedList<ZipArchiveEntry> linkedList = nameMap.get(name);
-        if (linkedList != null) {
-            entriesOfThatName = linkedList.toArray(entriesOfThatName);
-            Arrays.sort(entriesOfThatName, offsetComparator);
+        if (duplicateNameMap != null) {
+            List<ZipArchiveEntry> list = duplicateNameMap.get(name);
+            if (list != null) {
+                ZipArchiveEntry[] entriesOfThatName = list.toArray(ZipArchiveEntry.EMPTY_ARRAY);
```
garydgregory, line 927: Instead of: 

1. allocate array
2. sort array
3. allocate list
 
would it be better to change the type from `List` to `LinkedList` (which is what is used in the instance variable decl and:

1. `clone()` the list
2. `Collections.sort()` the cloned list

?

src/main/java/org/apache/commons/compress/archivers/zip/ZipFile.java
```diff
@@ -792,11 +798,24 @@ private BoundedArchiveInputStream createBoundedInputStream(final long start, fin
...
-            entriesOfThatName.addLast(ze);
+            ZipArchiveEntry firstEntry = nameMap.putIfAbsent(name, ze);
+
+            if (firstEntry != null) {
+                if (duplicateNameMap == null) {
+                    duplicateNameMap = new HashMap<>();
+                }
+
+                List<ZipArchiveEntry> entriesOfThatName = duplicateNameMap.computeIfAbsent(name, k -> {
+                    ArrayList<ZipArchiveEntry> list = new ArrayList<>(2);
```
garydgregory, line 812: Why 2?

garydgregory (changes requested):
@Glavo (CC @kinow)
Thank you for the PR. Please see my comments.

src/main/java/org/apache/commons/compress/archivers/zip/ZipFile.java
Glavo: > Instead of:
> 
> 1. allocate array
> 2. sort array
> 3. allocate list
> 
> would it be better to change the type from `List` to `LinkedList` (which is what is used in the instance variable decl and:
> 
> 1. `clone()` the list
> 2. `Collections.sort()` the cloned list
> 
> ?

I think it might be cleaner to keep it as it is now.

The return type of the `clone` method of `ArrayList`/`LinkedList` is `Object`, which means that after calling it, a cast is required and `SuppressWarnings` annotation is required to suppress unchecked warnings.

src/main/java/org/apache/commons/compress/archivers/zip/ZipFile.java
Glavo: The reason for using `ArrayList` instead of `LinkedList` here is that `ArrayList` is almost always better than `LinkedList`.

In general, we can calculate the memory footprint of a lists with `N` elements like this:

* LinkedList: `24 + 24*N`
* ArrayList: 
  * Minimal (internal array is fully used, no wasted space): 
    `24 (ArrayList) + 16 (array object header) + 4*N`
  * Maximum (internal array has just grown, the reserved space is the largest):
    `24 (ArrayList) + 16 (array object header) + 6*N`
  
If the list has only two elements, `ArrayList` with 2 as the initial capacity saves 24 bytes of memory compared to `LinkedList`.

As the list grows, the memory footprint of `ArrayList` is usually only 1/6~1/4 of that of `LinkedList`. 

The combined memory footprint of all objects allocated during `ArrayList` construction (including those small temporary arrays) is smaller than the long-term memory footprint of `LinkedList`, so using `ArrayList` is more memory efficient.

For smaller lists, iterating over a `LinkedList` might be faster, but `ArrayList` has better data locality. So usually we have no reason to use `LinkedList`, `ArrayList`/`ArrayDeque` are better choices.

Glavo:
Can someone review this PR again?

Glavo:
@garydgregory Can you review this PR again?

Glavo:
@garydgregory Hey, can you take a look at this PR?

elharo:
Looks OK to me, though claims of performance improvements really need measurements and benchmarks.

garydgregory:
> > Looks OK to me, though claims of performance improvements really need measurements and benchmarks.
> 
> Here are some JMH benchmark results: [#378 (comment)](https://github.com/apache/commons-compress/pull/378#issuecomment-1501863937)

But where is the benchmark? You should make it part of this PR so thar anyone can run it.

Glavo:
> But where is the benchmark?

The source code is in the `Benchmark source code` folding block of the comment.

image

src/main/java/org/apache/commons/compress/archivers/zip/ZipFile.java
```diff
@@ -820,7 +839,13 @@ public Enumeration<ZipArchiveEntry> getEntries() {
      * @since 1.6
      */
     public Iterable<ZipArchiveEntry> getEntries(final String name) {
-        return nameMap.getOrDefault(name, ZipArchiveEntry.EMPTY_LINKED_LIST);
```
DamnedElric, line 823: You mentioned a potential 30% slowdown in this method in return for additional safety, which may or may not be a fair tradeoff. But I'm a bit concerned that this change potentially breaks the current behaviour. As you say, this currently returns a modifiable LinkedList, which some people may (though I hope not) depend on? Because it requires an explicit cast, I'm not too worried.

DamnedElric (commented):
Made a couple of (mostly nitpicky) comments on the diffs. 

Regarding the benchmarks: it might be nice to have the benchmarks as part of the testsuite, but there are currently no benchmarks in compress. Might be useful as part of another MR?

src/main/java/org/apache/commons/compress/archivers/zip/ZipFile.java
```diff
@@ -368,9 +369,14 @@ public static void closeQuietly(final ZipFile zipFile) {
...
-     * Maps String to list of ZipArchiveEntrys, name -> actual entries.
+     * Maps a string to the first entry by that name.
      */
-    private final Map<String, LinkedList<ZipArchiveEntry>> nameMap = new HashMap<>(HASH_SIZE);
+    private final Map<String, ZipArchiveEntry> nameMap = new HashMap<>(HASH_SIZE);
+
+    /**
+     * If multiple entries have the same name, maps the name to entries by that name.
+     */
+    private Map<String, List<ZipArchiveEntry>> duplicateNameMap;
```
kvr000, line 379: Given that this is extremely unlikely, would it make sense to keep the original Map and change the value to Object and let it hold either ZipArchiveEntry or the List?  It would eventually complicate the lookup but that should be a single place only.

src/main/java/org/apache/commons/compress/archivers/zip/ZipFile.java
kvr000: I think the slowdown is only in case there are duplicit entries which is extremely unlikely.  I'd still simplify the code, as according to earlier comment.

src/main/java/org/apache/commons/compress/archivers/zip/ZipFile.java
garydgregory: I  REALLY don't like that. When you look at the pack200 code, which is written at the pre-generics Java version, and some of its lists, you can't tell if some of the lists are buggy, sloppy, or trying to be clever. That or the lack of generics let's you get away with not having fully thought through the hierarchy of types you put in those lists.

kvr000: That's true, such optimizations typically come at cost of readability and safety.  However, keeping the amount of code touching the data structures limited may help - we'll need only three methods - `first-by-name`, `all-by-name` and `add-entry` .  And document the field to not touch except via those three methods :-)

Just a suggestion if it comes to true optimization, given this PR is purely about improving time and space performance.

garydgregory:
Hello All
I am -1 until tests are provided that show the new data structure is correctly populated. A package-private getter can be added and queried by new unit tests for example. Otherwise, it's just too easy undo in the future without visibility. If this were indeed undone in the future, then forcing a maintainer to remove tests will make it obvious that this was a intended feature. Edge cases should IMO account for no duplicates, some duplicated, all duplicates use cases.

garydgregory:
Hello everyone,

In commit a8cb750e5bdd0fe136d5d9cab38781f9447b2dd3, I introduced `org.apache.commons.compress.archivers.zip.ZipFileNameMapBenchmark`. 

However, the benchmark indicates a 30% performance degradation in one method. This raises concerns about whether the change is justified. Could you help me evaluate whether this trade-off is acceptable, or if I'm overlooking something?

kvr000:
> In commit [a8cb750](https://github.com/apache/commons-compress/commit/a8cb750e5bdd0fe136d5d9cab38781f9447b2dd3), I introduced `org.apache.commons.compress.archivers.zip.ZipFileNameMapBenchmark`.
> 
> However, the benchmark indicates a 30% performance degradation in one method. This raises concerns about whether the change is justified. Could you help me evaluate whether this trade-off is acceptable, or if I'm overlooking something?

It depends on typical usage.  The benchmark opens the file only once and then repeatedly requests the zip entry.  As getEntries() now dynamically wrap the list with Collections.singletonList(), it adds to the time.

However, is it the typical scenario that user opens the file once and then asks repeatedly for the same entry?  Unlikely - maybe in jar files but I guess those are rather cached.

A compromise may be to to use the original code and wrap it into Collections.singletonList() if there is single entry only - this will add one object wrapper only, improve the space consumption, improve population phase speed and keep the lookup at the same speed (rather slightly faster as iterating over hardcoded singleton list is more trivial).

Alternatively, keeping this code and introduce getOnlyEntry() (crashing if there are more) and getFirstEntry() (return first if there are more) - which would be the typical usages anyway.  Then the new code will be better by avoiding the singletonList() wrapper for all entries in memory and avoiding unnecessary iteration.

Glavo:
We already have a way to retrieve the first entry: `getEntry`.

I don’t think `getEntries` would be particularly useful, because most code expects each name to map to a unique entry rather than being designed to handle entries with conflicting names. The main value of `getEntries` would probably be to detect duplicate entries and report an error. In my view, optimizing for that path would not justify sacrificing memory usage, the performance of other methods, or thread safety.

By comparison, I think it would be more useful to introduce a new option for opening ZIP files. When this option is enabled, the operation would fail immediately if conflicting entries are detected.

garydgregory:
> We already have a way to retrieve the first entry: `getEntry`.
> 
> I don’t think `getEntries` would be particularly useful, because most code expects each name to map to a unique entry rather than being designed to handle entries with conflicting names. The main value of `getEntries` would probably be to detect duplicate entries and report an error. In my view, optimizing for that path would not justify sacrificing memory usage, the performance of other methods, or thread safety.
> 
> By comparison, I think it would be more useful to introduce a new option for opening ZIP files. When this option is enabled, the operation would fail immediately if conflicting entries are detected.

I like that idea 💡!

kvr000:
I'd be cautious about that - maybe not too common with zip archives but some software may simply append to file when the entry is modified, instead of rearranging the existing stream (central directory must be still rewritten or newly appended though).

So formally this is valid situation and the last entry should be returned as the valid one (in case one entry is requested) - correcting my earlier proposal to return the first one - that would be wrong.

Anyway, if the benchmark is slower for multiple entries but it's the same or rather faster for single entry - that is good.

garydgregory:
I converted this PR to draft.

PR #420, opened 2023-09-30
Fixing Bug: hasNext() should be called before calling next() on iterator object

SabbirMeraj:
It is recommended to call the hasNext() method before calling the next() method to ensure that there is another element in the collection to retrieve. This practice helps avoid exceptions or errors that may occur when attempting to retrieve elements that don't exist.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ target
 *~
 /.externalToolBuilders/
 /maven-eclipse.xml
+/bin/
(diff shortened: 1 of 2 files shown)


garydgregory:
-1: Note this is test code. If you want, you can create NEW test methods to test different patterns. What's worse, is this PR will hide errors in the code because where next() might have returned null or thrown an exception, this will no longer be the case.

PR #435, opened 2023-11-09
COMPRESS-632: Adding a check for invalid maxCodeSize

yakovsh:
Adds a validation check and unit test for possible invalid values of maxCodeSize when parsing Z files. This was discovered as part of the fuzzing expansion covered in COMPRESS-632.

This addresses the fact the that maxCodeSize value that is used to initialize the decompression is based on raw byte read by ZCompressorInputStream without validation (unlike the UnshrinkingInputStream in ZIP where is it simply set to a default value). The validation adds three checks:
- Detect values less or equal to zero, otherwise initializeTables() in LZWInputStream will throw an IllegalArgumentException (lines 198-201)
- Detect that (1 << maxCodeSize) is less than 256, otherwise the loop in initializeTables() in LZWInputStream will throw array exceptions (lines 181-184)
- Detect that maxCodeSize is larger than codeSize, otherwise uncaught exceptions will be thrown when the stream is parsed. Specifically, addEntry() in LZWInputStream, lines 79-80.

### DIFF
--- a/src/main/java/org/apache/commons/compress/compressors/z/ZCompressorInputStream.java
+++ b/src/main/java/org/apache/commons/compress/compressors/z/ZCompressorInputStream.java
@@ -72,11 +72,6 @@ public class ZCompressorInputStream extends LZWInputStream {
         if (blockMode) {
             setClearCode(DEFAULT_CODE_SIZE);
         }
-        // maxCodeSize shifted cannot be less than 256, otherwise the loop in initializeTables() will throw an ArrayIndexOutOfBoundsException
-        // maxCodeSize cannot be smaller than getCodeSize(), otherwise addEntry() will throw an ArrayIndexOutOfBoundsException
-        if ((1 << maxCodeSize) < 256 || getCodeSize() > maxCodeSize) {
-            throw new IOException("Invalid maxCodeSize value: " + maxCodeSize);
-        }
         initializeTables(maxCodeSize, memoryLimitInKb);
         clearEntries();
     }


src/test/java/org/apache/commons/compress/compressors/z/ZCompressorInputStreamTest.java
```diff
@@ -68,4 +70,24 @@ public void testSingleByteReadConsistentlyReturnsMinusOneAtEof() throws IOExcept
...
+
+            // Test all possible maxCodeSize values, it will either process correctly, or throw an IOException
+            for (int maxCodeSize = Byte.MIN_VALUE; maxCodeSize <= Byte.MAX_VALUE; maxCodeSize++) {
+                content[2] = (byte) maxCodeSize;
+                try {
+                    final ZCompressorInputStream in = new ZCompressorInputStream(new ByteArrayInputStream(content), 1024*1024);
+                    IOUtils.toByteArray(in);
+                    assertEquals(-1, in.read());
+                } catch(Exception e) {
+                    assertInstanceOf(IOException.class, e);
```
garydgregory, line 87: Hello @yakovsh 
This is test is confusing: What is the expected behavior? What method call, _exactly_, do you expect to throw `IOException`? The current test makes it look like we expect _anything_ (the `ZCompressorInputStream` constructor, the call to `toByteArray()`, or the call to `read()`) in the try block to throw `IOException`, which surely is not what we want to assert. I imaging that you care about the `read()` call. If so, use `assertThrow()` for that one call.
TY!

garydgregory (changes requested):
Hi @yakovsh 
Thank you for your PR. Please see my code comment.

src/main/java/org/apache/commons/compress/compressors/z/ZCompressorInputStream.java
```diff
@@ -72,6 +72,12 @@ public ZCompressorInputStream(final InputStream inputStream, final int memoryLim
         if (blockMode) {
             setClearCode(DEFAULT_CODE_SIZE);
         }
+        // maxCodeSize cannot be zero, otherwise initializeTables() will throw an IllegalArgumentException
+        // maxCodeSize shifted cannot be less than 256, otherwise the loop in initializeTables() will throw an ArrayIndexOutOfBoundsException
+        // maxCodeSize cannot be smaller than getCodeSize(), otherwise addEntry() will throw an ArrayIndexOutOfBoundsException
+        if (maxCodeSize <= 0 || (1 << maxCodeSize) < 256 || getCodeSize() > maxCodeSize) {
+            throw new IOException("Invalid maxCodeSize value: " + maxCodeSize);
```
garydgregory, line 79: Hello @yakovsh 
Thank you for your update. I still don't understand the point of this change. Without this check, the code throws `java.lang.IllegalArgumentException: maxCodeSize is 0, must be bigger than 0` from `LZWInputStream.initializeTables(int, int)` . So why do we need this check instead of the one in `initializeTables()`?

The test seems fine.

src/test/java/org/apache/commons/compress/compressors/z/ZCompressorInputStreamTest.java
```diff
@@ -68,4 +70,28 @@ public void testSingleByteReadConsistentlyReturnsMinusOneAtEof() throws IOExcept
...
+                content[2] = (byte) maxCodeSize;
+
+                // Test that no unexpected exceptions are thrown when initializing and reading the stream,
+                // since maxCodeSize impacts both initialization and reading.
+                assertDoesNotThrow(() -> {
+                    try {
+                        final ZCompressorInputStream in =
+                                new ZCompressorInputStream(new ByteArrayInputStream(content), 1024*1024);
+                        IOUtils.toByteArray(in);
+                    } catch(IOException ignore) {
```
garydgregory, line 90: Please follow the style from the other files: 
- Space before `(`
- Space around `*`

garydgregory: Hello @yakovsh 

The test is still too obtuse to maintain IMO:
- The test asserts that it does NOT throw, but
- Then it expects to catch an IOException and then ignores it
This is impossible for the mere mortal maintainer to understand.
The test would read better by testing which range of values are expected to NOT throw, and, which range of values are expected to throw.

TY!

src/test/java/org/apache/commons/compress/compressors/z/ZCompressorInputStreamTest.java
```diff
@@ -68,4 +73,32 @@ public void testSingleByteReadConsistentlyReturnsMinusOneAtEof() throws IOExcept
...
+
+            for (int value : invalidValues) {
+                content[2] = (byte) value;
+
+                // Test that invalid values always throw an IOException
+                assertThrows(IOException.class, () ->
+                        new ZCompressorInputStream(new ByteArrayInputStream(content), 1024 * 1024)
+                );
+            }
+        }
```
garydgregory, line 101: Simplify and clarify: You can read the file test fixture's content in one call. You can just stream over the IntStreams.
```
        final byte[] bytes = Files.readAllBytes(AbstractTest.getPath("bla.tar.Z"));

        // @formatter:off
        final IntStream[] invalid = {
            IntStream.range(Byte.MIN_VALUE, -120),
            IntStream.range(-97, -88),
            IntStream.range(-65, -56),
            IntStream.range(-33, -24),
            IntStream.range(-1, 8),
            IntStream.range(31, 40),
            IntStream.range(63, 72),
            IntStream.range(95, 104),
            IntStream.range(127, 127)
            };
        // @formatter:on

        Stream.of(invalid).forEach(ints -> ints.forEach(i -> {
            bytes[2] = (byte) i;
            assertThrows(IllegalArgumentException.class, () -> new ZCompressorInputStream(new ByteArrayInputStream(bytes), 1024 * 1024),
                    () -> "value=" + i);
        }));
```
Also, the exception should be thrown in the initializeTables method where the check is already performed and can be expanded.

garydgregory: Many methods throw unchecked exceptions in this library and in most libraries and IN the JDK, so any truly billet-proof app will know where it needs to watch out for this based on testing, fuzzing, and general empirical usage. All of this to say that the method ALREADY threw IAE, so now it does so under another use case. IOW, it's fine the way it is IMO. I'm happy to hear other opinions.

garydgregory:
Hi @yakovsh 
Please close this PR when you are ok with git master.

PR #437, opened 2023-11-11
COMPRESS-650 lz4 index out of bounds

cpreisler:
I noticed that the change I made yesterday caused writing byte by byte to produces a different result than writing multiple bytes at one time. I fixed the code and added unit tests. Also I verified that this code is now producing the same output as the previous 1.24.0 version. Also verified that the lz4 command line utility is able to decompress the output from this code.


cpreisler:
Shoot. I just noticed that I put the framed tests in the block tests class. Should I create a new test class for framed? Currently there is only the framed round trip class. The new tests aren't really round trip tests.

garydgregory:
@cpreisler 
Sure, might as well.

cpreisler:
New class added and test are moved.

src/test/java/org/apache/commons/compress/compressors/lz4/FramedLZ4CompressorRoundtripTest.java
```diff
@@ -104,15 +104,18 @@ private void roundTripTest(final String testFile, final FramedLZ4CompressorOutpu
 
     @Test
     public void test64KMultipleBlocks() throws IOException {
-        final ByteArrayOutputStream buffer = new ByteArrayOutputStream();
-        final byte[] expected = new byte[98304];
+        ByteArrayOutputStream buffer = new ByteArrayOutputStream();
```
garydgregory, line 107: Hello @cpreisler 
Thank you for your update.
Why has this method changed? I can't spot the functional differences. Why is `count` declared but not used? Revert if this is just formatting or noise.

garydgregory (changes requested):
Please see my test comment.

src/test/java/org/apache/commons/compress/compressors/lz4/FramedLZ4CompressorRoundtripTest.java
cpreisler: You are correct. There was no functional changes. I guess I got a little confused during the initial commit of my changes today. I did remove the unused count variable. I will be more careful with my merge/commits in the future.

src/test/java/org/apache/commons/compress/compressors/lz4/FramedLZ4CompressorRoundtripTest.java
```diff
@@ -104,15 +104,17 @@ private void roundTripTest(final String testFile, final FramedLZ4CompressorOutpu
...
-                new FramedLZ4CompressorOutputStream.Parameters(FramedLZ4CompressorOutputStream.BlockSize.K64, true, false, false))) {
+        try (FramedLZ4CompressorOutputStream compressor = 
+                new FramedLZ4CompressorOutputStream(buffer, 
+                        new FramedLZ4CompressorOutputStream.Parameters(FramedLZ4CompressorOutputStream.BlockSize.K64, true, false, false))) {
             compressor.write(expected);
         }
         try (FramedLZ4CompressorInputStream sis = new FramedLZ4CompressorInputStream(new ByteArrayInputStream(buffer.toByteArray()))) {
-            assertArrayEquals(expected, IOUtils.toByteArray(sis));
+            final byte[] actual = IOUtils.toByteArray(sis);
+            assertArrayEquals(expected, actual);
```
garydgregory, line 117: Hello @cpreisler 
Same question as before:
Why is this method different? Don't change it unless it is required. Non-required changes make PRs harder and longer to review.

src/test/java/org/apache/commons/compress/compressors/lz4/FramedLZ4CompressorOutputStreamTest.java
```diff
@@ -0,0 +1,72 @@
...
+
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.util.Arrays;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import org.junit.jupiter.api.Test;
+
+/**
+ *
+ * @author Chad Preisler
```
garydgregory, line 29: We do not use author tags, please remove; instead you can say `Tests {@link TheClass}.`. You will be credited in `changes.xml` after the merge,

src/test/java/org/apache/commons/compress/compressors/lz4/FramedLZ4CompressorOutputStreamTest.java
```diff
@@ -0,0 +1,72 @@
...
+ */
+public class FramedLZ4CompressorOutputStreamTest {
+    
+    @Test
+    public void testWriteByteArrayVsWriteByte() throws IOException {
+        ByteArrayOutputStream buffer = new ByteArrayOutputStream();
+        StringBuilder sb = new StringBuilder();
+        for (int i = 0; i < 1; i++) {
```
garydgregory, line 37: Why do you use a loop when this only done once?

cpreisler:
I’m away from my computer right now. Let me look this over and make sure
the right code got committed. Something may have gotten messed up when I
tried to revert the other change.

garydgregory:
@cpreisler 
OK, ty for the update. FYI, I plan on cutting a release candidate this weekend.

cpreisler:
I’ll get it done today.

src/test/java/org/apache/commons/compress/compressors/lz4/FramedLZ4CompressorOutputStreamTest.java
cpreisler: I removed the comment from the code.

src/test/java/org/apache/commons/compress/compressors/lz4/FramedLZ4CompressorOutputStreamTest.java
cpreisler: I got rid of the useless for loop. I reused my round trip code and this was left over crud. Thanks for catching this.

src/test/java/org/apache/commons/compress/compressors/lz4/FramedLZ4CompressorRoundtripTest.java
cpreisler: let me try to get this out of here. Again, when I committed my changes the merge was showing conflicts with master, so that's how it got here.

cpreisler: This is gone. From my perspective this is good to go. Please let me know if you want any additional changes. I'm available tonight and tomorrow.

garydgregory:
Why is this:
```
        final StringBuilder sb = new StringBuilder();
        sb.append("abcdefghijklmnop");
        final byte[] random = sb.toString().getBytes();
```
Not just:
```
final byte[] random = "abcdefghijklmnop".getBytes();
```
?
I updated git master.

cpreisler:
No specific reason. Your change will work also.

PR #438, opened 2023-11-13
COMPRESS-632: Check for invalid PAX values in TarArchiveEntry

yakovsh:
This is related to the fuzzing work in COMPRESS-632. The PR adds a validation check for time values in PAX headers that are valid Long values but out of range for java.time.Instant.

https://issues.apache.org/jira/browse/COMPRESS-632

### DIFF
--- a/src/main/java/org/apache/commons/compress/archivers/tar/TarArchiveEntry.java
+++ b/src/main/java/org/apache/commons/compress/archivers/tar/TarArchiveEntry.java
@@ -290,8 +290,8 @@ public class TarArchiveEntry implements ArchiveEntry, TarConstants, EntryStreamO
         try {
             return Instant.ofEpochSecond(seconds, nanos);
         } catch (DateTimeException | ArithmeticException e) {
-            // DateTimeException if the instant exceeds the maximum or minimum instant
-            // ArithmeticException if numeric overflow occurs
+            // DateTimeException: Thrown if the instant exceeds the maximum or minimum instant.
+            // ArithmeticException: Thrown if numeric overflow occurs.
             throw new IOException("Corrupted PAX header. Time field value is invalid '" + value + "'", e);
         }
     }


garydgregory:
Hello @yakovsh 
Thank you for your PR.
Can you point to the TAR or PAX specifications for the legal value range? If there are legal values that we cannot represent in a Java Instant, then we need to not blow up but instead do... something else IMO. This range, whatever it is, should be documented in the code.
TY!

garydgregory:
We are not dealing with ints or longs here though, we are creating a time object based on a string, so maybe this is not what we need to worry about. It could be that using a FileTime is just wrong in this case, but this seems like the best type we have in the JRE. In the meantime, I am considering the following as more bullet-proof:
```
diff --git a/src/main/java/org/apache/commons/compress/archivers/tar/TarArchiveEntry.java b/src/main/java/org/apache/commons/compress/archivers/tar/TarArchiveEntry.java
index 360f940..eba51fc 100644
--- a/src/main/java/org/apache/commons/compress/archivers/tar/TarArchiveEntry.java
+++ b/src/main/java/org/apache/commons/compress/archivers/tar/TarArchiveEntry.java
@@ -30,6 +30,7 @@
 import java.nio.file.attribute.DosFileAttributes;
 import java.nio.file.attribute.FileTime;
 import java.nio.file.attribute.PosixFileAttributes;
+import java.time.DateTimeException;
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Collections;
@@ -286,7 +287,13 @@
         final BigDecimal epochSeconds = new BigDecimal(value);
         final long seconds = epochSeconds.longValue();
         final long nanos = epochSeconds.remainder(BigDecimal.ONE).movePointRight(9).longValue();
-        return Instant.ofEpochSecond(seconds, nanos);
+        try {
+            return Instant.ofEpochSecond(seconds, nanos);
+        } catch (DateTimeException | ArithmeticException e) {
+            // DateTimeException if the instant exceeds the maximum or minimum instant
+            // ArithmeticException if numeric overflow occurs
+            throw new IOException("Corrupted PAX header. Time field value is invalid '" + value + "'", e);
+        }
     }
 
     /** The entry's name. */
```

WDYT?

PR #439, opened 2023-11-13
COMPRESS-632: Fix for zero size headers in ArjInputStream

yakovsh:
This adds a validation check for detecting when the header size in ARJ files is set to zero and ArjInputStream fails to parse the file. Related to fuzzing work tracked under COMPRESS-632.


garydgregory:
Hello @yakovsh 
Thank you for the PR.
If we thrown an exception, we should document what part of the ARJ specification the file violates.
Otherwise, what about:
```
diff --git a/src/main/java/org/apache/commons/compress/utils/IOUtils.java b/src/main/java/org/apache/commons/compress/utils/IOUtils.java
index 15dfcc1..88b6305 100644
--- a/src/main/java/org/apache/commons/compress/utils/IOUtils.java
+++ b/src/main/java/org/apache/commons/compress/utils/IOUtils.java
@@ -165,7 +165,7 @@
         if (bufferSize < 1) {
             throw new IllegalArgumentException("bufferSize must be bigger than 0");
         }
-        final byte[] buffer = new byte[(int) Math.min(bufferSize, len)];
+        final byte[] buffer = new byte[(int) Math.min(bufferSize, Math.max(0, len))];
         int n = 0;
         long count = 0;
         while (count < len && -1 != (n = input.read(buffer, 0, (int) Math.min(len - count, buffer.length)))) {
```
?
The above without the change to `ArjArchiveInputStream` still passes the new test. So either the test is not strict enough or the file is legal.

WDYT?

I think I'll bulet-proof the `IOUtils` method anyway.

garydgregory (changes requested):
See my comment. Also, put the new test fixture somewhere in in `src/test/resources/org/apache/commons/compress` instead of the root of the resources, it's a bit cluttered as it is.  Maybe in a new folder `src/test/resources/org/apache/commons/compress/arj`

PR #441, opened 2023-11-14
COMPRESS-632: Fixes for CPIO bugs

yakovsh:
This PR adds a try/catch for long value parsing in CpioArchiveInputStream and for the overflow in one of the methods in CpioArchiveEntry. Both of these fixes are related to the fuzzing work tracked under COMPRESS-632.


src/test/java/org/apache/commons/compress/archivers/cpio/CpioArchiveEntryTest.java
```diff
@@ -0,0 +1,34 @@
...
+
+import static org.junit.Assert.assertThrows;
+
+public class CpioArchiveEntryTest {
+    @Test
+    public void testGetHeaderPadCountOverflow() throws Exception {
+        CpioArchiveEntry entry = new CpioArchiveEntry(CpioConstants.FORMAT_NEW);
+        entry.setName("test name");
+        assertThrows(IllegalArgumentException.class, () -> {
+            entry.getHeaderPadCount(Long.MAX_VALUE);
```
garydgregory, line 31: You don't need a block here.

src/main/java/org/apache/commons/compress/archivers/cpio/CpioArchiveInputStream.java
```diff
@@ -406,7 +406,11 @@ public int read(final byte[] b, final int off, final int len)
     private long readAsciiLong(final int length, final int radix)
             throws IOException {
         final byte[] tmpBuffer = readRange(length);
-        return Long.parseLong(ArchiveUtils.toAsciiString(tmpBuffer), radix);
+        try {
+            return Long.parseLong(ArchiveUtils.toAsciiString(tmpBuffer), radix);
+        } catch(NumberFormatException ex) {
```
garydgregory, line 411: Fix formatting

garydgregory (changes requested):
@yakovsh 
Please see my comments

src/main/java/org/apache/commons/compress/archivers/cpio/CpioArchiveInputStream.java
garydgregory: You need a space after "catch" and before "(". I'll have to fix our checkstyle configuration...

garydgregory (changes requested):
Hello @yakovsh 
Thank you for your PR.
Please rebase on git master, which now implements Checkstyle.
TY!

src/main/java/org/apache/commons/compress/archivers/cpio/CpioArchiveEntry.java
```diff
@@ -565,7 +565,12 @@ public int getHeaderPadCount(final long nameSize) {
         }
         int size = this.headerSize + 1; // Name has terminating null
         if (name != null) {
-            size = ExactMath.add(size, nameSize);
+            try {
+                size = ExactMath.add(size, nameSize);
```
garydgregory, line 569: There a dozen call sites of `ExactMath#add()`; surely it would be redundant to duplicate this pattern. Why not create a wrapper method in `ExactMath`?

garydgregory (changes requested):
Hi @yakovsh 
Thank you for your PR.
See my comment.

src/test/java/org/apache/commons/compress/archivers/cpio/CpioArchiveEntryTest.java
```diff
@@ -0,0 +1,32 @@
...
+package org.apache.commons.compress.archivers.cpio;
+
+import static org.junit.Assert.assertThrows;
+
+import org.junit.jupiter.api.Test;
+
+public class CpioArchiveEntryTest {
+    @Test
+    public void testGetHeaderPadCountOverflow() throws Exception {
+        CpioArchiveEntry entry = new CpioArchiveEntry(CpioConstants.FORMAT_NEW);
```
garydgregory, line 28: Use final where you can.

garydgregory (changes requested):
Hello @yakovsh 
A minor comment needs to be addressed. TY!

src/main/java/org/apache/commons/compress/archivers/cpio/CpioArchiveInputStream.java
```diff
@@ -362,7 +362,11 @@ public int read(final byte[] b, final int off, final int len) throws IOException
 
     private long readAsciiLong(final int length, final int radix) throws IOException {
         final byte[] tmpBuffer = readRange(length);
-        return Long.parseLong(ArchiveUtils.toAsciiString(tmpBuffer), radix);
+        try {
```
garydgregory, line 365: Hi @yakovsh 
It turns out, we do this kind of Long parse-and-rethrow-IOException in at least three other places. I'd think this should be refactored. Also, in total, there are ten call sites for Long parsing. Shouldn't all these be addressed uniformly in this PR? There are also Integer.parseXXX() calls that likely need the same treatment. WDYT?

garydgregory (changes requested):
Hi @yakovsh 
Please see my comment.

src/main/java/org/apache/commons/compress/archivers/cpio/CpioArchiveInputStream.java
garydgregory: Hello @yakovsh 
It can be moved, sure. I would not use ExactMath since that's supposed to be a wrapper for Math exact APIs. If the refactored new API can be package-private, then please do that, if not, we'll have to live with a larger public API footprint. TY!

src/main/java/org/apache/commons/compress/archivers/tar/TarArchiveEntry.java
```diff
@@ -1634,19 +1627,19 @@ private void processPaxHeader(final String key, final String val, final Map<Stri
             setLinkName(val);
             break;
         case "gid":
-            setGroupId(Long.parseLong(val));
+            setGroupId(ParsingUtils.parseLongValue(val, 10));
```
garydgregory, line 1630: The PR now add this "10" magic number in many places. Is it ever NOT 10? Or, do we need a specialized version of `parseLongValue` as `parseLongValueBase10`?

garydgregory (changes requested):
Hi @yakovsh 
Thank you for your update. Please see my comment.

src/main/java/org/apache/commons/compress/harmony/pack200/Pack200Adapter.java
```diff
@@ -41,11 +42,11 @@ public void addPropertyChangeListener(final PropertyChangeListener listener) {
      *
      * @param value Completion between 0..1.
      */
-    protected void completed(final double value) {
+    protected void completed(final double value) throws IOException {
```
garydgregory, line 45: Hello @yakovsh 
Thank for your updates.
Note that if you add or remove exceptions to public or protected method signatures, you must update their Javadoc comments ;-)
TY

src/main/java/org/apache/commons/compress/utils/ParsingUtils.java
```diff
@@ -0,0 +1,83 @@
...
+ * limitations under the License.
+ */
+package org.apache.commons.compress.utils;
+
+import java.io.IOException;
+
+/**
+ * Utility methods for parsing data and converting it to other formats.
+ *
+ * @since 1.26
```
garydgregory, line 24: Note that the next version will be "1.26.0" as opposed to "1.26".

src/main/java/org/apache/commons/compress/utils/ParsingUtils.java
```diff
@@ -0,0 +1,83 @@
...
+ */
+public final class ParsingUtils {
+    /**
+     * Tries to parse the provided string value to an Integer, assuming a base-10 radix
+     *
+     * @param value string value to parse
+     * @return parsed value as an int
+     * @throws IOException
```
garydgregory, line 32: This Javadoc is incomplete because the throws tag is missing a description.

src/main/java/org/apache/commons/compress/utils/ParsingUtils.java
```diff
@@ -0,0 +1,83 @@
...
+     * @param value string value to parse
+     * @return parsed value as a long
+     * @throws IOException
+     */
+    public static long parseLongValue(final String value) throws IOException {
+        return parseLongValue(value, 10);
+    }
+
+    /**
+     * Tries to parse the provided string value to a Long
```
garydgregory, line 66: I would rephrase these comments from "Tries..." to "Parses the given Foo into a Bar". All method "try" to do something ;-) In general, for these types of methods, I use the same verb to start the comment as the method name. This feels consistent and drives to the reader what is the main point of the method.

src/test/java/org/apache/commons/compress/utils/ParsingUtilsTest.java
```diff
@@ -0,0 +1,54 @@
...
+    @ParameterizedTest
+    @ValueSource(strings = {Integer.MIN_VALUE + "1", "x.x", "9e999", "1.1", "one", Integer.MAX_VALUE + "1"})
+    public void testParseIntValueInvalidValues(final String value) {
+        assertThrows(IOException.class, () -> ParsingUtils.parseIntValue(value, 10));
+    }
+
+    @ParameterizedTest
+    @ValueSource(strings = {Integer.MIN_VALUE + "", "-1", "1", "123456", Integer.MAX_VALUE + ""})
+    public void testParseIntValueValidValues(final String value) {
+        assertDoesNotThrow(() -> ParsingUtils.parseIntValue(value, 10));
```
garydgregory, line 39: We do not need the additional layer of `assertDoesNotThrow` for test methods that only test a single call.

garydgregory (changes requested):
Hello @yakovsh 
Thank you for your updates. More comments ;-)

src/test/java/org/apache/commons/compress/utils/ParsingUtilsTest.java
```diff
@@ -0,0 +1,58 @@
...
+    @ParameterizedTest
+    @ValueSource(strings = {Integer.MIN_VALUE + "1", "x.x", "9e999", "1.1", "one", Integer.MAX_VALUE + "1"})
+    public void testParseIntValueInvalidValues(final String value) {
+        assertThrows(IOException.class, () -> ParsingUtils.parseIntValue(value, 10));
+    }
+
+    @ParameterizedTest
+    @ValueSource(strings = {Integer.MIN_VALUE + "", "-1", "1", "123456", Integer.MAX_VALUE + ""})
+    public void testParseIntValueValidValues(final String value) throws Exception {
+        ParsingUtils.parseIntValue(value, 10);
```
garydgregory, line 43: This asserts that the code does not blow up, only. You should assert the expected result.

src/test/java/org/apache/commons/compress/utils/ParsingUtilsTest.java
```diff
@@ -0,0 +1,58 @@
...
+    @ParameterizedTest
+    @ValueSource(strings = {Long.MIN_VALUE + "1", "x.x", "9e999", "1.1", "one", Long.MAX_VALUE + "1"})
+    public void testParseLongValueInvalidValues(final String value) {
+        assertThrows(IOException.class, () -> ParsingUtils.parseLongValue(value, 10));
+    }
+
+    @ParameterizedTest
+    @ValueSource(strings = {Long.MIN_VALUE + "", "-1", "1", "12345678901234", Long.MAX_VALUE + ""})
+    public void testParseLongValueValidValues(final String value) throws Exception {
+        ParsingUtils.parseLongValue(value, 10);
```
garydgregory, line 55: This asserts that the code does not blow up, only. You should assert the expected result.

garydgregory (changes requested):
Hello @yakovsh 
Please see my comments
TY!

PR #442, opened 2023-11-15
COMPRESS-632: Fixes for dump file parsing

yakovsh:
This PR adds a few fixes and unit tests for dump file parsing, related to fuzzing work in COMPRESS-632.


garydgregory:
Please hold off creating PRs while I add checkstyle to the build, it will just cause conflicts; -)

garydgregory (changes requested):
Hello @yakovsh 
Thank you for your PR.
Please rebase on git master, which now implements Checkstyle.
TY!

src/main/java/org/apache/commons/compress/archivers/dump/DumpArchiveInputStream.java
```diff
@@ -526,7 +526,7 @@ private void readDirectoryEntry(DumpArchiveEntry entry) throws IOException {
 
             final byte[] peekBytes = raw.peek();
 
-            if (!DumpArchiveUtil.verify(peekBytes)) {
+            if (peekBytes == null || !DumpArchiveUtil.verify(peekBytes)) {
```
garydgregory, line 529: Why not put this check in `DumpArchiveUtil.verify(byte[])`?

garydgregory (changes requested):
@yakovsh 
Thank you for your PR. Please see my comment.

src/main/java/org/apache/commons/compress/archivers/dump/TapeInputStream.java
```diff
@@ -294,6 +294,9 @@ public void resetBlockSize(final int recsPerBlock, final boolean isCompressed) t
             throw new IOException("Block with " + recsPerBlock + " records found, must be at least 1");
         }
         blockSize = RECORD_SIZE * recsPerBlock;
+        if (blockSize < 1) {
+            throw new IOException("Block size " + blockSize + " cannot be negative");
```
garydgregory, line 298: Either the test or the error message is incorrect. If the test is correct then the error message should say `"Block size cannot be less than or equal to 0: " + blockSize"`

garydgregory (changes requested):
Hi @yakovsh 
Please see my comment. TY!

src/main/java/org/apache/commons/compress/archivers/dump/DumpArchiveConstants.java
```diff
@@ -26,7 +26,7 @@ public final class DumpArchiveConstants {
      * The type of compression.
      */
     public enum COMPRESSION_TYPE {
-        ZLIB(0), BZLIB(1), LZO(2);
+        UNKNOWN(-1), ZLIB(0), BZLIB(1), LZO(2);
```
garydgregory, line 29: Is this required? Does it avoid an NPE? I do not see it used outside of this enum,

PR #446, opened 2023-11-25
COMPRESS-652: Support converting an existing ZIP file to a ZipArchiveOutputStream for appending

nathell:
Description and rationale in a JIRA ticket: https://issues.apache.org/jira/projects/COMPRESS/issues/COMPRESS-652.

In addition to adding a test, I’ve verified manually that this works both for small (non-zip64) files and zip64 files with >64K entries in them. While it works, instantiating a ZIPFile on the latter seems to have a significant memory overhead. I intend to improve it, perhaps in another PR.


garydgregory:
This seems to me like too much of a niche use case.
Let's see what others think.

garydgregory:
Closing.

PR #455, opened 2023-12-27
COMPRESS-653: Fix split archive updating incorrect file

kvr000:
Updates local file header metadata even for split archive, so it's readable by `zip -FF` utility.

### DIFF
--- a/src/main/java/org/apache/commons/compress/archivers/zip/FileRandomAccessOutputStream.java
+++ b/src/main/java/org/apache/commons/compress/archivers/zip/FileRandomAccessOutputStream.java
@@ -1,0 +1,79 @@
+/*
+ *  Licensed to the Apache Software Foundation (ASF) under one or more
+ *  contributor license agreements.  See the NOTICE file distributed with
+ *  this work for additional information regarding copyright ownership.
+ *  The ASF licenses this file to You under the Apache License, Version 2.0
+ *  (the "License"); you may not use this file except in compliance with
+ *  the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.apache.commons.compress.archivers.zip;
+
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.nio.channels.FileChannel;
+import java.nio.file.OpenOption;
+import java.nio.file.Path;
+import java.nio.file.StandardOpenOption;
+
+
+/**
+ * {@link RandomAccessOutputStream} implementation based on Path file.
+ */
+class FileRandomAccessOutputStream extends RandomAccessOutputStream {
+
+    private final FileChannel channel;
+
+    private long position = 0L;
+
+    FileRandomAccessOutputStream(final Path file) throws IOException {
+        this(file, StandardOpenOption.CREATE, StandardOpenOption.TRUNCATE_EXISTING, StandardOpenOption.WRITE);
+    }
+
+    FileRandomAccessOutputStream(final Path file, OpenOption... options) throws IOException {
+        this(FileChannel.open(file, options));
+    }
+
+    FileRandomAccessOutputStream(final FileChannel channel) throws IOException {
+        this.channel = channel;
+    }
+
+    FileChannel channel() {
+        return channel;
+    }
+
+    @Override
+    public synchronized void write(final byte[] b, final int off, final int len) throws IOException {
+        ZipIoUtil.writeFully(this.channel, ByteBuffer.wrap(b, off, len));
+        position += len;
+    }
+
+    @Override
+    public synchronized long position() {
+        return position;
+    }
+
+    @Override
+    public void writeFullyAt(final byte[] b, final int off, final int len, final long atPosition) throws IOException {
+        ByteBuffer buf = ByteBuffer.wrap(b, off, len);
+        for (long currentPos = atPosition; buf.hasRemaining(); ) {
+            int written = this.channel.write(buf, currentPos);
+            if (written <= 0) {
+                throw new IOException("Failed to fully write to file: written=" + written);
+            }
+            currentPos += written;
+        }
+    }
+
+    @Override
+    public void close() throws IOException {
+        channel.close();
+    }
+}
--- a/src/main/java/org/apache/commons/compress/archivers/zip/FileRandomAccessOutputStream.java
+++ b/src/main/java/org/apache/commons/compress/archivers/zip/FileRandomAccessOutputStream.java
@@ -31,7 +31,7 @@ class FileRandomAccessOutputStream extends RandomAccessOutputStream {
 
     private final FileChannel channel;
 
-    private long position = 0L;
+    private long position;
 
     FileRandomAccessOutputStream(final Path file) throws IOException {
         this(file, StandardOpenOption.CREATE, StandardOpenOption.TRUNCATE_EXISTING, StandardOpenOption.WRITE);
--- a/src/main/java/org/apache/commons/compress/archivers/zip/RandomAccessOutputStream.java
+++ b/src/main/java/org/apache/commons/compress/archivers/zip/RandomAccessOutputStream.java
@@ -1,0 +1,70 @@
+/*
+ *  Licensed to the Apache Software Foundation (ASF) under one or more
+ *  contributor license agreements.  See the NOTICE file distributed with
+ *  this work for additional information regarding copyright ownership.
+ *  The ASF licenses this file to You under the Apache License, Version 2.0
+ *  (the "License"); you may not use this file except in compliance with
+ *  the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.apache.commons.compress.archivers.zip;
+
+import java.io.IOException;
+import java.io.OutputStream;
+
+
+/**
+ * Abstraction over OutputStream which also allows random access writes.
+ */
+abstract class RandomAccessOutputStream extends OutputStream {
+
+    /**
+     * Provides current position in output.
+     *
+     * @return
+     *      current position.
+     */
+    public abstract long position() throws IOException;
+
+    /**
+     * Writes given data to specific position.
+     *
+     * @param position
+     *      position in the stream
+     * @param b
+     *      data to write
+     * @throws IOException
+     *      when write fails.
+     */
+    public void writeFullyAt(final byte[] b, final long position) throws IOException {
+        writeFullyAt(b, 0, b.length, position);
+    }
+
+    /**
+     * Writes given data to specific position.
+     *
+     * @param position
+     *      position in the stream
+     * @param b
+     *      data to write
+     * @param off
+     *      offset of the start of data in param b
+     * @param len
+     *      the length of data to write
+     * @throws IOException
+     *      when write fails.
+     */
+    abstract void writeFullyAt(byte[] b, int off, int len, long position) throws IOException;
+
+    @Override
+    public void write(final int b) throws IOException {
+        write(new byte[]{ (byte) b });
+    }
+}
--- a/src/main/java/org/apache/commons/compress/archivers/zip/SeekableChannelRandomAccessOutputStream.java
+++ b/src/main/java/org/apache/commons/compress/archivers/zip/SeekableChannelRandomAccessOutputStream.java
@@ -1,0 +1,62 @@
+/*
+ *  Licensed to the Apache Software Foundation (ASF) under one or more
+ *  contributor license agreements.  See the NOTICE file distributed with
+ *  this work for additional information regarding copyright ownership.
+ *  The ASF licenses this file to You under the Apache License, Version 2.0
+ *  (the "License"); you may not use this file except in compliance with
+ *  the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.apache.commons.compress.archivers.zip;
+
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.nio.channels.SeekableByteChannel;
+
+
+/**
+ * {@link RandomAccessOutputStream} implementation for SeekableByteChannel.
+ */
+class SeekableChannelRandomAccessOutputStream extends RandomAccessOutputStream {
+
+    private final SeekableByteChannel channel;
+
+    private long position;
+
+    SeekableChannelRandomAccessOutputStream(final SeekableByteChannel channel) {
+        this.channel = channel;
+    }
+
+    @Override
+    public synchronized void write(final byte[] b, final int off, final int len) throws IOException {
+        ZipIoUtil.writeFully(this.channel, ByteBuffer.wrap(b, off, len));
+    }
+
+    @Override
+    public synchronized long position() throws IOException {
+        return channel.position();
+    }
+
+    @Override
+    public synchronized void writeFullyAt(final byte[] b, final int off, final int len, final long position) throws IOException {
+        long saved = channel.position();
+        try {
+            channel.position(position);
+            ZipIoUtil.writeFully(channel, ByteBuffer.wrap(b, off, len));
+        } finally {
+            channel.position(saved);
+        }
+    }
+
+    @Override
+    public synchronized void close() throws IOException {
+        channel.close();
+    }
+}
--- a/src/main/java/org/apache/commons/compress/archivers/zip/ZipFile.java
+++ b/src/main/java/org/apache/commons/compress/archivers/zip/ZipFile.java
@@ -59,9 +59,9 @@ import org.apache.commons.compress.utils.IOUtils;
 import org.apache.commons.compress.utils.InputStreamStatistics;
 import org.apache.commons.compress.utils.SeekableInMemoryByteChannel;
 import org.apache.commons.io.Charsets;
+import org.apache.commons.io.FilenameUtils;
 import org.apache.commons.io.build.AbstractOrigin.ByteArrayOrigin;
 import org.apache.commons.io.build.AbstractStreamBuilder;
-import org.apache.commons.io.FilenameUtils;
 import org.apache.commons.io.input.CountingInputStream;
 
 /**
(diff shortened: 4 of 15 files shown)


garydgregory:
@kvr000 
-1: This PR breaks binary compatibility. 
Run 'mvn' before you push.

src/main/java/org/apache/commons/compress/archivers/zip/FileRandomAccessOutputStream.java
```diff
@@ -0,0 +1,79 @@
...
+/**
+ * {@link RandomAccessOutputStream} implementation based on Path file.
+ */
+class FileRandomAccessOutputStream extends RandomAccessOutputStream {
+
+    private final FileChannel channel;
+
+    private long position = 0L;
+
+    public FileRandomAccessOutputStream(Path file) throws IOException {
```
garydgregory, line 40: Use `final` where you can.

src/main/java/org/apache/commons/compress/archivers/zip/FileRandomAccessOutputStream.java
```diff
@@ -0,0 +1,79 @@
...
+import java.nio.channels.FileChannel;
+import java.nio.file.OpenOption;
+import java.nio.file.Path;
+import java.nio.file.StandardOpenOption;
+
+
+/**
+ * {@link RandomAccessOutputStream} implementation based on Path file.
+ */
+class FileRandomAccessOutputStream extends RandomAccessOutputStream {
```
garydgregory, line 34: This new class needs a unit test to prove this class behaves as an output stream and that all APIs do what they say.

src/main/java/org/apache/commons/compress/archivers/zip/FileRandomAccessOutputStream.java
```diff
@@ -0,0 +1,79 @@
...
+        position += len;
+    }
+
+    @Override
+    public synchronized long position() {
+        return position;
+    }
+
+    @Override
+    public void writeFullyAt(final byte[] b, final int off, final int len, long atPosition) throws IOException {
```
garydgregory, line 64: Calling the API `write` should be enough, especially if Javadocs are comprehensive.

src/main/java/org/apache/commons/compress/archivers/zip/RandomAccessOutputStream.java
```diff
@@ -0,0 +1,70 @@
...
+package org.apache.commons.compress.archivers.zip;
+
+import java.io.IOException;
+import java.io.OutputStream;
+
+
+/**
+ * Abstraction over OutputStream which also allows random access writes.
+ */
+abstract class RandomAccessOutputStream extends OutputStream {
```
garydgregory, line 26: This new class needs a unit test to prove this class behaves as an output stream and that all APIs do what they say.

src/main/java/org/apache/commons/compress/archivers/zip/SeekableChannelRandomAccessOutputStream.java
```diff
@@ -0,0 +1,69 @@
...
+import java.nio.ByteBuffer;
+import java.nio.channels.FileChannel;
+import java.nio.channels.SeekableByteChannel;
+import java.nio.file.Path;
+
+
+/**
+ * {@link RandomAccessOutputStream} implementation for SeekableByteChannel.
+ */
+class SeekableChannelRandomAccessOutputStream extends RandomAccessOutputStream {
```
garydgregory, line 32: This new class needs a unit test to prove this class behaves as an output stream and that all APIs do what they say.

src/main/java/org/apache/commons/compress/archivers/zip/SeekableChannelRandomAccessOutputStream.java
```diff
@@ -0,0 +1,69 @@
...
+
+    private final SeekableByteChannel channel;
+
+    private long position = 0L;
```
garydgregory, line 36: Don't initialize instance variables to their default values.

src/main/java/org/apache/commons/compress/utils/IOUtils.java
```diff
@@ -275,6 +277,49 @@ public static byte[] readRange(final ReadableByteChannel input, final int len) t
...
+     * Writes full buffer to channel.
+     *
+     * @param channel
+     *      channel to write to
+     * @param buf
+     *      buffer to write
+     * @throws IOException
+     *      when writing fails
+     */
+    public static void writeFully(SeekableByteChannel channel, ByteBuffer buf) throws IOException {
```
garydgregory, line 290: `writeFully` -> `write` or is a "partial write" expected default behavior? That would be odd. Flip the arguments such that the source comes first and the destination second. IOW, "Writes from Foo to Bar", which happens to be the phrasing you use in the Javadoc ;-)

src/main/java/org/apache/commons/compress/utils/IOUtils.java
```diff
@@ -275,6 +277,49 @@ public static byte[] readRange(final ReadableByteChannel input, final int len) t
...
+     * @param channel
+     *      channel to write to
+     * @param buf
+     *      buffer to write
+     * @param position
+     *      position to write at
+     * @throws IOException
+     *      when writing fails
+     */
+    public static void writeFullyAt(FileChannel channel, ByteBuffer buf, long position) throws IOException {
```
garydgregory, line 312: See above. All new public and protected elements shoyuld have an `@since` Javadoc tag.

garydgregory (changes requested):
Hello @kvr000 !
- Thank you for your PR.
- Do fix the build and see my comments scattered throughout.
- Big picture: I am considering adding some random IO classes to Apache Commons IO like Apache PDFBox has, see https://github.com/apache/pdfbox/tree/trunk/io/src/main/java/org/apache/pdfbox/io
- In light of this, it is good that the random IO classes you've added are package-protected. This will let us possibly replace this implementation with this very code ported to Commons IO or with code from PDFBox ported to Commons IO. We do not want Commons Compress to depend on PDFBox, especially since PDFBox already depends on Commons IO.
- To reduce the new public API surface, please move the new IOUtils method into package-protected place in the `zip` package.
- I'll study all of this some more and see what is generic enough to add and reuse from Commons IO.

garydgregory:
@kvr000 
Ping?

kvr000:
> @kvr000 Ping?

Thanks for your feedback, @garydgregory , and sorry for the delay.  I'll take a look next few days.

src/main/java/org/apache/commons/compress/utils/IOUtils.java
kvr000: Moved to package private ZipIoUtil as requested above.

src/main/java/org/apache/commons/compress/utils/IOUtils.java
kvr000: The methods in `SeekableByteChannel` have `write` semantics to attempt to write and return as soon as the `write` is attempted.  So I wanted to avoid naming conflict with those and be explicit that this writes full buffer or fails.
Other libraries use the same semantics, such as https://hadoop.apache.org/docs/stable/api/org/apache/hadoop/io/IOUtils.html#writeFully-java.nio.channels.FileChannel-java.nio.ByteBuffer-long-

The example also specifies the same order of parameters which in my opinion makes sense - it operates on object which comes as first parameter, not really in a sense of source of destination.  I can amend this but it's inconsistent with other libraries.

src/main/java/org/apache/commons/compress/archivers/zip/SeekableChannelRandomAccessOutputStream.java
kvr000: I wanted to be explicit here that the position starts at `0`, despite it technically gets `0` by JDK.

src/main/java/org/apache/commons/compress/archivers/zip/FileRandomAccessOutputStream.java
kvr000: See the comment later - the default semantics on `SeekableByteChannel` is to write part of the buffer, not necessarily full.  It's consistent with naming convention in other libraries such as https://hadoop.apache.org/docs/stable/api/org/apache/hadoop/io/IOUtils.html#writeFully-java.nio.channels.FileChannel-java.nio.ByteBuffer-long-

src/main/java/org/apache/commons/compress/archivers/zip/FileRandomAccessOutputStream.java
kvr000: Right, added.

src/main/java/org/apache/commons/compress/archivers/zip/SeekableChannelRandomAccessOutputStream.java
kvr000: Added.

src/main/java/org/apache/commons/compress/archivers/zip/RandomAccessOutputStream.java
kvr000: Added, just the method which actually is provided.

src/main/java/org/apache/commons/compress/archivers/zip/FileRandomAccessOutputStream.java
kvr000: I added `final` to few places, hope I didn't miss something.

kvr000:
> * Big picture: I am considering adding some random IO classes to Apache Commons IO like Apache PDFBox has, see https://github.com/apache/pdfbox/tree/trunk/io/src/main/java/org/apache/pdfbox/io

That totally makes sense, each library provides its own IOUtils currently.  I'd suggest few things:
1. FileChannel is typically more meaningful than SeekableByteChannel.  Even here in ZipFile and related, the SeekableByteChannel is used in a sense of FileChannel.  I provided fix for this a while ago which allowed lock-free reading from multiple ZipFile entries in parallel but there is the same work in other parts.
2. The JDK unfortunately misses interface for position_read and position_write methods which makes custom implementations more tricky.  I was addressing this issue in https://github.com/dryuf/dryuf-bigio (FlatChannel), you may consider this for inspiration when working on commons-io stuff.
3. The PdfBox in my opinion is quite complicated and may suffer from diamond inheritance problem.  The above may be more straightforward and more efficient.


> * In light of this, it is good that the random IO classes you've added are package-protected. This will let us possibly replace this implementation with this very code ported to Commons IO or with code from PDFBox ported to Commons IO. We do not want Commons Compress to depend on PDFBox, especially since PDFBox already depends on Commons IO.
> * To reduce the new public API surface, please move the new IOUtils method into package-protected place in the `zip` package.

That was done.

> * I'll study all of this some more and see what is generic enough to add and reuse from Commons IO.

It may require more work to make it more generic.  For now the package private is probably better choice.

src/main/java/org/apache/commons/compress/archivers/zip/FileRandomAccessOutputStream.java
garydgregory: @kvr000 
The example you use does not use the "At" in the method name postfix, which feels awkward to me, and we do not see it in the JRE. The fact that the method writes its argument at a position is implied by the fact that the position is passed in. I would not use "At" in the method name or parameter name.

src/main/java/org/apache/commons/compress/archivers/zip/RandomAccessOutputStream.java
garydgregory, line 26: Hello @kvr000 and Happy New Year!
The API design does not feel quite right to me. A `SeekableByteChannel` has `position()` and `position(long)`, and `RandomAccessFile` has `getFilePointer()` and `seek(long)`. In both cases, there is one API for "where am I?" and another for "go to this location". I think we should have the same here. This avoids adding a "position" argument everytime a new API is needed in this method. It also could let the implementation decide if a write is "full" or not. Avoiding the need for "fully" APIs in the first place. Thoughts?
@elharo, any thoughts here?

src/main/java/org/apache/commons/compress/archivers/zip/ZipSplitOutputStream.java
```diff
@@ -42,14 +48,20 @@ final class ZipSplitOutputStream extends OutputStream {
...
     private final long splitSize;
+    private long totalPosition;
     private int currentSplitSegmentIndex;
     private long currentSplitSegmentBytesWritten;
     private boolean finished;
     private final byte[] singleByte = new byte[1];
 
+    private List<Long> diskToPosition = new ArrayList<>();
+
+    private TreeMap<Long, Path> positionToFiles = new TreeMap<>();
```
garydgregory, line 63: Is there going to be an issue here on case-insensitive file systems? Say, Windows, macOS?

src/main/java/org/apache/commons/compress/archivers/zip/RandomAccessOutputStream.java
kvr000: The purpose is to provide append only `OutputStream` with possibility to write to arbitrary position.  The `pread/pwrite` semantics is much more meaningful than `seek/tell` semantics in this case as the caller should not be able to change actual position of `OutputStream` stream.  Apart from following the callers usage, this scales much better in concurrent environment (not necessarily the case here, more useful in `ZipFile` reader which already uses this approach).

I'm not sure I fully followed the second part.  The meaning of `write()` - full or partial - should be defined in interface, and the implementations should behave consistently with the interface.  I used `writeFully` name to avoid conflict with `FileChannel.write` which declares that it may be only partial write and such naming is also consistent with other 3rd-parties libraries.  I agree the `At` can be removed from `writeFullyAt`, though in my opinion it's often better to be explicit with naming rather than relying on different method signature.

src/main/java/org/apache/commons/compress/archivers/zip/ZipSplitOutputStream.java
kvr000: Not here, the name created and name re-opened for `writeFullyAt` will be the same.

The issue may be in the reader but it's rather the opposite - it may fail on case-sensitive filesystem if the names are not consistent.  Perhaps we can check both `.z01` and `.Z01` extensions.

src/main/java/org/apache/commons/compress/archivers/zip/FileRandomAccessOutputStream.java
kvr000: Answered in later comment.  It's possible to remove `At`, though I prefer more explicit names rather than different method signatures.

garydgregory:
@kvr000 
I will review this PR tomorrow. I appreciate your patience.

kvr000:
> @kvr000 I will review this PR tomorrow. I appreciate your patience.

Sure, take your time.

garydgregory (changes requested):
Hello @kvr000 
Thank you for your patience. Instead of adding yet another constructor, I've refactored the code to use a `ZipFile` builder (see `ZipFile.Builderr`). I think this is better than an options object builder. You can add and use the options you need if any for this PR to `ZipFile.Builder`. The options builder should no longer be needed so you can ignore the comments I made there.
TY!

kvr000:
> Hello @kvr000 Thank you for your patience. Instead of adding yet another constructor, I've refactored the code to use a `ZipFile` builder (see `ZipFile.Builderr`). I think this is better than an options object builder. You can add and use the options you need if any for this PR to `ZipFile.Builder`. The options builder should no longer be needed so you can ignore the comments I made there. TY!

Hi Gary, thanks for your response.

The `ZipFile` builder certainly works too, in a way, it's a shortcut to pass parameters directly instead of via another object.  The reason why I chose separate Options object is:
- It's pretty common pattern for complex initialization.  It is actually used in SevenZFileOptions as well.
- The configuration is typically made once and then reused through the application.  Having said that, to make it fully reusable, I should have probably passed the path separately, though `toBuilder().modify().build()` still works.
- It should be more reusable across different archivers, both readers and writers.  Some would support only subset but generally, having single ZipOptions would allow sharing at least among ZipFile, ZipArchiveInputStream and ZipArchiveOutputStream.

I'm fine with the current option as well though.  Please let me know whether the current choice is final and if so, I can adjust my code.

I'd still recommend passing full `Builder` to constructor instead of all arguments separately - it scales better and is less prone to incompatible changes.

src/main/java/org/apache/commons/compress/archivers/zip/ZipFile.java
```diff
@@ -684,6 +715,53 @@ public ZipFile(final String name, final String encoding) throws IOException {
...
+            } else {
+                channel.position(channel.position() + ZipConstants.WORD);
+                ByteBuffer buf = ByteBuffer.allocate(ZipConstants.SHORT);
+                buf.order(ByteOrder.LITTLE_ENDIAN);
+                IOUtils.readFully(channel, buf);
+                buf.flip();
+                numberOfFiles = (buf.getShort() & 0xffff) + 1;
+            }
+            if (numberOfFiles > Math.min(options.maxNumberOfDisks(), Integer.MAX_VALUE)) {
+                throw new IOException("Too many files in zip archive, max=" +
```
garydgregory, line 740: Hello @kvr000 
Either the option name or the error message is wrong: are we creating a bound on files in an archive or a count of volumes ("drives")? Or is there something I am missing?

src/main/java/org/apache/commons/compress/archivers/zip/ZipFile.java
```diff
@@ -684,6 +715,53 @@ public ZipFile(final String name, final String encoding) throws IOException {
...
+                IOUtils.readFully(channel, buf);
+                buf.flip();
+                numberOfFiles = buf.getInt() & 0xffffffffL;
+            } else {
+                channel.position(channel.position() + ZipConstants.WORD);
+                ByteBuffer buf = ByteBuffer.allocate(ZipConstants.SHORT);
+                buf.order(ByteOrder.LITTLE_ENDIAN);
+                IOUtils.readFully(channel, buf);
+                buf.flip();
+                numberOfFiles = (buf.getShort() & 0xffff) + 1;
```
garydgregory, line 737: It looks like other methods in this file use `ZipLong.getValue(ByteBuffer)` to get this value. Should this call site do the same?

src/main/java/org/apache/commons/compress/archivers/zip/ZipFile.java
```diff
@@ -435,6 +444,39 @@ public static void closeQuietly(final ZipFile zipFile) {
 
     private long firstLocalFileHeaderOffset;
 
+
+    public ZipFile(ZipOpenOptions options) throws IOException {
```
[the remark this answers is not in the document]
garydgregory: The rework is in git master.

garydgregory:
> > Hello @kvr000 Thank you for your patience. Instead of adding yet another constructor, I've refactored the code to use a `ZipFile` builder (see `ZipFile.Builderr`). I think this is better than an options object builder. You can add and use the options you need if any for this PR to `ZipFile.Builder`. The options builder should no longer be needed so you can ignore the comments I made there. TY!
> 
> Hi Gary, thanks for your response.
> 
> The `ZipFile` builder certainly works too, in a way, it's a shortcut to pass parameters directly instead of via another object. The reason why I chose separate Options object is:
> 
>     * It's pretty common pattern for complex initialization.  It is actually used in SevenZFileOptions as well.
> 
>     * The configuration is typically made once and then reused through the application.  Having said that, to make it fully reusable, I should have probably passed the path separately, though `toBuilder().modify().build()` still works.
> 
>     * It should be more reusable across different archivers, both readers and writers.  Some would support only subset but generally, having single ZipOptions would allow sharing at least among ZipFile, ZipArchiveInputStream and ZipArchiveOutputStream.
> 
> 
> I'm fine with the current option as well though. Please let me know whether the current choice is final and if so, I can adjust my code.
> 
> I'd still recommend passing full `Builder` to constructor instead of all arguments separately - it scales better and is less prone to incompatible changes.

Hi @kvr000 
I think the best part forward is for you to rebase on git master and adapt to the new Builder. TY!

src/main/java/org/apache/commons/compress/archivers/zip/ZipFile.java
kvr000: There isn't `ZipLong.getValue(ByteBuffer)` method, only `byte[]` overloads.

src/main/java/org/apache/commons/compress/archivers/zip/ZipFile.java
kvr000: Correct, the message was misleading.  I corrected to `Too many disks...` .

PR #459, opened 2024-01-09
COMPRESS-656-EntryPackedSize

valimaties:
Added property and getter for entry packed size in SevenZOutputFile class file.

### DIFF
--- a/src/main/java/org/apache/commons/compress/archivers/sevenz/SevenZOutputFile.java
+++ b/src/main/java/org/apache/commons/compress/archivers/sevenz/SevenZOutputFile.java
@@ -93,6 +93,7 @@ public class SevenZOutputFile implements Closeable {
             }
             compressedCrc32.update(b, off, len);
             fileBytesWritten += len;
+            entryPackedSize +=len;
         }
 
         @Override
@@ -102,6 +103,7 @@ public class SevenZOutputFile implements Closeable {
             channel.write(buffer);
             compressedCrc32.update(b);
             fileBytesWritten++;
+            entryPackedSize++;
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/apache/commons/compress/archivers/sevenz/SevenZOutputFile.java
```diff
@@ -128,6 +130,15 @@ private static <T> Iterable<T> reverse(final Iterable<T> i) {
 
     private AES256Options aes256Options;
 
+    private long entryPackedSize = 0;
+
+    /**
+     * This method returns the packedSize of the entry.
+     */
```
garydgregory, line 137: All new public and protected methods need a Javadoc since tag.

garydgregory (changes requested):
@valimaties 
Thank you for your PR.
Please add tests to cover the new code and see my comments.

src/main/java/org/apache/commons/compress/archivers/sevenz/SevenZOutputFile.java
```diff
@@ -128,6 +130,15 @@ private static <T> Iterable<T> reverse(final Iterable<T> i) {
 
     private AES256Options aes256Options;
 
+    private long entryPackedSize = 0;
```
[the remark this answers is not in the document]
valimaties: Ok, deleted redundant initialization in branch.

src/main/java/org/apache/commons/compress/archivers/sevenz/SevenZOutputFile.java
valimaties: Do I have to write a more detailed information?

valimaties:
Doing some more tests I found that this solution is not good, the packed size is not correct. I will delete this PR.

PR #460, opened 2024-01-11
[COMPRESS-655] Fix TAR directory entries being misinterpreted as files

sschuberth:
As requested by @garydgregory.


sschuberth:
Looks like in

    public boolean isFile() {
        if (file != null) {
            return Files.isRegularFile(file, linkOptions);
        }
        if (linkFlag == LF_OLDNORM || linkFlag == LF_NORMAL) {
            return true;
        }
        return !getName().endsWith("/");
    }

`file` is `null`, but `!getName().endsWith("/")` returns `true`. I'll investigate a bit further.

sschuberth:
Ok, I believe to have a fix.

garydgregory:
I have a fix as well ;-)

sschuberth:
> I have a fix as well ;-)

Is it different from mine?

> Ah! You'll want to run `mvn` (by itself) before pushing to catch all build errors.

Should be all good now.

PS: I hope the test resource I've added is not too large... the uncompressed TAR is 1.9M 🙄

garydgregory:
TY @sschuberth ! Merged.

garydgregory:
https://issues.apache.org/jira/browse/COMPRESS-657

PR #471, opened 2024-01-27
Support preamble garbage in ZipArchiveInputStream

kvr000:
`zip` can ignore any garbage at the beginning of file up to certain size.  This is supported in `ZipFile` but not in `ZipArchiveInputStream`.  This PR fixes that, including correct check for empty archive.

### DIFF
--- a/src/main/java/org/apache/commons/compress/archivers/zip/ZipArchiveInputStream.java
+++ b/src/main/java/org/apache/commons/compress/archivers/zip/ZipArchiveInputStream.java
@@ -177,6 +177,8 @@ public class ZipArchiveInputStream extends ArchiveInputStream<ZipArchiveEntry> i
         }
     }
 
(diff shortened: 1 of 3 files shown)


garydgregory:
Hello @kvr000 
Is this documented anywhere? PREAMBLE_GARBAGE_MAX_SIZE?

kvr000:
> Hello @kvr000 Is this documented anywhere? PREAMBLE_GARBAGE_MAX_SIZE?

I haven't found "official" support for skipping the garbage in the specification.  `zip` program and `ZipFile` from apache compress naturally do it because they go backward from the end of central directory, though `zip` eventually prints a warning about garbage at the beginning.

I used `4096` as it should be enough for typical case when the program execution is added at the beginning: `#!/usr/bin/env -S whatever with the arguments such as java -Xmx32M -jar`

garydgregory:
@kvr000 
Thank you for the clarification; merged.

PR #483, opened 2024-02-20
Remove out of date jar and scripts

beatbrot:
Using the maven wrapper is currently broken since the `org.apache.maven.plugins:maven-enforcer-plugin`  plugin requires at least maven 3.6.3. The committed maven wrapper uses version 3.5.0.

This PR updates the wrapper straight to 3.9.6 which fixes the issue.


garydgregory:
We have jar files in the repository? Gross. I'd rather get rid of this since it will automatically go out of date with the next release, and the next one. I never use it. The GitHub CI doesn't either.

beatbrot:
I mean...personally, I like the wrapper scripts. But if you want, I can change this PR to simply remove the wrapper :) Should I do that?

garydgregory:
Hi @beatbrot 
Let's see if anyone else chimes in who is in love with these scripts...

garydgregory:
> I mean...personally, I like the wrapper scripts. But if you want, I can change this PR to simply remove the wrapper :) Should I do that?

Sure, go for it. TY.

beatbrot:
@garydgregory Done :)

PR #500, opened 2024-03-20
[COMPRESS-674] Validate TarArchiveEntry checksums

fatkudu:
For test file, thought it had 1 (corrupt) directory entry only but still thought it was a tar file

https://issues.apache.org/jira/browse/COMPRESS-674


garydgregory:
TY @fatkudu

PR #512, opened 2024-04-05
COMPRESS-675 -- create unit test for COMPRESS-675

tballison:
unit test for putting two unpack files within an archive

### DIFF
--- a/src/test/java/org/apache/commons/compress/compressors/pack200/Pack200UtilsTest.java
+++ b/src/test/java/org/apache/commons/compress/compressors/pack200/Pack200UtilsTest.java
@@ -18,8 +18,14 @@
  */
 package org.apache.commons.compress.compressors.pack200;
 
+import static org.junit.Assert.assertEquals;
+
 import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileOutputStream;
+import java.io.IOException;
 import java.io.InputStream;
+import java.io.OutputStream;
 import java.nio.file.Files;
 import java.nio.file.StandardCopyOption;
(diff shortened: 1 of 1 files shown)


tballison:
@garydgregory This is either user error or something is really wrong with unpack.

I couldn't get this test to pass with a tar archive stream or a zip archive stream -- different results!

I couldn't get this test to pass with a FIleInputStream, a BufferedInputStream(ByteArrayInputStream) or a Files.newInputStream(archive.toPath)

In the debugger on the Tika issue, I can see that Unpack unwraps down to a ChannelInputStream and then closes the underlying file. I can see similar behavior in this unit test if I open with Files.newInputStream()...so that's what you can use to replicate what we're seeing on Tika.

I'm perplexed that I can't get anything to pass, though... again, probably user error.

tballison:
@garydgregory This is either user error or something is really wrong with unpack.

I couldn't get this test to pass with a tar archive stream or a zip archive stream -- different results!

I couldn't get this test to pass with a FIleInputStream, a BufferedInputStream(ByteArrayInputStream) or a Files.newInputStream(archive.toPath)

In the debugger on the Tika issue, I can see that Unpack unwraps down to a ChannelInputStream and then closes the underlying file. I can see similar behavior in this unit test if I open with Files.newInputStream()...so that's what you can use to replicate what we're seeing on Tika.

I'm perplexed that I can't get anything to pass, though... again, probably user error.

garydgregory:
Hi @tballison 
Would you please rebase on git master? Let's see what the tests say this time around...

PR #571, opened 2024-09-10
ZipArchiveOutputStream.addRawArchiveEntry() should check is2PhaseSource

yujincheng08:
This allows users to add manually compressed raw data to the zip file under an unsupported compression method (e.g., XZ) or encrypt method.

Sample usage:

```kotlin
val data = "hello world".toByteArray()
ZipArchiveOutputStream(File("test.zip")).use { zip ->
    val entry = ZipArchiveEntry("test")
    entry.method = ZipMethod.XZ.code
    entry.crc = CRC32().apply { update(data) }.value
    ByteArrayOutputStream().use { os ->
        XZCompressorOutputStream(os).use {
            it.write(data)
        }
        entry.size = data.size.toLong()
        entry.compressedSize = os.size().toLong()
        zip.addRawArchiveEntry(entry, ByteArrayInputStream(os.toByteArray()))
    }
}
```

Previously these codes would throw an exception about the unsupported compression method.

### DIFF
--- a/src/main/java/org/apache/commons/compress/archivers/zip/ZipArchiveOutputStream.java
+++ b/src/main/java/org/apache/commons/compress/archivers/zip/ZipArchiveOutputStream.java
@@ -505,7 +505,7 @@ public class ZipArchiveOutputStream extends ArchiveOutputStream<ZipArchiveEntry>
         final boolean is2PhaseSource = ae.getCrc() != ZipArchiveEntry.CRC_UNKNOWN && ae.getSize() != ArchiveEntry.SIZE_UNKNOWN
                 && ae.getCompressedSize() != ArchiveEntry.SIZE_UNKNOWN;
         putArchiveEntry(ae, is2PhaseSource);
-        copyFromZipInputStream(rawStream);
+        copyFromZipInputStream(rawStream, is2PhaseSource);
         closeCopiedEntry(is2PhaseSource);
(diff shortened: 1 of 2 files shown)


garydgregory (changes requested):
Hello @yujincheng08 

Thank you for your PR.

This PR is missing tests.

Where is this feature documented?

yujincheng08:
Hi, @garydgregory,

Test added.

And this feature is already documented:

https://github.com/apache/commons-compress/blob/287e94beb6a81ddf111403b4c56ab80f957308b5/src/main/java/org/apache/commons/compress/archivers/zip/ZipArchiveOutputStream.java#L483-L496

According to the document, the `rawStream` may be compressed/encrypted, and if crc, size and compressed size are supplied on the entry, these values will be used as-is. So there's no reason to check if the compression method or encryption method is supported or not when crc, size and compressed size are supplied (we don't need to calculate these values and we don't need to compress or encrypt the data).

garydgregory:
@yujincheng08 
Please do not use star imports, this causes the build checks to fail.
Run `mvn` by itself before you push to catch all build errors.
TY!

yujincheng08:
@garydgregory Thank you. Fixed.

garydgregory:
@yujincheng08 

Thank you for your work. Merged!

PR #573, opened 2024-09-15
Use InputStream.skip to skip stream

yujincheng08:
This can avoid unexpected IO when using InputStream that implements skip


tballison:
InputStream.skip() is dangerous, especially on FileInputStreams. See why commons-io's IOUtils uses {{read()}} under its {{skipFully()}}

https://issues.apache.org/jira/browse/IO-203?focusedCommentId=12842641&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-12842641

garydgregory:
Closing: No reply from OP. TY @tballison for the comment and link.

PR #604, opened 2024-11-15
Adding support for gzip extra subfields.

ddeschenes-1:
In order to put metadata in a gzip extra fields, this library needed to write and read them.

The extras being tagged with 2 bytes, I choose to make it conveninent by using a 2-chars iso-8859-1 (latin1) string
or if you prefer, using only bytes from 0-255 in ascii (single byte charset). It seems more in line with the spirit of the gzip rfc which uses letters.

A bit of convenience was added to the Extra class to list/add/find subfields.


garydgregory:
Hello @ddeschenes-1 

Thank you for your PR

The build breaks due to not having followed our guidelines here: https://github.com/apache/commons-compress#:~:text=Before%20you%20pushing%20a%20PR%2C%20run%20mvn%20(by%20itself)%2C%20this%20runs%20the%20default%20goal%2C%20which%20contains%20all%20build%20checks.

```
[INFO] There are 43 errors reported by Checkstyle 9.3 with /home/runner/work/commons-compress/commons-compress/src/conf/checkstyle.xml ruleset.
```

ddeschenes-1:
Thanks for the heads up. I'll retry next week, likely with less friction this time.

I did use the java convention formatter built in eclipse. I guess it needs tuning.
The mvn call was not obvious enough from "run mvn (by itself)" to NOT call 'mvn test' like I did successfully.
Perhaps a more explicit mvn phase/goal should be in the guide.

I expect issues with the missing jira too (or '(doc)' commit comment.
I couldn't get the jira account. Something's wrong with self signup. We'll see.

I'll arrange what I can.

garydgregory:
Hello @ddeschenes-1 

As documented, running mvn by itself runs the default goal (as defined in the POM) which includes verify and other goals.

If you think our read me file could be better, feel free to create a PR in the commons-build-plugin repository.

I'll be happy to review.

😊 
HTH

garydgregory:
Hello @ddeschenes-1 

Thank you for your updates. While I merged the PR, I had to strip out code that was untested. If you need the APIs, please create a PR that contains the new APIs with unit tests. I also added missing Javadoc comments. 

In the future, please make sure you provide Javadoc for all new code.

TY!

ddeschenes-1:
Definitely need the extra apis for a caller to read the extra!

garydgregory:
Sure, see my previous comment 😀

PR #607, opened 2024-11-17
Remove Commons Lang dependency

ppkarwasz:
Commons Compress version 1.26.0 depends on Commons Lang, which is a 650 KiB dependency. In reality, however, only 4 methods accounting for less than 1% of Commons Lang are used.

We replace the dependency on `commons-lang3` with a local copy of these 4 methods.

### DIFF
--- a/src/main/java/org/apache/commons/compress/internal/lang3/SystemProperties.java
+++ b/src/main/java/org/apache/commons/compress/internal/lang3/SystemProperties.java
@@ -20,14 +20,10 @@ package org.apache.commons.compress.internal.lang3;
 
 public final class SystemProperties {
 
-  public static String getOsName() {
-    try {
-      return System.getProperty("os.name");
-    } catch (final SecurityException e) {
-      return null;
+    public static String getOsName() {
+        return System.getProperty("os.name");
     }
-  }
 
-  private SystemProperties() {
-  }
+    private SystemProperties() {
+    }
 }
--- a/src/main/java/org/apache/commons/compress/internal/lang3/SystemProperties.java
+++ b/src/main/java/org/apache/commons/compress/internal/lang3/SystemProperties.java
@@ -1,0 +1,33 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.commons.compress.internal.lang3;
+
+public final class SystemProperties {
+
+  public static String getOsName() {
+    try {
+      return System.getProperty("os.name");
+    } catch (final SecurityException e) {
+      return null;
+    }
+  }
+
+  private SystemProperties() {
+  }
+}
--- a/src/main/java/org/apache/commons/compress/internal/lang3/reflect/FieldUtils.java
+++ b/src/main/java/org/apache/commons/compress/internal/lang3/reflect/FieldUtils.java
@@ -1,0 +1,61 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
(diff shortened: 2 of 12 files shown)


src/main/java/org/apache/commons/compress/internal/lang3/SystemProperties.java
```diff
@@ -0,0 +1,33 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.commons.compress.internal.lang3;
+
+public final class SystemProperties {
+
+  public static String getOsName() {
+    try {
+      return System.getProperty("os.name");
```
ebourg, line 25: A direct call to System.getProperty("os.name") would be simpler

ebourg: I wouldn't bother dealing with the SecurityManager, the `os.name` property is granted read access in the default policy anyway.

ppkarwasz: Fixed in https://github.com/apache/commons-compress/pull/607/commits/18a4442972842ce8e456d49e0589acbe4bd14aef

ppkarwasz:
> To make things even worse, this PR adds public code (the package name is irrelevant), with ZERO Javadoc and ZERO unit tests.

Sure, there is work to be done on this PR (I marked it as draft). Don't worry, if you are willing to accept this solution, I will make the package really private (in OSGi and JPMS) and add the relevant Javadoc and unit tests.

> This is not the example we should set for contributors and even less for ourselves. This seems to me like a reflection of short term hacking without long term care. I am -1, but it seems pointless as some people are in love with copy-paste over long term maintenance.

Let's ignore `FieldUtils` for now. What kind of maintenance work do you think this code will require in the following years? As I see it:

- The usages of `ArrayFill` in Commons Compress are overkill: all the usages provide a non-null array parameter, so the method can be simplified to:
  ```java
  public int[] fill(int[] a, int val) {
    Arrays.fill(a, val);
    return a;
  }
  ```
  Also note that Commons Lang does **not** provide a helper method for the most common usage of `Arrays.fill`, which is:
  ```java
  public int[] createFilledArray(int size, int val) {
    int[] a = new int[size];
    Arrays.fill(a, val);
    return a;
  }
  ```
- The implementation of `SystemProperties.getOsName()` in Commons Lang delegates to a complex wrapper to `System.getProperty(name)` that handles security exceptions and an empty or `null` parameter `name`. None of these conditions apply to the `os.name` property, which **must** have a valuet (see Javadoc for [System.getProperties()](https://docs.oracle.com/javase/8/docs/api/java/lang/System.html#getProperties--) and should be allowed by `SecurityManager`.

### `FieldUtils`

The code that relies on `FieldUtils` constitutes a maintenance problem itself, since it relies on the **protected** [`FilterInputStream.in`](https://docs.oracle.com/javase/8/docs/api/java/io/FilterInputStream.html#in) field. The field is protected for a reason: users of `FilterInputStream` should not skip access the underlying stream directly, but use the wrapping `FilterInputStream` instead.

Glavo:
I released [Kala Compress](https://github.com/Glavo/kala-compress) 1.27.1-2. It is based on commons-compress, but does not depend on commons-io and commons-lang. Maybe this will be what you want.

ppkarwasz:
> I released [Kala Compress](https://github.com/Glavo/kala-compress) 1.27.1-1. It is based on commons-compress, but does not depend on commons-io and commons-lang. Maybe this will be what you want.

I don't believe that forking Commons Compress is the right approach. Your changes in Kala Compress are aligned with what we plan for Commons Compress 2.x.

I have created a [`2.x` branch](https://github.com/apache/commons-compress/tree/2.x). Feel free to open PRs against that branch to split Commons Compress in separate modules. Probably the first module we need to split is a `commons-compress2-api` module that would include the interfaces to access the various compression algorithms.

acasademont:
We would really appreciate this one, it came as a surprise to use that when we updated from 1.25.0 we had to add a new dependency on commons lang.

src/main/java/org/apache/commons/compress/internal/lang3/reflect/FieldUtils.java
```diff
@@ -0,0 +1,61 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.commons.compress.internal.lang3.reflect;
+
+import static java.util.Objects.requireNonNull;
+
+import java.lang.reflect.Field;
+
+public final class FieldUtils {
```
vlsi, line 25: This is used in a pack200 which has been deprecated for removal in Java 11.
Have you considered moving the implementation to pack200 so the other compressors won't accidentally depend on reflection-like approaches?

vlsi (approved):
The change looks good, and it does solve the true pain for the end-users.

Currently `commons-compress` brings 700KiB `commons-lang3` dependency while it uses only a few methods from it. It does sound like a bloat that the PR addresses.

The point of "the code duplicates lang3" is somewhat valid, however, the only "complicated" code relates to `FieldUtils` which is used in a to-be-deprecated pack200, so I believe it is absolutely fine to move `FieldUtils` right into pack200 so the "duplicated getField" would die with pack200.

garydgregory:
Closing: Won't do.

ebourg:
@garydgregory I still strongly disagree with your decision, there was no consensus for this change.

PR #608, opened 2024-11-18
(doc) Add gzip's ExtraField public apis for accessing SubField(s).

ddeschenes-1:
Improve javadoc, particularly the important immutability part when using the iterator.

### DIFF
--- a/src/main/java/org/apache/commons/compress/compressors/gzip/ExtraField.java
+++ b/src/main/java/org/apache/commons/compress/compressors/gzip/ExtraField.java
@@ -21,6 +21,7 @@ package org.apache.commons.compress.compressors.gzip;
 
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Objects;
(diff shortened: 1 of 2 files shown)


ddeschenes-1:
Hi Gary,
Please refrain to make conflicting changes a 3rd time before merging this 2 files patch.
You know I'm working on fixes you asked and we are the only two contributors active on these 2 files.
Thanks for your understanding.

garydgregory:
@ddeschenes-1 
In the future, please do not reformat to use shorter line lengths or make style changes, it makes the PR noisy and takes longer to review. I'll reformat to [160 line length](https://github.com/apache/commons-compress/blob/f49db00466e6a28f744de3b7f3a5ecf5df11a900/src/conf/checkstyle.xml#L64) after the merge, unless you do it here.

garydgregory:
We keep methods sorted in AB order but I don't know of a checkstyle rule for that. Maybe spotless does that but I have not looked (we don't use spotless here though). We usually don't keep IDE Metadata in git because there are many IDEs and many IDE versions and keeping all that in sync would be painful. Usually people do it the other way around and use the Eclipse Checkstyle plug in.
HTH.

garydgregory:
@ddeschenes-1 
In your experience, are IDs ever duplicated in an extra field? I don't see mention of it in the RFC which is too bad because a Map could be used in place of a List...

ddeschenes-1:
Yeah, that's explicitely why I use a list. It must not be a map.
To be honest, it's clever to be able to repeat an ID, like annotations on java entities, or query parameters in http.
We couldn't live without it.

For gzip, it could be for anything to tag more than once, like signers, certs, hashes...

Yet, I precisely added a findFirst method just for those like me who will be expecting the id only once.

garydgregory:
@ddeschenes-1 
Thank you for the explanation.

garydgregory:
I did some minor internet searching and did not come up with anything
beyond the RFC. Do you see anything?

Gary

ddeschenes-1:
I've looked pretty hard and I've insulted many ChatGPTs for their BS...
There is one Mark Adler (no less) who commented in 2020 on stackoverflow to say there is no more such registry.
https://stackoverflow.com/a/65191556
The BGZF format uses 'BC' (further in that thread).

I need my own subfield id. These will be proprietary ids probably forever unless a new (actually shareable) standards emerge to define some new ones...(<wink>, we'll see).

PR #618, opened 2024-11-29
Throw an IllegalArgumentException when a file name or comment in gzip parameters encodes to a byte array with a 0 byte

ddeschenes-1:
When gzip parameters' filename or comment are encoded into bytes, no verification was made to avoid the \0 that internally terminates those gzip fields.

Thus, the string argument could have a \0, and/or encode into bytes that have a \0 (like with UTF-16 charsets).
This would silently corrupt the gzip member, as a reader would prematurely stop reading the filename or comment and begin reading the next piece, either the comment but inevitably the deflate stream, which would fail.

I thought also about limiting the length on write but the gzip spec has no limit. Limiting length on read (to protect memory) would be imposing an arbitrary limit too, like some gzip tools do. Either way, it would have to be an option like the charset on write, and it would be unusual to ask for such option on the GzipDecompressorInputStream. Using jdk 17 ScopedValues is not possible yet and it would be a very forgettable option to set anyway. Without some discussion, there is no point in placing another api facing change on constructors.

### DIFF
--- a/src/main/java/org/apache/commons/compress/compressors/gzip/GzipParameters.java
+++ b/src/main/java/org/apache/commons/compress/compressors/gzip/GzipParameters.java
@@ -26,7 +26,7 @@ import java.util.zip.Deflater;
 
 import org.apache.commons.io.Charsets;
 import org.apache.commons.lang3.ArrayUtils;
-
+import org.apache.commons.lang3.StringUtils;
 
 /**
  * Parameters for the GZIP compressor.
@@ -300,11 +300,8 @@ public class GzipParameters {
     private int deflateStrategy = Deflater.DEFAULT_STRATEGY;
 
     private String requireNonNulByte(final String text) {
-        if (text != null && text.length() > 0) {
-            if (ArrayUtils.contains(text.getBytes(fileNameCharset), (byte) 0)) {
-                throw new IllegalArgumentException(
-                        "String encoded in Charset '" + fileNameCharset + "' contains the nul byte 0 which is not supported in gzip.");
-            }
+        if (StringUtils.isNotEmpty(text) && ArrayUtils.contains(text.getBytes(fileNameCharset), (byte) 0)) {
+            throw new IllegalArgumentException("String encoded in Charset '" + fileNameCharset + "' contains the nul byte 0 which is not supported in gzip.");
         }
         return text;
     }
--- a/src/main/java/org/apache/commons/compress/compressors/gzip/GzipParameters.java
+++ b/src/main/java/org/apache/commons/compress/compressors/gzip/GzipParameters.java
@@ -25,6 +25,7 @@ import java.time.Instant;
 import java.util.zip.Deflater;
 
 import org.apache.commons.io.Charsets;
+import org.apache.commons.lang3.ArrayUtils;
 
 
 /**
@@ -298,6 +299,16 @@ public class GzipParameters {
     private int bufferSize = 512;
     private int deflateStrategy = Deflater.DEFAULT_STRATEGY;
 
+    private String requireNonNulByte(final String text) {
+        if (text != null && text.length() > 0) {
+            if (ArrayUtils.contains(text.getBytes(fileNameCharset), (byte) 0)) {
+                throw new IllegalArgumentException(
+                        "String encoded in Charset '" + fileNameCharset + "' contains the nul byte 0 which is not supported in gzip.");
+            }
+        }
+        return text;
+    }
+
     /**
      * Gets size of the buffer used to retrieve compressed data.
      *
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/commons/compress/compressors/gzip/GzipParameters.java
```diff
@@ -298,6 +298,18 @@ public int type() {
     private int bufferSize = 512;
     private int deflateStrategy = Deflater.DEFAULT_STRATEGY;
 
+    private void checkEncodedStringHasNoZero(String text) {
+        if (text != null && text.length() > 0) {
```
garydgregory, line 302: Just call `org.apache.commons.lang3.ArrayUtils.contains(byte[], byte)`.

src/test/java/org/apache/commons/compress/compressors/gzip/GzipParametersTest.java
```diff
@@ -46,4 +50,33 @@ public void testToString() {
...
+        "          , hello\0world, false",
+        "ISO-8859-1, helloworld, true",
+        "ISO-8859-1, hello�world, true",
+        "ISO-8859-1, hello\0world, false",
+        "UTF-8     , helloworld, true",
+        "UTF-8     , hello�world, true",
+        "UTF-8     , hello\0world, false",
+        "UTF-16BE  , helloworld, false",
+    })//@formatter:on
+    public void testIllegalCommentOrFileName(String charset, String text, boolean shouldPass) {
```
garydgregory, line 67: Use `final` where you can.

garydgregory (changes requested):
Hello @ddeschenes-1 

Thank you for your PR!

- Please see my 6 comments.
- Make you to run `mvn` by itself to run all build checks before you push.

src/test/java/org/apache/commons/compress/compressors/gzip/GzipParametersTest.java
ddeschenes-1: This is a polluting practice that should never be enforced.

There are IDE warnings to catch assignment to method parameters, for those who care; even PMD can be set to warn on AvoidReassigningParameters. There is no justification for polluting the code with final, or generating a warning from a missing final if it is not even assigned anywhere.

src/test/java/org/apache/commons/compress/compressors/gzip/GzipParametersTest.java
```diff
@@ -46,4 +50,33 @@ public void testToString() {
         gzipParameters.setOS(GzipParameters.OS.Z_SYSTEM);
         assertTrue(gzipParameters.toString().contains("Z_SYSTEM"));
     }
+
+    @ParameterizedTest
+    @CsvSource({ //@formatter:off
```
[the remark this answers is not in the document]
ddeschenes-1: Ok, but note that if the fixture was any bigger, a split would waste the parameterized test benefit (code reuse and single topic).

src/test/java/org/apache/commons/compress/compressors/gzip/GzipParametersTest.java
```diff
@@ -46,4 +50,44 @@ public void testToString() {
...
+        "ISO-8859-1, helloworld",
+        "ISO-8859-1, helloéworld",
+        "UTF-8     , helloworld",
+        "UTF-8     , helloéworld"
+    })
+    //@formatter:on
+    public void testLegalCommentOrFileName(final String charset, final String text) {
+        final GzipParameters p = new GzipParameters();
+        if (charset != null) {
+            p.setFileNameCharset(Charset.forName(charset));
```
garydgregory, line 68: @ddeschenes-1 
In the future, make the parameter a `Charset` instead of a String and you won't need the test to do the conversion.

garydgregory:
@ddeschenes-1 

Merged 🚀 TY!

src/test/java/org/apache/commons/compress/compressors/gzip/GzipParametersTest.java
ddeschenes-1: There is no automatic converter from String to Charset in junit5.
Adding the necessary lines for a @ConvertWith converter class would be far more code than those 2 lines.

garydgregory: Except when you're wrong: https://github.com/junit-team/junit5/blob/main/junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/StringToCommonJavaTypesConverter.java#L38

ddeschenes-1: Great. thank you internet for misleading me again. what is it gonna be in 10 more years...

btw, when is 1.28.0 releasing?

garydgregory: I'd first like to finish shepherding the current other Apache Commons release candidates to completion. Could be another week or two before this components gets a release candidate.

PR #624, opened 2024-12-14
(doc) fix gzip compressor incorrect MTIME value which is already in seconds

ddeschenes-1:
Fix the gzip compressor inputstream and outputstream incorrect MTIME value which is already in seconds.
They should not divide or multiply by 1000.


garydgregory:
Hello @ddeschenes-1 

Thank you for your PR.

- The PR title is incorrect: This is NOT a documentation change
- The PR description is empty.

Otherwise, looks good!

ddeschenes-1:
I thought the doc said that without a jira, the (doc) must be added in commit first line.
So, that's what I do.

garydgregory:
Hi @ddeschenes-1 

Are there user-visible consequences to this fix?

ddeschenes-1:
The saved information in the gzip file is wrong. So, yes?

ddeschenes-1:
so, 1.28.0 this weekend? ;-)

PR #627, opened 2024-12-21
Optionally add FHCRC to GZIP header.

ddeschenes-1:
Given the boolean to generate the header CRC in the gzip parameters, the GzipCompressorOutputStream will
compute and add the CRC16 (actually low 16 bits of the CRC32) of the headers, as per the RFC.

The GzipCompressorInputStream is still handling the presence of HCRC but still not computing nor checking it,
because this would be too much a risk of regression. It will however raise the HCRC boolean in the loaded gzip parameters
to signal it was seen.

### DIFF
--- a/src/main/java/org/apache/commons/compress/compressors/gzip/GzipCompressorOutputStream.java
+++ b/src/main/java/org/apache/commons/compress/compressors/gzip/GzipCompressorOutputStream.java
@@ -47,6 +47,9 @@ public class GzipCompressorOutputStream extends CompressorOutputStream<OutputStr
     /** Header flag indicating a comment follows the header */
     private static final int FCOMMENT = 1 << 4;
 
+    /** Header flag indicating a header CRC follows the header */
+    private static final int FHCRC = 1 << 1;
+
     /** Deflater used to compress the data */
     private final Deflater deflater;
 
@@ -161,8 +164,11 @@ public class GzipCompressorOutputStream extends CompressorOutputStream<OutputStr
      */
     private void writeC(final String value, final Charset charset) throws IOException {
         if (value != null) {
-            out.write(value.getBytes(charset));
+            byte[] ba = value.getBytes(charset);
+            out.write(ba);
             out.write(0);
+            crc.update(ba);
+            crc.update(0);
         }
     }
 
@@ -174,7 +180,11 @@ public class GzipCompressorOutputStream extends CompressorOutputStream<OutputStr
         buffer.order(ByteOrder.LITTLE_ENDIAN);
         buffer.putShort((short) GZIPInputStream.GZIP_MAGIC);
         buffer.put((byte) Deflater.DEFLATED); // compression method (8: deflate)
-        buffer.put((byte) ((extra != null ? FEXTRA : 0) | (fileName != null ? FNAME : 0) | (comment != null ? FCOMMENT : 0))); // flags
+        buffer.put((byte) ((extra != null ? FEXTRA : 0)
+                | (fileName != null ? FNAME : 0)
+                | (comment != null ? FCOMMENT : 0)
+                | (parameters.hasHeaderCRC() ? FHCRC : 0)
+        )); // flags
         buffer.putInt((int) parameters.getModificationInstant().getEpochSecond());
         // extra flags
         final int compressionLevel = parameters.getCompressionLevel();
@@ -187,13 +197,23 @@ public class GzipCompressorOutputStream extends CompressorOutputStream<OutputStr
         }
         buffer.put((byte) parameters.getOperatingSystem());
         out.write(buffer.array());
+        crc.update(buffer.array());
         if (extra != null) {
             out.write(extra.length & 0xff); // little endian
             out.write(extra.length >>> 8 & 0xff);
             out.write(extra);
+            crc.update(extra.length & 0xff);
+            crc.update(extra.length >>> 8 & 0xff);
+            crc.update(extra);
         }
         writeC(fileName, parameters.getFileNameCharset());
         writeC(comment, parameters.getFileNameCharset());
+        if (parameters.hasHeaderCRC()) {
+            int v = (int) crc.getValue() & 0xffff;
+            out.write(v & 0xff);
+            out.write((v >>> 8) & 0xff);
+        }
+        crc.reset();
     }
 
     private void writeTrailer() throws IOException {
--- a/src/main/java/org/apache/commons/compress/compressors/gzip/GzipCompressorOutputStream.java
+++ b/src/main/java/org/apache/commons/compress/compressors/gzip/GzipCompressorOutputStream.java
@@ -164,7 +164,7 @@ public class GzipCompressorOutputStream extends CompressorOutputStream<OutputStr
      */
     private void writeC(final String value, final Charset charset) throws IOException {
         if (value != null) {
-            byte[] ba = value.getBytes(charset);
+            final byte[] ba = value.getBytes(charset);
             out.write(ba);
             out.write(0);
             crc.update(ba);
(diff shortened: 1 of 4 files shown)


src/main/java/org/apache/commons/compress/compressors/gzip/GzipCompressorOutputStream.java
```diff
@@ -47,6 +47,9 @@ public class GzipCompressorOutputStream extends CompressorOutputStream<OutputStr
     /** Header flag indicating a comment follows the header */
     private static final int FCOMMENT = 1 << 4;
 
+    /** Header flag indicating a header crc follows the header */
```
garydgregory, line 50: Javadoc "crc" -> "CRC" because CRC is an acronym.

src/main/java/org/apache/commons/compress/compressors/gzip/GzipCompressorOutputStream.java
```diff
@@ -185,14 +196,26 @@ private void writeHeader(final GzipParameters parameters) throws IOException {
             buffer.put((byte) 0);
         }
         buffer.put((byte) parameters.getOperatingSystem());
+
```
garydgregory, line 199: Remove extra whitespace, if you need to point something out, use an `// inline comment`.

src/main/java/org/apache/commons/compress/compressors/gzip/GzipCompressorOutputStream.java
```diff
@@ -185,14 +196,26 @@ private void writeHeader(final GzipParameters parameters) throws IOException {
...
             out.write(extra.length & 0xff); // little endian
             out.write(extra.length >>> 8 & 0xff);
             out.write(extra);
+            crc.update(extra.length & 0xff);
+            crc.update(extra.length >>> 8 & 0xff);
+            crc.update(extra);
         }
         writeC(fileName, parameters.getFileNameCharset());
         writeC(comment, parameters.getFileNameCharset());
+
```
garydgregory, line 212: Remove extra whitespace, if you need to point something out, use an `// inline comment`.

src/main/java/org/apache/commons/compress/compressors/gzip/GzipParameters.java
```diff
@@ -432,6 +433,15 @@ public OS getOS() {
         return operatingSystem;
     }
 
+    /**
+     * Returns if the header crc is to be added (when writing) or was present (when reading).
+     *
+     * @return true is header crc si to be added (on write) or was found (after read).
+     */
```
garydgregory, line 440: - This is not a good public API name IMO. "Hcrc" -> "HeaderCrc()". 
- New public and protected elements need Javadoc `@since` tags.

src/main/java/org/apache/commons/compress/compressors/gzip/GzipParameters.java
```diff
@@ -539,6 +550,15 @@ public void setFileNameCharset(final Charset charset) {
         this.fileNameCharset = Charsets.toCharset(charset, GzipUtils.GZIP_ENCODING);
     }
 
+    /**
+     * Establishes the presence of the header flag FLG.FHCRC and its headers crc16 value.
+     *
+     * @param hcrc when true, the header crc16 is calculated and inserted in the gzip header on write; on read it means it was present.
+     */
+    public void setHcrc(boolean hcrc) {
```
garydgregory, line 558: - This is not a good public API name IMO. "Hcrc" -> "HeaderCrc()". 
- New public and protected elements need Javadoc `@since` tags.

src/test/java/org/apache/commons/compress/compressors/gzip/GzipCompressorOutputStreamTest.java
```diff
@@ -214,4 +220,57 @@ public void testFileNameChinesePercentEncoded() throws IOException {
...
+                + "0300" // empty deflate stream
+                + "00000000" // crs32
+                + "00000000" // isize
+        );
+        assertArrayEquals(expected, result);
+        try (GZIPInputStream gis = new GZIPInputStream(new ByteArrayInputStream(result))) {
+            //if it does not fail, the hcrc is good.
+        }
+        try (GzipCompressorInputStream gis = new GzipCompressorInputStream(new ByteArrayInputStream(result))) {
+            GzipParameters md = gis.getMetaData();
```
garydgregory, line 263: "md" is not a good name for maintenance IMO -> "metaData"

src/test/java/org/apache/commons/compress/compressors/gzip/GzipCompressorOutputStreamTest.java
```diff
@@ -214,4 +220,57 @@ public void testFileNameChinesePercentEncoded() throws IOException {
...
+        parameters.setModificationTime(0x66554433); // avoid changing time
+        parameters.setFileName("AAAA");
+        parameters.setComment("ZZZZ");
+        parameters.setOS(OS.UNIX);
+        final ExtraField extra = new ExtraField();
+        extra.addSubField("BB", "CCCC".getBytes(StandardCharsets.ISO_8859_1));
+        parameters.setExtraField(extra);
+        ByteArrayOutputStream baos = new ByteArrayOutputStream();
+        try (GzipCompressorOutputStream gos = new GzipCompressorOutputStream(baos, parameters)) {
+            // no thing to write for this test.
```
garydgregory, line 242: "no thing" -> "nothing"

src/test/java/org/apache/commons/compress/compressors/gzip/GzipCompressorOutputStreamTest.java
```diff
@@ -214,4 +220,57 @@ public void testFileNameChinesePercentEncoded() throws IOException {
...
+        assertArrayEquals(expected, result);
+        try (GZIPInputStream gis = new GZIPInputStream(new ByteArrayInputStream(result))) {
+            //if it does not fail, the hcrc is good.
+        }
+        try (GzipCompressorInputStream gis = new GzipCompressorInputStream(new ByteArrayInputStream(result))) {
+            GzipParameters md = gis.getMetaData();
+            assertTrue(md.hasHcrc());
+            assertEquals(0x66554433, md.getModificationTime());
+            assertEquals(1, md.getExtraField().size());
+            SubField sf = md.getExtraField().iterator().next();
```
garydgregory, line 267: Use `final` where you can.

garydgregory (changes requested):
Hello @ddeschenes-1 

- Thank you for the PR.
- Rebase on git master to pick up the updated checkstyle configuration which will now notify you of missing `final` keywords and `@Override`
- Git master now also adds all missing `@Override` annotations (a small conflict with this PR)
- Please update the PR with my scattered comments.

TY!

src/test/java/org/apache/commons/compress/compressors/gzip/GzipCompressorOutputStreamTest.java
```diff
@@ -214,4 +220,57 @@ public void testFileNameChinesePercentEncoded() throws IOException {
...
+            assertEquals("BB", sf.getId());
+            assertEquals("CCCC", new String(sf.getPayload(), StandardCharsets.ISO_8859_1));
+            assertEquals("AAAA", md.getFileName());
+            assertEquals("ZZZZ", md.getComment());
+            assertEquals(OS.UNIX, md.getOS());
```
garydgregory, line 272: Since the stream now computes the header CRC, add an assertion that it is correct.

src/main/java/org/apache/commons/compress/compressors/gzip/GzipCompressorOutputStream.java
ddeschenes-1: nitpicking.

src/test/java/org/apache/commons/compress/compressors/gzip/GzipCompressorOutputStreamTest.java
ddeschenes-1: We cannot access the CRC which is not saved.
The value is asserted precisely with the GZIPInputStream constructor.
If the value was wrong, this constructor would fail.

garydgregory (approved):
TY.

PR #629, opened 2024-12-30
Updated javadoc to clarify how GzipParameters behave on multiple gzip members

ddeschenes-1:
Updated javadoc to clarify how GzipParameters behave on multiple gzip members.

### DIFF
--- a/src/main/java/org/apache/commons/compress/compressors/gzip/GzipCompressorInputStream.java
+++ b/src/main/java/org/apache/commons/compress/compressors/gzip/GzipCompressorInputStream.java
@@ -135,6 +135,10 @@ public class GzipCompressorInputStream extends CompressorInputStream implements 
      * If {@code decompressConcatenated} is {@code false}: This decompressor might read more input than it will actually use. If {@code inputStream} supports
      * {@code mark} and {@code reset}, then the input position will be adjusted so that it is right after the last byte of the compressed stream. If
      * {@code mark} isn't supported, the input position will be undefined.
+     * <p>
+     * If {@code decompressConcatenated} is {@code true}: decompress until the end of the input. The getMetadata()'s GzipParameters instance
+     * is updated but not cleared by each successive gzip member encountered. Details presented like fileName, comment, header CRC and extra fields
+     * may linger if not present on subsequent members. Extra fields replaced as a whole when seen (no union of subfields).
      *
      * @param inputStream            the InputStream from which this object should be created of
      * @param decompressConcatenated if true, decompress until the end of the input; if false, stop after the first .gz member
--- a/src/main/java/org/apache/commons/compress/compressors/gzip/GzipCompressorInputStream.java
+++ b/src/main/java/org/apache/commons/compress/compressors/gzip/GzipCompressorInputStream.java
@@ -137,8 +137,9 @@ public class GzipCompressorInputStream extends CompressorInputStream implements 
      * {@code mark} isn't supported, the input position will be undefined.
      * <p>
      * If {@code decompressConcatenated} is {@code true}: decompress until the end of the input. The getMetadata()'s GzipParameters instance
-     * is updated but not cleared by each successive gzip member encountered. Details presented like fileName, comment, header CRC and extra fields
-     * may linger if not present on subsequent members. Extra fields replaced as a whole when seen (no union of subfields).
+     * is updated thus new gzip paramaters are integrated by each successive gzip member encountered.
+     * Details previously presented like file name, comment, header CRC and extra fields will remain set even when not presented on subsequent members.
+     * Extra fields are replaced as a whole when observed (no union of subfields).
      *
      * @param inputStream            the InputStream from which this object should be created of
      * @param decompressConcatenated if true, decompress until the end of the input; if false, stop after the first .gz member


ddeschenes-1:
I'm somewhat ok with the current behavior of lingering filename/comment/extras/hcrc, but I's like a list of GzipParameters (GP), except that I know better and it would leak memory in some cases.

If we had a limited list, it would be surprising to users to loose any of them. So I think I would opt for a callback like a "onNewGzipMember(GzipParameters parameters)".

This callback doesn't have to be on constructor since the result of the constructor is necessarily to produce the first GP.
So, we could have a .withMemberListener(Consumer<GzipParameters> listener) -> GzipCompressorInputStream,
which preserves the api and avoids overloading again.

The callback would be called synchronously in init() before returning (and no try catch).

What do you think?

garydgregory:
Hello @ddeschenes-1 

Thank you for your PR. In general, I like the idea of documenting the existing behavior, so I am OK with the current PR, BUT:

The "lingering" GzipParameters data in an input stream sounds like a bug.

I would rather have this metadata be sound and predictable before we invent a new setting to detect when a new `GzipParameters` is created/initialized.

WRT to adding new settings, I want to stop inflating the constructor space, so I am using builders from now on, for example, see `org.apache.commons.compress.archivers.sevenz.SevenZFile.Builder`.

I added support for `equals()` and `hashCode()` to `GzipParameters` to facilitate any checks.

How could the input stream be changed to avoid the "lingering" issue?

ddeschenes-1:
It started like a bug in my mind, but it evolved... (it's nto a bug; it's a feature? lol)
We'd have to choose between keeping the 1st GP and ignoring the subsequent ones, or keeping the last GP and clobbering the previous one. That's where I got torn. If a filename or comment is set, it is likely meaningful in the 1st member. If extras are added, it could be valuable anywhere. "mtime" is meant to indicate modification time, so taking it from the last member is likely to be more meaningful. But again, it's the mtime of the member really, not the whole file. See, there is no good answer so I'd pass the puck to the user with a listener.

As for what the GCIS should do regardless of a listener, I think the lingering fields are not so bad because it provides a **default integrator** of the whole stream. Maybe that's the phrasing we should use to put a positive spin on this behavior.

garydgregory:
One not very clean possibility, short of adding a new method like `onGzipParameters()` would be to make `init()` protected instead of private. Then a call site could create an anonymous subclass and override `init()`. It's not great but possible.

A little cleaner, would be to add a method called `onGzipParameters()` which is called from `init()` at its end, and same as above, a call site could create an anonymous subclass and override `onGzipParameters()`; this avoids the stream keeping an extra instance variable.

Whether or not the above is "better" than adding an instance variable depends on how often this use case happens and how much we should change the class to accommodate for it in the design or simply allow for it do just be doable with a subclass.

Back to bugs: It's hard to see how this is not a bug:
```java
        switch (inData.readUnsignedByte()) { // extra flags
        case 2:
            parameters.setCompressionLevel(Deflater.BEST_COMPRESSION);
            break;
        case 4:
            parameters.setCompressionLevel(Deflater.BEST_SPEED);
            break;
        default:
            // ignored for now
            break;
        }
```
The above means that for a new member, the GP carries the previous member's value, right? How is that not a bug? It seems to me the code should be:
```java
        switch (inData.readUnsignedByte()) { // extra flags
        case 2:
            parameters.setCompressionLevel(Deflater.BEST_COMPRESSION);
            break;
        case 4:
            parameters.setCompressionLevel(Deflater.BEST_SPEED);
            break;
        default:
            parameters.setCompressionLevel(Deflater.DEFAULT_COMPRESSION);
            break;
        }
```
The above change does not break the build (running `mvn` by itself) but we may not have the right kind of test for this use case.

Checking code coverage with `mvn clean install site -P jacoco` shows that the new line is executed but the one for `parameters.setCompressionLevel(Deflater.BEST_SPEED);` isn't, so the coverage is not worse.

WDYT?

ddeschenes-1:
Good catch on the compression level. One more field on the pile...
At this point I think we need to create a new GP and (atomically) assign it when ready.

If a user is interested in the first GP, he does so after the constructor.
Any other time, he calls getMetadata() and gets whatever last seen GP is there.
For now, that's good enough to stabilize the behavior in a non surprising way. 

+1 on adding a protected method called **onGzipMember()**  (no-arg)
The call would be made **after** setting the GP of course so the subclass at can read it.
This covers the initial member on construction too.

I'm not too keen on coding it and wasting time on build failures and spaces... So go ahead!
We can drop this PR, you'd likely write the proper javadoc anyway. Thanks.

ddeschenes-1:
And happy new year!

ddeschenes-1:
Actually, I changed my mind. Remembering that composition is preferable over inheritance.
Asking for a simple listener as a lambda surely, is lighter on the user.

Client perpspective:
`gzis.withMemberListener(gp -> .....));`

Implementor perspective:
`Optional<Consumer<GzipParameters>> memberListener;`
...
`memberListener.ifPresent(consumer -> consumer.accept(gp));`

garydgregory:
I have a solution done locally but there are a couple of wrinkles I am working out.

garydgregory:
Hello @ddeschenes-1 ,

Please see git master's `GzipCompressorInputStreamTest.testOnMemberSaveAsFiles()` and sibling methods.

PR #649, opened 2025-03-20
[COMPRESS-695] Ability to use different InputStreams for Zstd

mehmet-karaman:
Exposed the instantiation of Zstd InputStream. This should give the ability to override and use a different Zstd compressor library.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -108,6 +108,12 @@ Brotli, Zstandard and ar, cpio, jar, tar, zip, dump, 7z, arj.
       <artifactId>junit-vintage-engine</artifactId>
       <scope>test</scope>
     </dependency>
+	<dependency>
+	    <groupId>io.airlift</groupId>
+	    <artifactId>aircompressor</artifactId>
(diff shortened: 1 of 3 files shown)


garydgregory:
Hello @mehmet-karaman 

New public and protected methods need a Javadoc `since` tag.

mehmet-karaman:
For the new test: I have to introduce new dependency to aircompressor (Will it introduce new IP problems? Is it easy to introduce new dependencies in apache projects?) . 

For the other things: I am going to adjust my commit according your comment.

garydgregory:
@mehmet-karaman 

aircompressor is Apache licensed so it shouldn't be an issue.

mehmet-karaman:
Ok.. Then I will add also a test for this.. Thanks..

mehmet-karaman:
test is added, `mvn` was successfull. Added the new dependency in test scope, so it shouldn't harm anything else.

mehmet-karaman:
Thank you :).

PR #656, opened 2025-03-24
[COMPRESS-695] Ability to use different InputStreams for Zstd

mehmet-karaman:
* Change ZstdCompressorInputStream::new to lambda expression, because it accesses the ZstdCompressorInputStream directly and tries to access ZstdInputStream. Now it accesses only if the apply method was called.

### DIFF
--- a/src/main/java/org/apache/commons/compress/archivers/zip/ZipFile.java
+++ b/src/main/java/org/apache/commons/compress/archivers/zip/ZipFile.java
@@ -93,7 +93,9 @@ import org.apache.commons.io.input.BoundedInputStream;
  */
 public class ZipFile implements Closeable {
 
-    private static final IOFunction<InputStream, InputStream> DEFAULT_ZSTD_INPUT_STREAM_FACTORY = ZstdCompressorInputStream::new;
+    private static final IOFunction<InputStream, InputStream> DEFAULT_ZSTD_INPUT_STREAM_FACTORY = i -> {
+        return new ZstdCompressorInputStream(i);
+    };
 
     /**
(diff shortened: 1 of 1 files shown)


src/main/java/org/apache/commons/compress/archivers/zip/ZipFile.java
```diff
@@ -186,7 +188,9 @@ public ZipFile get() throws IOException {
          * @since 1.28.0
          */
         public Builder setZstdInputStreamFactory(final IOFunction<InputStream, InputStream> zstdInpStreamFactory) {
-            this.zstdInputStream = zstdInpStreamFactory == null ? DEFAULT_ZSTD_INPUT_STREAM_FACTORY : zstdInpStreamFactory;
+            this.zstdInputStream = zstdInpStreamFactory == null ? i -> {
```
mehmet-karaman, line 191: Going to adjust this back.. Might be not necessarey anymore.. I am going to test this with the zip editor.

mehmet-karaman: done..

mehmet-karaman:
@garydgregory Do you need a screenshot from the bytecode diff from the function reference and the lambda expression or is it ok?

garydgregory:
I will look at this later, I have a couple of higher priorities ATM.

garydgregory:
Hello @mehmet-karaman 

I did this differently, please check git master for your use case.

mehmet-karaman:
Tested this with the zip editor client it worked. Thanks you @garydgregory

PR #663, opened 2025-04-12
[COMPRESS-697] Move BitStream.nextBit() method to BitInputStream

fkjellberg:
I'm using BitInputStream for Huffman decoding and having a method for reading a single bit would simplify the code somewhat. There is a method nextBit used for Huffman decoding already in the subclass BitStream used by the Zip implementation. This class is not available in the public API of commons-compress so I cannot use it. I think it makes sense to move the nextBit method to BitInputStream and make it available in the public API.


garydgregory:
Hello @fkjellberg 

Please rebase on git master to pick up a build fix which should make your Java 8 builds green.

fkjellberg:
Hi @garydgregory 

I rebased the PR

PR #664, opened 2025-04-16
Allow setting preset for LZMACompressorOutputStream

dwalluck:


garydgregory (changes requested):
Hi @dwalluck 

Thank for your PR.

Let's not go start going down this road or we'll end up with asks for variations of:
```
    public LZMA2Options(int dictSize, int lc, int lp, int pb, int mode,
                        int niceLen, int mf, int depthLimit)
```
Instead, please try to use this pattern in git master:
```
        try (LZMACompressorOutputStream out = LZMACompressorOutputStream.builder()
                .setPath(outPath)
                .setLzma2Options(options)
                .get()) {
            out.writeUtf8(data);
        }
```
I can see keeping the "normal" one-arg constructor around and using the builder for anything that requires a non-default configuration.

dwalluck:
@garydgregory I understand. I only thought to add it for parity with XZ which already has this constructor.

dwalluck:
Wait, maybe I don't understand. You want to create a Builder just for the options? But, if we add a second arg of type `LZMA2Options`, that also would take care of that, no?

ppkarwasz:
There is already a `Builder.setLzma2Options(options)` method, so  there is no need to add a new constructor.

dwalluck:
Ah, it was just added that same day? That was fast! Thanks!

However, if there were a builder for the options themselves, then the same pattern could be shared with xz, too. Otherwise, it seems xz needs its own builder, too, right?

Similarly, I think zstd needs a way to get at all of the options that are available via the `ZstdOutputStream::set` methods. In particular, I wanted to get at `ZstdOutputStream::setWorkers` to enable parallel compression. Then the `ZstdCompressorOutputStream(final OutputStream outStream, final int level, ...)` constructors could also be deprecated the way you did with LZMA.

ppkarwasz:
> Ah, it was just added that same day? That was fast! Thanks!

Right, I didn't notice, this was pushed directly to `master` in 55ecb7f1b8f58408c525fef742dcf27d51efebf3.
@garydgregory, could we use PRs for this kind of changes? A direct push to `master` kind of cuts short discussions on PRs likes this one. If you create a new PR, the discussion can continue in the new PR.

garydgregory:
> Ah, it was just added that same day? That was fast! Thanks!
> 
> However, if there were a builder for the options themselves, then the same pattern could be shared with xz, too. Otherwise, it seems xz needs its own builder, too, right?
> 
> Similarly, I think zstd needs a way to get at all of the options that are available via the `ZstdOutputStream::set` methods. In particular, I wanted to get at `ZstdOutputStream::setWorkers` to enable parallel compression. Then the `ZstdCompressorOutputStream(final OutputStream outStream, final int level, ...)` constructors could also be deprecated the way you did with LZMA.

Hello @dwalluck 

Please see this PR https://github.com/apache/commons-compress/pull/666 to add a builder to `ZstdCompressorOutputStream`.

TY,
Gary

garydgregory:
Closing: The feature is in git master through a builder.

PR #681, opened 2025-08-01
[COMPRESS-702] Performance issue in SevenZFile

Lastoneee:

In version 1.27.1 of Apache Commons Compress, the `buildDecodingStream()` method did not pass the `countThreshold ` parameter into the `ChecksumInputStream` constructor, resulting in the `getRemaining()` method of `ChecksumInputStream` returning an incorrect number of remaining bytes. This will affect `SevenZfile` # `getInputStream()` 's judgment of `hasCurrentIndexBeeReyad()` and `skipEnteriesWhenNeed()` in random access mode, resulting in the system being unable to determine whether the current entry has been read, and re extracting the entire folder content from the file header every time, seriously affecting performance. By displaying the call to. `setCountThreshold (file. getSize())` during stream construction, it is possible to ensure accurate calculation of remaining bytes, thereby correctly skipping processed entries and significantly improving random access performance.


CPPPAN:
This bug fix has genuinely helped us resolve issues we encountered during development—thumbs up!

PR #685, opened 2025-08-10
Simplify `ArArchiveInputStream` special record handling

ppkarwasz:
Refactors `ArArchiveInputStream.getNextEntry()` to follow the same structural approach as TarArchiveInputStream.getNextEntry()`, improving consistency and maintainability.

**Key changes:**

* Unifies parsing of the GNU string table with the pattern used for TAR special records.
* Replaces the _ad-hoc_ byte counter with the inherited `count()` method for more accurate and readable tracking.
* Adds a sanity check: throws an exception if a regular file entry does not follow a special AR member (GNU string table).v
* Consistently throws `EOFException` in case of truncated content.
* Improves consistency of error messages.

- [x] I used AI to proofread the pull request.

### DIFF
--- a/src/main/java/org/apache/commons/compress/archivers/ar/ArArchiveInputStream.java
+++ b/src/main/java/org/apache/commons/compress/archivers/ar/ArArchiveInputStream.java
@@ -59,11 +59,6 @@ public class ArArchiveInputStream extends ArchiveInputStream<ArArchiveEntry> {
     private static final Pattern GNU_LONGNAME_PATTERN = Pattern.compile("^/\\d+");
 
     /**
-     * Signature {@code 3c21 7261 6863 0a3e}
-     */
-    private static final byte[] SIGNATURE = { 0x21, 0x3c, 0x61, 0x72, 0x63, 0x68, 0x3e, 0x0a };
-
-    /**
      * Does the name look like it is a long name (or a name containing spaces) as encoded by BSD ar?
      * <p>
      * From the FreeBSD ar(5) man page:
@@ -103,8 +98,8 @@ public class ArArchiveInputStream extends ArchiveInputStream<ArArchiveEntry> {
      * are number of characters, not line or string number within the "//" file.
      * </p>
      */
-    private static boolean isGNUStringTable(final String name) {
-        return GNU_STRING_TABLE_NAME.equals(name);
+    private static boolean isGNUStringTable(final ArArchiveEntry entry) {
+        return GNU_STRING_TABLE_NAME.equals(entry.getName());
     }
 
     /**
@@ -115,10 +110,8 @@ public class ArArchiveInputStream extends ArchiveInputStream<ArArchiveEntry> {
      * @return true, if this stream is an Ar archive stream, false otherwise.
      */
     public static boolean matches(final byte[] buffer, final int ignored) {
-        return ArrayUtils.startsWith(buffer, SIGNATURE);
+        return ArrayUtils.startsWith(buffer, ArArchiveEntry.HEADER_BYTES);
     }
-
-    private long offset;
 
     private boolean closed;
 
@@ -131,7 +124,7 @@ public class ArArchiveInputStream extends ArchiveInputStream<ArArchiveEntry> {
     private byte[] namebuffer;
 
     /**
-     * The offset where the current entry started. -1 if no entry has been called
+     * The offset where the data for the current entry starts.
      */
     private long entryOffset = -1;
 
@@ -147,24 +140,22 @@ public class ArArchiveInputStream extends ArchiveInputStream<ArArchiveEntry> {
         super(inputStream, StandardCharsets.US_ASCII.name());
     }
 
-    private int asInt(final byte[] byteArray, final int offset, final int len) throws IOException {
-        return asInt(byteArray, offset, len, 10, false);
+    private int asOptionalDecimal(final byte[] byteArray, final int offset, final int len) throws IOException {
+        return asInt(byteArray, offset, len, 10);
     }
 
-    private int asInt(final byte[] byteArray, final int offset, final int len, final boolean treatBlankAsZero) throws IOException {
-        return asInt(byteArray, offset, len, 10, treatBlankAsZero);
+    private int asOptionalOctal(final byte[] byteArray, final int offset, final int len) throws IOException {
+        return asInt(byteArray, offset, len, 8);
     }
 
     private int asInt(final byte[] byteArray, final int offset, final int len, final int base) throws IOException {
-        return asInt(byteArray, offset, len, base, false);
+        final String string = ArchiveUtils.toAsciiString(byteArray, offset, len).trim();
+        return string.isEmpty() ? 0 : ParsingUtils.parseIntValue(string, base);
     }
 
-    private int asInt(final byte[] byteArray, final int offset, final int len, final int base, final boolean treatBlankAsZero) throws IOException {
+    private long asOptionalLong(final byte[] byteArray, final int offset, final int len) throws IOException {
         final String string = ArchiveUtils.toAsciiString(byteArray, offset, len).trim();
-        if (string.isEmpty() && treatBlankAsZero) {
-            return 0;
-        }
-        return ParsingUtils.parseIntValue(string, base);
+        return string.isEmpty() ? 0 : ParsingUtils.parseLongValue(string);
     }
 
     private long asLong(final byte[] byteArray, final int offset, final int len) throws IOException {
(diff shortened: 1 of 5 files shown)


src/main/java/org/apache/commons/compress/archivers/ar/ArArchiveInputStream.java
```diff
@@ -241,102 +232,172 @@ private String getExtendedName(final int offset) throws IOException {
...
-                // hit eof
-                return null;
+
+            // Read the next header record
+            final byte[] headerBuf = getRecord();
+            if (headerBuf == null) {
+                // If we encountered special records but no file entry, the archive is malformed
+                if (foundGNUStringTable) {
+                    throw new EOFException("Premature end of ar archive: no regular entry after GNU string table.");
+                }
```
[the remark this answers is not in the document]
ppkarwasz: I’d keep the exception as it is unless we can point to a real-world tool that writes a trailing GNU string table. The FreeBSD `ar(5)` manual explicitly states:

> If present, this member immediately follows the archive symbol table if an archive symbol table is present, or is the first member otherwise.

Source: [FreeBSD man ar(5)](https://man.freebsd.org/cgi/man.cgi?query=ar&sektion=5)

src/main/java/org/apache/commons/compress/archivers/ar/ArArchiveInputStream.java
```diff
@@ -211,25 +202,25 @@ private String getBSDLongName(final String bsdLongName) throws IOException {
      */
     private String getExtendedName(final int offset) throws IOException {
         if (namebuffer == null) {
-            throw new ArchiveException("Cannot process GNU long file name as no // record was found");
+            throw new ArchiveException("Cannot process GNU long file name as no GNU string table was found");
+        }
+        if (offset < 0 || offset >= namebuffer.length) {
```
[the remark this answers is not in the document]
ppkarwasz: While `ParsingUtils.parseIntValue` will **not** throw an exception for negative values, the parsed string can **not** contain a sign, so this improvement is justified.

src/main/java/org/apache/commons/compress/archivers/ar/ArArchiveInputStream.java
```diff
@@ -147,24 +140,22 @@ public ArArchiveInputStream(final InputStream inputStream) {
         super(inputStream, StandardCharsets.US_ASCII.name());
     }
 
-    private int asInt(final byte[] byteArray, final int offset, final int len) throws IOException {
-        return asInt(byteArray, offset, len, 10, false);
+    private int asOptionalDecimal(final byte[] byteArray, final int offset, final int len) throws IOException {
```
garydgregory: I don't understand these method name changes. What's optional if there's always a return value? Either the method names can remain the same, or some Javadoc would be helpful. WDYT?

src/main/java/org/apache/commons/compress/archivers/ar/ArArchiveInputStream.java
```diff
@@ -241,102 +232,172 @@ private String getExtendedName(final int offset) throws IOException {
...
         }
-        {
-            final byte[] expected = ArchiveUtils.toAsciiBytes(ArArchiveEntry.TRAILER);
-            final byte[] realized = IOUtils.readRange(in, expected.length);
-            final int read = realized.length;
-            trackReadBytes(read);
-            if (read != expected.length) {
-                throw new ArchiveException("Failed to read entry trailer. Occurred at byte: " + getBytesRead());
+
+        currentEntry = new ArArchiveEntry(
```
garydgregory: I'm not a fan of what I call "newspaper" formatting. Why is this API call in a column and not another?

garydgregory (changes requested):
Some minor comments plus `org.apache.commons.compress.archivers.ar.ArArchiveInputStreamTest.testGetNextArEntry()` needs to be enabled and updated.

src/main/java/org/apache/commons/compress/archivers/ar/ArArchiveInputStream.java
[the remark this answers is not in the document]
garydgregory: Then let's leave a link to the spec in the `//` comment.

src/main/java/org/apache/commons/compress/archivers/ar/ArArchiveInputStream.java
ppkarwasz: Since Commons does **not** enforce a deterministic formatting style, I stick to the Palantir Java format so I don’t have to decide when to insert a newline and when not.

I’m fairly agnostic about the specific style, my main goal is to avoid thinking about formatting at all, so if you have a “better” option in mind, I’m happy to switch.

src/main/java/org/apache/commons/compress/archivers/ar/ArArchiveInputStream.java
ppkarwasz: I renamed/added these methods because I slightly changed how `ar` entries are parsed.

Previously, GNU string table entries were handled differently from other `ar` entries:

* For GNU string table entries, only the length and name fields were parsed, since the other fields are usually blank.
* For all other entries, every header field was parsed.

In this PR, I introduced a unified `parseEntry` method that parses both kinds of members. The new `asOptionalDecimal`, etc., methods handle cases where numeric fields are blank (as they often are for the GNU string table).

Initially, I considered adding an `if` to skip parsing `mtime`, `uid`, `gid`, and `mode` for GNU string table entries. But those fields aren’t critical even for normal files, so it’s cleaner to parse them as optional and avoid throwing conversion exceptions when they’re blank.

What do you think? Should we keep the original parsing logic or relax it?

src/main/java/org/apache/commons/compress/archivers/ar/ArArchiveInputStream.java
garydgregory: It does enforce some formatting through Checkstyle (for example, line length). I certainly hope we'll never format our code in this style anywhere, like... not worth mentioning ;-)

src/main/java/org/apache/commons/compress/archivers/ar/ArArchiveInputStream.java
garydgregory: > I renamed/added these methods because I slightly changed how `ar` entries are parsed.
> 
> Previously, GNU string table entries were handled differently from other `ar` entries:
> 
> * For GNU string table entries, only the length and name fields were parsed, since the other fields are usually blank.
> * For all other entries, every header field was parsed.
> 
> In this PR, I introduced a unified `parseEntry` method that parses both kinds of members. The new `asOptionalDecimal`, etc., methods handle cases where numeric fields are blank (as they often are for the GNU string table).
> 
> Initially, I considered adding an `if` to skip parsing `mtime`, `uid`, `gid`, and `mode` for GNU string table entries. But those fields aren’t critical even for normal files, so it’s cleaner to parse them as optional and avoid throwing conversion exceptions when they’re blank.
> 
> What do you think? Should we keep the original parsing logic or relax it?

All of the above doesn't answer my question. I think we leave the behavior as is.

ppkarwasz: Fixed in https://github.com/apache/commons-compress/pull/685/commits/27f247a40d2f984b2dab79a7860b1d2e69088421

src/main/java/org/apache/commons/compress/archivers/ar/ArArchiveInputStream.java
ppkarwasz: Fixed in https://github.com/apache/commons-compress/pull/685/commits/da05e0ecd6860c51934b7751637961cc14011f41

src/main/java/org/apache/commons/compress/archivers/ar/ArArchiveInputStream.java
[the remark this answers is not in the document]
ppkarwasz: Documented in https://github.com/apache/commons-compress/pull/685/commits/bce21564d764e24eb842bba8ce221b307c92a4b0

ppkarwasz:
> Some minor comments plus `org.apache.commons.compress.archivers.ar.ArArchiveInputStreamTest.testGetNextArEntry()` needs to be enabled and updated.

I enabled `testGetNextArEntry` in 1f1f82fcd660546ed05f5cf41d7f37cecdd04d51, end removed the last (only) file entry, so that it fails with an `EOFException` as discussed in https://github.com/apache/commons-compress/pull/685#discussion_r2265367933

PR #687, opened 2025-08-14
Fix grammar issues and typos

mawiesne:
What's the purpose of this PR?
-
- fixes some grammar issues, mainly in JavaDoc and in `examples.xml`
- gets rid of several typos

Tasks
-

- [ ] Write unit tests that match behavioral changes, where the tests fail if the changes to the runtime are not applied. This may not always be possible, but it is a best-practice.

### DIFF
--- a/src/main/java/org/apache/commons/compress/harmony/unpack200/bytecode/ExceptionTableEntry.java
+++ b/src/main/java/org/apache/commons/compress/harmony/unpack200/bytecode/ExceptionTableEntry.java
@@ -44,6 +44,7 @@ public class ExceptionTableEntry {
      * {@code finally} clause (which has no catchType). In the class file, the {@code finally} clause is represented as catchType == 0.
      * <p>
      * To create a {@code finally} clause with this method, pass in null for the catchType.
+     * </p>
      *
      * @param startPC   int
      * @param endPC     int


src/main/java/org/apache/commons/compress/archivers/zip/ExplodingInputStream.java
```diff
@@ -68,7 +68,7 @@ final class ExplodingInputStream extends InputStream implements InputStreamStati
     private long treeSizes;
 
     /**
-     * Constructs a new stream decompressing the content of the specified stream using the explode algorithm.
+     * Constructs a new stream decompressing the content of the specified stream using the <i>explode</i> algorithm.
```
garydgregory: Don't HTML style elements in Javadoc! That's for the stylesheet to decide. If you want to call something out, you can use an emphasis tag like `<em>this is an emphasis example</em>`. Similarly, you'd use `strong` instead of `b`.

mawiesne: Will address it in next 30mins.

mawiesne: @garydgregory Thx for the pointer / hint. FYI: There is one other (existing) use of `<i>` in Pack200Exception line 77. Should I fix this one en passant, or do you prefer a separate PR for that one?

PR #690, opened 2025-08-17
[COMPRESS-706] Add support for reading LHA archive format

fkjellberg:

- [X] Read the [contribution guidelines](CONTRIBUTING.md) for this project.
- [x] I used AI to create any part of, or all of, this pull request.
- [X] Run a successful build using the default [Maven](https://maven.apache.org/) goal with `mvn`; that's `mvn` on the command line by itself.
- [X] Write unit tests that match behavioral changes, where the tests fail if the changes to the runtime are not applied. This may not always be possible, but it is a best-practice.
- [X] Write a pull request description that is detailed enough to understand what the pull request does, how, and why.
- [X] Each commit in the pull request should have a meaningful subject line and body. Note that a maintainer may squash commits during the merge process.

This is an read only implementation of the LHA/LZH archive format supporting a subset (but most commonly used) of the compression algorithms (lh0, lh4, lh5, lh6, lh7).

There is currently support for LHA header levels 0, 1 and 2. I added support for header level 3 but since I have not been able to create such archives myself, I removed the support for now. I may create another PR in the future adding it. It seems only the lha tool on OS/2 ever supported level 3 headers and I don't have access to that platform myself.

A large number of archives were downloaded from FUNET and AMINET for mainly the Amiga and Atari platforms and most of them can be successfully decompressed. The small number of archives that don't work are corrupt in one way or the other and the lha tools I tried are not able to decompress them either. Some very old archives use lh1 compression that is currently not supported by this implementation.

I tried to align the implementation with other similar decompression algorithms found in the commons-compress repo and reuse code wherever possible. The `ExplodingInputStream` from the zip implementation and the `HuffmanDecoder` from deflate64 are using similar compression algorithms.

I ended up reusing the `CircularBuffer` from the zip implementation and I've refactored it into the utils package to make it available for reuse. I added some checks to make sure parts of the buffer that have not been read yet were never overwritten and that distance never goes farther back than the size of the buffer. This is a separate commit in the PR and could possibly be reviewed and merged separately. The `HuffmanDecoder` from deflate64 is using `DecodingMemory` that is similar to `CircularBuffer` but it would require some refactoring of `HuffmanDecoder` to use `CircularBuffer` instead.

The `ExplodingInputStream` and deflate64's `HuffmanDecoder` are both using binary trees for Huffman decoding. I ended up using the `BinaryTree` from `ExplodingInputStream` but since the storage and construction of the actual tree differs, I copied the code into the LHA implementation and kept it package private for now. The common code of this class could possibly go to the utils package and the code to build the tree kept in each implementation package in a future refactoring. The `HuffmanDecoder` could also possibly use such a refactored `BinaryTree` implementation.

I also added a CRC-16 implementation. It is needed in both the archiver and the compressor packages so I put it in the utils package as a public class.


garydgregory (changes requested):
@fkjellberg 

Thank your for the PR. I'll review once the builds are green.

src/main/java/org/apache/commons/compress/utils/CRC16.java
```diff
@@ -0,0 +1,96 @@
...
+import java.util.zip.Checksum;
+
+/**
+ * CRC-16 checksum implementation based on polynomial x^16 + x^15 + x^2 + 1 (0x8005) and
+ * the initial value 0x0000. This CRC variant is also known as CRC-16-MODBUS, CRC-16-IBM
+ * and CRC-16-ANSI.
+ *
+ * @since 1.29
+ */
+public class CRC16 implements Checksum {
```
garydgregory: Keep this class package-private.

garydgregory (changes requested):
Hello @fkjellberg 

The build is still failing.

src/main/java/org/apache/commons/compress/utils/CRC16.java
fkjellberg: Yes, you are right. I'm actually only using CRC16 from LhaArchiveInputStream. I'll fix it.

fkjellberg:
@garydgregory I think I fixed all the build issues now.

garydgregory:
Hello @fkjellberg 

Would you please rebase on git master to avoid the one failing test?
TY!

src/main/java/org/apache/commons/compress/archivers/ArchiveStreamFactory.java
```diff
@@ -175,6 +178,13 @@ public class ArchiveStreamFactory implements ArchiveStreamProvider {
      */
     public static final String JAR = "jar";
 
+    /**
+     * Constant (value {@value}) used to identify the LHA archive format.
+     * Not supported as an output stream type.
+     * @since 1.29
```
garydgregory: The next feature version will be `1.29.0`.

src/main/java/org/apache/commons/compress/archivers/ArchiveStreamFactory.java
```diff
@@ -256,6 +267,20 @@ public static String detect(final InputStream in) throws ArchiveException {
         if (DumpArchiveInputStream.matches(dumpsig, signatureLength)) {
             return DUMP;
         }
+
```
garydgregory: You don't need extra blank lines all over the place, in general, if you want to call something out, use an inline comment, which you do here 😄

src/main/java/org/apache/commons/compress/archivers/lha/LhaArchiveEntry.java
```diff
@@ -0,0 +1,216 @@
...
+    public Optional<Integer> getUnixGroupId() {
+        return unixGroupId;
+    }
+
+    public void setUnixGroupId(Optional<Integer> unixGroupId) {
+        this.unixGroupId = unixGroupId;
+    }
+
+    /**
+     * Returns the MS-DOS file attributes if available for this entry.
```
garydgregory: Javadoc: In general, please use "Gets..." for getters, "Sets..." for setters, and "Tests..." for `isFoo()` methods.

src/main/java/org/apache/commons/compress/archivers/lha/LhaArchiveEntry.java
```diff
@@ -0,0 +1,216 @@
...
+
+    public void setUnixUserId(Optional<Integer> unixUserId) {
+        this.unixUserId = unixUserId;
+    }
+
+    public Optional<Integer> getUnixGroupId() {
+        return unixGroupId;
+    }
+
+    public void setUnixGroupId(Optional<Integer> unixGroupId) {
```
garydgregory: All new public and protected elements need a Javadoc comment.

src/main/java/org/apache/commons/compress/compressors/lha/AbstractLhStaticHuffmanCompressorInputStream.java
```diff
@@ -0,0 +1,344 @@
...
+import org.apache.commons.compress.utils.CircularBuffer;
+import org.apache.commons.compress.utils.InputStreamStatistics;
+import org.apache.commons.io.input.CloseShieldInputStream;
+
+/**
+ * This is an implementation of a static Huffman compressor input stream for LHA files that
+ * supports lh4, lh5, lh6 and lh7 compression methods.
+ *
+ * This implementation is based on the documentation that can be found at
+ * https://github.com/jca02266/lha/blob/master/Hacking_of_LHa
```
garydgregory: Please add a URL to a specification so maintainers know where to look for canonical definitions, especially if a spec version matters. A URL to a random GitHub page doesn't feel authoritative IMO.

src/main/java/org/apache/commons/compress/compressors/lha/AbstractLhStaticHuffmanCompressorInputStream.java
```diff
@@ -0,0 +1,344 @@
...
+            } finally {
+                this.bin = null;
+                this.buffer = null;
+                this.blockSize = -1;
+            }
+        }
+    }
+
+    /**
+     * Get the threshold for copying data from the sliding dictionary. This is the minimum
```
garydgregory: "Get" -> "Gets".

src/main/java/org/apache/commons/compress/compressors/lha/AbstractLhStaticHuffmanCompressorInputStream.java
```diff
@@ -0,0 +1,344 @@
...
+        return getDictionaryBits() + 1;
+    }
+
+    /**
+     * Get the maximum match length for the copy command.
+     *
+     * @return the maximum match length
+     */
+    protected int getMaxMatchLength() {
+        return 256;
```
garydgregory, line 145: Is this magic number carry the same meaning as in the method below? This is where constants with comments help  😉

src/main/java/org/apache/commons/compress/utils/CircularBuffer.java
```diff
@@ -59,11 +62,18 @@ public boolean available() {
      * @param length   the number of bytes to copy
      */
     public void copy(final int distance, final int length) {
+        if (distance < 1) {
+            throw new IllegalArgumentException("Distance must be at least 1");
+        }
+
```
garydgregory: Remove extra blank lines please.

src/main/java/org/apache/commons/compress/compressors/lha/package-info.java
```diff
@@ -0,0 +1,23 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/**
+ * Provides stream classes for decompressing streams found in LHA archives.
```
garydgregory: A URL to what LHA is would help here, even if it's just to Wikipedia.

src/main/java/org/apache/commons/compress/utils/CircularBuffer.java
```diff
@@ -17,14 +17,14 @@
...
+package org.apache.commons.compress.utils;
 
 /**
  * Circular byte buffer.
  *
- * @since 1.7
+ * @since 1.29
  */
-final class CircularBuffer {
+public class CircularBuffer {
```
garydgregory: Might as well leave it final.

src/main/java/org/apache/commons/compress/archivers/lha/LhaArchiveEntry.java
```diff
@@ -0,0 +1,216 @@
...
+
+import org.apache.commons.compress.archivers.ArchiveEntry;
+
+/**
+ * Represents an entry in a LHA archive.
+ *
+ * @since 1.29
+ */
+public class LhaArchiveEntry implements ArchiveEntry {
+    private String name;
```
garydgregory: Can any of this state be final?

src/main/java/org/apache/commons/compress/compressors/lha/AbstractLhStaticHuffmanCompressorInputStream.java
```diff
@@ -0,0 +1,344 @@
...
+/**
+ * This is an implementation of a static Huffman compressor input stream for LHA files that
+ * supports lh4, lh5, lh6 and lh7 compression methods.
+ *
+ * This implementation is based on the documentation that can be found at
+ * https://github.com/jca02266/lha/blob/master/Hacking_of_LHa
+ */
+abstract class AbstractLhStaticHuffmanCompressorInputStream extends CompressorInputStream implements InputStreamStatistics {
+    // Constants for command tree decoding
+    private static final int COMMAND_DECODING_LENGTH_BITS = 5; // Number of bits used to encode the command decoding tree length
```
garydgregory: Replace // comments for constants and ivars with Javadoc, it's much more expressive, easier to edit, especially when referring to specs and other parts of the code.

src/main/java/org/apache/commons/compress/compressors/lha/AbstractLhStaticHuffmanCompressorInputStream.java
```diff
@@ -0,0 +1,344 @@
...
+    /**
+     * Read the command decoding tree. The command decoding tree is used when reading the command tree
+     * which is then actually used to decode the commands (literals or copy commands).
+     *
+     * @return the command decoding tree
+     * @throws IOException if an I/O error occurs
+     */
+    protected BinaryTree readCommandDecodingTree() throws IOException {
+        // Number of code lengths to read
+        final int numCodeLengths = (int) bin.readBits(COMMAND_DECODING_LENGTH_BITS);
```
garydgregory: What if numCodeLengths is -1?

src/main/java/org/apache/commons/compress/compressors/lha/AbstractLhStaticHuffmanCompressorInputStream.java
```diff
@@ -0,0 +1,344 @@
...
+
+        if (numCodeLengths > MAX_NUMBER_OF_COMMAND_DECODING_CODE_LENGTHS) {
+            throw new CompressorException("Code length table has invalid size (%d > %d)", numCodeLengths, MAX_NUMBER_OF_COMMAND_DECODING_CODE_LENGTHS);
+        } else if (numCodeLengths == 0) {
+            // If numCodeLengths is zero, we read a single code length of COMMAND_DECODING_LENGTH_BITS bits and use as root of the tree
+            return new BinaryTree(new int[] { (int) bin.readBits(COMMAND_DECODING_LENGTH_BITS) });
```
garydgregory: What if the result of readBits(int) is -1? I think throw a CompressorException.

src/main/java/org/apache/commons/compress/compressors/lha/AbstractLhStaticHuffmanCompressorInputStream.java
```diff
@@ -0,0 +1,344 @@
...
+        } else {
+            // Read all code lengths
+            final int[] codeLengths = new int[numCodeLengths];
+            for (int index = 0; index < numCodeLengths; index++) {
+                codeLengths[index] = readCodeLength();
+
+                if (index == 2) {
+                    // After reading the first three code lengths, we read a 2-bit skip range
+                    index += (int) bin.readBits(2);
```
garydgregory: Account for -1.

src/main/java/org/apache/commons/compress/compressors/lha/AbstractLhStaticHuffmanCompressorInputStream.java
```diff
@@ -0,0 +1,344 @@
...
+
+    /**
+     * Read the command tree which is used to decode the commands (literals or copy commands).
+     *
+     * @param commandDecodingTree the Huffman tree used to decode the command lengths
+     * @return the command tree
+     * @throws IOException if an I/O error occurs
+     */
+    protected BinaryTree readCommandTree(final BinaryTree commandDecodingTree) throws IOException {
+        final int numCodeLengths = (int) bin.readBits(COMMAND_TREE_LENGTH_BITS);
```
garydgregory: What if the result of readBits(int) is -1?

src/main/java/org/apache/commons/compress/compressors/lha/AbstractLhStaticHuffmanCompressorInputStream.java
```diff
@@ -0,0 +1,344 @@
...
+
+        if (numCodeLengths > getMaxNumberOfCommands()) {
+            throw new CompressorException("Code length table has invalid size (%d > %d)", numCodeLengths, getMaxNumberOfCommands());
+        } else if (numCodeLengths == 0) {
+            // If numCodeLengths is zero, we read a single code length of COMMAND_TREE_LENGTH_BITS bits and use as root of the tree
+            return new BinaryTree(new int[] { (int) bin.readBits(COMMAND_TREE_LENGTH_BITS) });
```
garydgregory: What if the result of readBits(int) is -1? Throw CompressorException I tihnk.

src/main/java/org/apache/commons/compress/compressors/lha/AbstractLhStaticHuffmanCompressorInputStream.java
```diff
@@ -0,0 +1,344 @@
...
+                    index++;
+                } else if (codeOrSkipRange == 1) {
+                    // Skip a range of code lengths, read 4 bits to determine how many to skip
+                    index += (int) bin.readBits(4) + 3;
+                } else if (codeOrSkipRange == 2) {
+                    // Skip a range of code lengths, read 9 bits to determine how many to skip
+                    index += (int) bin.readBits(9) + 20;
+                } else {
+                    // Subtract 2 from the codeOrSkipRange to get the code length
+                    codeLengths[index++] = codeOrSkipRange - 2;
```
garydgregory, line 320: I don't think you want to be here if `commandDecodingTree.read(bin)` returns -1.

src/main/java/org/apache/commons/compress/compressors/lha/AbstractLhStaticHuffmanCompressorInputStream.java
```diff
@@ -0,0 +1,344 @@
...
+        final int numCodeLengths = (int) bin.readBits(getDistanceBits());
+
+        if (numCodeLengths > getDistanceCodeSize()) {
+            throw new CompressorException("Code length table has invalid size (%d > %d)", numCodeLengths, getDistanceCodeSize());
+        } else if (numCodeLengths == 0) {
+            // If numCodeLengths is zero, we read a single code length of getDistanceBits() bits and use as root of the tree
+            return new BinaryTree(new int[] { (int) bin.readBits(getDistanceBits()) });
+        } else {
+            // Read all code lengths
+            final int[] codeLengths = new int[numCodeLengths];
```
garydgregory, line 345: I don't think you want to be here if commandDecodingTree.read(bin) returns -1.

src/main/java/org/apache/commons/compress/compressors/lha/BinaryTree.java
```diff
@@ -0,0 +1,150 @@
...
+     * binary tree as values in the array and the index into the array as the value of the leaf node.
+     *
+     * If the array contains a single value, this is a special case where there is only one node in the
+     * tree (the root node) and it contains the value. For this case, the array contains the value of
+     * the root node instead of the depth in the tree. This special case also means that no bits will
+     * be read from the bit stream when the read method is called, as there are no children to traverse.
+     *
+     * @param array the array to build the binary tree from
+     */
+    BinaryTree(final int[] array) {
```
garydgregory: Use a vararg declaration here and call sites won't have to explicitly declare an array each time.

src/main/java/org/apache/commons/compress/compressors/lha/AbstractLhStaticHuffmanCompressorInputStream.java
```diff
@@ -0,0 +1,344 @@
...
+     */
+    private int readDistance() throws IOException {
+        // Determine the number of bits to read for the distance by reading an entry from the distance tree
+        final int bits = distanceTree.read(bin);
+
+        if (bits == 0 || bits == 1) {
+            // This is effectively run length encoding
+            return bits;
+        } else {
+            // Bits minus one is the number of bits to read for the distance
```
garydgregory, line 370: I don't think you want to be here if commandDecodingTree.read(bin) returns -1.

garydgregory (changes requested):
Hello @fkjellberg 

Thank you for your updates.

If you see updates on git master, please rebase. The build in this branch has some failures unrelated to this new code, so you can ignore those. Changes in master may help, but nothing at the time I am writing this. The master builds are green.

I've left some comments throughout. 

In general, when parsing input, you must assume an end-of-file can happen at any time in the input stream.

You can also assume this will be a new attack vector to try and cause OutOfMemoryError and StackOverflowError from the VM, so see the call sites of `MemoryLimitException` for examples on how to guard the code.

TY!

src/main/java/org/apache/commons/compress/archivers/lha/LhaArchiveInputStream.java
```diff
@@ -0,0 +1,701 @@
...
+    /**
+     * Read the next LHA header from the input stream.
+     *
+     * @return the next header entry, or null if there are no more entries
+     * @throws IOException
+     */
+    protected LhaArchiveEntry readHeader() throws IOException {
+        // Header level is not known yet. Read the minimum length header.
+        final byte[] buffer = new byte[HEADER_GENERIC_MINIMUM_HEADER_LENGTH];
+        final int len = in.read(buffer);
```
ppkarwasz: Using `IOUtils.readFully` prevents short reads, unless the end of file is actually reached:

```suggestion
        final int len = IOUtils.readFully(in, buffer);
```

src/main/java/org/apache/commons/compress/archivers/lha/LhaArchiveEntry.java
```diff
@@ -0,0 +1,216 @@
...
+     *
+     * @return operating system id if available
+     */
+    public Optional<Integer> getOsId() {
+        return osId;
+    }
+
+    public void setOsId(Optional<Integer> osId) {
+        this.osId = osId;
+    }
```
ppkarwasz: I’d avoid using `Optional` as a setter parameter here. `Optional` was designed mainly as a **return type** to signal “no result” without using `null`. As [Brian Goetz explains](https://stackoverflow.com/questions/26327957/should-java-8-getters-return-optional-type/26328555#26328555), it’s not intended for fields, parameters, or JavaBean properties.

Using it in setters makes the API harder to use (callers must wrap values in `Optional`), without adding safety. A plain `Integer` (possibly `null`) is the more conventional and expected choice in this case.

src/main/java/org/apache/commons/compress/archivers/lha/LhaArchiveInputStream.java
```diff
@@ -0,0 +1,701 @@
...
+            // EOF
+            return null;
+        } else if (len == 1 && buffer[0] == 0) {
+            // Last byte of the file is zero indicating no more entries
+            return null;
+        } else if (len < HEADER_GENERIC_MINIMUM_HEADER_LENGTH) {
+            throw new ArchiveException("Invalid header length");
+        }
+
+        final ByteBuffer header = ByteBuffer.wrap(buffer).order(ByteOrder.LITTLE_ENDIAN);
```
ppkarwasz, line 292: It looks like the `ByteBuffer` is only being used to read an `int` from the byte array. To avoid the extra wrapping, you could use [`EndianUtils`](https://commons.apache.org/proper/commons-io/apidocs/org/apache/commons/io/EndianUtils.html) directly for little-endian conversions.

ppkarwasz (changes requested):
Hi @fkjellberg,

Thanks for the PR! I haven’t had the chance to do a full review yet, but here are a few suggestions that could help improve it:

src/main/java/org/apache/commons/compress/archivers/lha/LhaArchiveEntry.java
garydgregory: I agree with @ppkarwasz

fkjellberg:
@ppkarwasz @garydgregory Thanks for the feedback. I'll work on fixing the raised issues in the coming days.

src/main/java/org/apache/commons/compress/archivers/lha/LhaArchiveInputStream.java
fkjellberg: This ByteBuffer is also sent to the methods doing the actual parsing of the headers where I'm using some nice features from ByteBuffer.

I'm using the position feature where it keeps track of the next position to read from and this is used for level 0 and 1 headers where the CRC is right after the filename which can be of variable length.

I'm also reading 8, 16 and 32 bits integer values from the ByteBuffer.

Java 9 added some more features to ByteBuffer that I really needed so I had to add a couple of supporting methods (e.g. `byteBufferSlice`) until this code base can be upgraded to a more recent java version.

src/main/java/org/apache/commons/compress/archivers/ArchiveStreamFactory.java
fkjellberg: I've been working on this PR on and off for over a year and this particular code predates the commit you made on the 30th of May reducing the vertical whitespace and I just aligned with what the code look like back then. :-)

I will fix this file, but I have a lot of "vertical whitespace" in the rest of the PR as well because I think they make the code easier to read and understand. If the coding convention of this project is not have any blank lines, I can update my PR and remove them. Should I do that @garydgregory ?

garydgregory: Hi @fkjellberg 
Fair enough and I'll leave it up to you.

src/main/java/org/apache/commons/compress/archivers/lha/LhaArchiveEntry.java
fkjellberg: Yes, I don't see that we will add support for writing this ancient format. I'll add the builder pattern and make all fields final and only expose this entire ArchiveEntry has read only.

src/main/java/org/apache/commons/compress/archivers/ArchiveStreamFactory.java
fkjellberg: I updated all since annotations to 1.29.0

src/main/java/org/apache/commons/compress/archivers/ArchiveStreamFactory.java
fkjellberg: Removed empty lines from this file

src/main/java/org/apache/commons/compress/archivers/lha/LhaArchiveEntry.java
fkjellberg: I think I fixed this everywhere now

src/main/java/org/apache/commons/compress/archivers/lha/LhaArchiveEntry.java
fkjellberg: I think I fixed this everywhere now. Can we get checkstyle to detect this?

src/main/java/org/apache/commons/compress/compressors/lha/AbstractLhStaticHuffmanCompressorInputStream.java
fkjellberg: I don't think a formal specification document exists for this compression format. I have not found any. This file contains some notes from someone that reversed engineered the format and it's written in Japanese so I used Google Translate so I could read and understand it. I removed this reference and I don't have any better to replace it with.

src/main/java/org/apache/commons/compress/compressors/lha/AbstractLhStaticHuffmanCompressorInputStream.java
fkjellberg: I added checks for EOF and it will throw CompressorException now since EOF is not expected in those places.

src/main/java/org/apache/commons/compress/compressors/lha/BinaryTree.java
fkjellberg: Fixed

src/main/java/org/apache/commons/compress/archivers/lha/LhaArchiveInputStream.java
fkjellberg: Yes, this is actually what I want! I ended up using the IOUtils.read method directly from commons-io instead of using the IOUtils in commons-compress.

src/main/java/org/apache/commons/compress/archivers/lha/LhaArchiveEntry.java
fkjellberg: Understood. I changed them to return Integer instead.

src/main/java/org/apache/commons/compress/compressors/lha/AbstractLhStaticHuffmanCompressorInputStream.java
fkjellberg: I added the constant `NUMBER_OF_LITERAL_CODES`

src/main/java/org/apache/commons/compress/compressors/lha/AbstractLhStaticHuffmanCompressorInputStream.java
garydgregory: Hi @fkjellberg How about this EOF path?

garydgregory (changes requested):
Hi @fkjellberg 
Yeah, don't worry about the memory limit stuff. We need to look over the whole library.
There are some unresolved EOF issues though.
TY!

src/main/java/org/apache/commons/compress/compressors/lha/AbstractLhStaticHuffmanCompressorInputStream.java
```diff
@@ -0,0 +1,389 @@
...
+import org.apache.commons.compress.utils.BitInputStream;
+import org.apache.commons.compress.utils.CircularBuffer;
+import org.apache.commons.compress.utils.InputStreamStatistics;
+import org.apache.commons.io.input.CloseShieldInputStream;
+
+/**
+ * This is an implementation of a static Huffman compressor input stream for LHA files that
+ * supports lh4, lh5, lh6 and lh7 compression methods.
+ */
+abstract class AbstractLhStaticHuffmanCompressorInputStream extends CompressorInputStream implements InputStreamStatistics {
```
ppkarwasz, line 36: Huffman coding is already used in several compressors in Commons Compress (DEFLATE, DEFLATE64, BZip2, Brotli, …), and so are variants of LZ77.

Rather than introducing new ad-hoc helpers for Huffman or LZSS, it would be better to refactor common logic into shared utilities. This helps reduce duplication and makes maintenance easier across codecs.

I’ve started work on a reusable Huffman helper in #701. It hasn’t yet been wired into DEFLATE64, but if it also fits LHA’s needs, we could merge that first and then build LHA on top of it.

For the LZSS side, could the implementation be expressed as extensions to `lz77support` instead of standalone classes?

src/main/java/org/apache/commons/compress/archivers/lha/LhaArchiveInputStream.java
```diff
@@ -0,0 +1,689 @@
...
+    }
+
+    /**
+     * Constructs the LhaArchiveInputStream, taking ownership of the inputStream that is passed in.
+     *
+     * @param inputStream the underlying stream, whose ownership is taken
+     * @param charsetName the charset used for file names in the archive. May be {@code null} to use US-ASCII as default.
+     * @param fileSeparatorChar the character used to separate file path elements
+     */
+    public LhaArchiveInputStream(final InputStream inputStream, final String charsetName, final char fileSeparatorChar) {
```
ppkarwasz: @garydgregory: Most archivers require an `InputStream`, and nearly all of them also need a `Charset` to decode file names.

Rather than adding multiple constructors, what do you think about:

* Adopting the builder pattern from the start, and
* Introducing a shared abstract builder to handle common parameters like `inputStream` and `charset` (or `charsetName`).

src/main/java/org/apache/commons/compress/archivers/lha/LhaArchiveInputStream.java
```diff
@@ -0,0 +1,689 @@
...
+    }
+
+    /**
+     * Read LHA header level 0.
+     *
+     * @param buffer the buffer containing the header data
+     * @return the LhaArchiveEntry read from the buffer
+     * @throws IOException
+     */
+    protected LhaArchiveEntry readHeaderLevel0(ByteBuffer buffer) throws IOException {
```
ppkarwasz: Please avoid exposing methods to the public API unless strictly necessary. Keep them private or package-private where possible: once they’re `protected` or `public`, we can’t remove them later without breaking compatibility.

Also note that `protected/public` methods should validate their parameters, which adds extra boilerplate. Keeping methods non-public helps keep the code smaller and cleaner.

src/main/java/org/apache/commons/compress/compressors/lha/AbstractLhStaticHuffmanCompressorInputStream.java
ppkarwasz: I found this on the WayBack machine: https://web.archive.org/web/20120211104049/http://homepage1.nifty.com/dangan/en/Content/Program/Java/jLHA/Notes/Notes.html

src/main/java/org/apache/commons/compress/archivers/lha/LhaArchiveInputStream.java
```diff
@@ -0,0 +1,689 @@
...
+            // EOF
+            return null;
+        } else if (len == 1 && buffer[0] == 0) {
+            // Last byte of the file is zero indicating no more entries
+            return null;
+        } else if (len < HEADER_GENERIC_MINIMUM_HEADER_LENGTH) {
+            throw new ArchiveException("Invalid header length");
+        }
+
+        final ByteBuffer header = ByteBuffer.wrap(buffer).order(ByteOrder.LITTLE_ENDIAN);
```
ppkarwasz, line 292: I noticed that while most of the code uses `ByteBuffer`, there are still a few places where a `byte[]` is allocated and then wrapped. Would it make sense to consistently use `ByteBuffer` throughout?

As you pointed out in https://github.com/apache/commons-compress/pull/690#discussion_r2291084875, `ByteBuffer` is useful here. If we stick with it exclusively, we could also have this class read directly from a `ReadableByteChannel` created from the `InputStream`:

```java
ReadableByteChannel channel = Channels.newChannel(inputStream);
```

This would remove the need for intermediate `byte[]` allocations.

I don’t think it’s necessary to allocate multiple `ByteBuffer` instances here. As far as I understand:

* Level-0 through Level-2 base headers are at most 256 bytes.
* Extended headers for Level-1 and Level-2 are also limited to 2^16 bytes.

Given that, it should be sufficient to keep just two buffers as class fields: one for the base header and one for extended headers, instead of creating new buffers each time.

src/main/java/org/apache/commons/compress/archivers/lha/LhaArchiveInputStream.java
```diff
@@ -0,0 +1,689 @@
...
+     * and '\' chars will be converted into the configured file path separator char.
+     * Any leading file path separator char will be removed to avoid extracting to
+     * absolute locations.
+     *
+     * @param buffer the buffer where to get the pathname from
+     * @param pathnameLength the length of the pathname
+     * @return pathname
+     */
+    protected String getPathname(final ByteBuffer buffer, final int pathnameLength) {
+        final byte[] pathnameBuffer = new byte[pathnameLength];
```
ppkarwasz: Am I correct that `pathnameLength` is always < 2¹⁶? If so, it might be helpful to note this in a comment: it makes it clear to readers that `new byte[pathnameLength]` won’t risk creating an unexpectedly large buffer. Alternatively you use `char` as type.

src/main/java/org/apache/commons/compress/archivers/lha/LhaArchiveInputStream.java
```diff
@@ -0,0 +1,689 @@
...
+        for (int i = 0; i < pathnameLength; i++) {
+            if (pathnameBuffer[i] == (byte) 0xFF) {
+                if (i > start) {
+                    // Decode the path segment into a string using the specified charset and append it to the result
+                    pathnameStringBuilder.append(new String(pathnameBuffer, start, i - start, getCharset())).append(fileSeparatorChar);
+                }
+
+                start = i + 1; // Move start to the next segment
+            }
+        }
```
ppkarwasz, line 565: Since almost the entire class uses NIO, should we also use `CharsetDecoder` and `CharBuffer` here?

ppkarwasz (changes requested):

src/main/java/org/apache/commons/compress/compressors/lha/AbstractLhStaticHuffmanCompressorInputStream.java
fkjellberg: Yes, I agree that the Huffman code from different compressors should be aligned for better code reuse. That was my plan to work on after this PR had been merged. You beat me to it. :-)

I reused the BinaryTree from the ExplodingInputStream class in the zip package which is extremely similar to the LHx compressors in this PR. I think that the Huffman code from the zip package, this PR, Deflate64's HuffmanDecoder and the implementation in Bzip2CompressorInputStream should be aligned and reuse as much code as possible. What seems to differ the most is how the tree is stored in the bit stream and the huffman tree reconstructed at decompress so I guess that has to be implementation specific.

I looked at the AbstractLZ77CompressorInputStream for reused while working on this code, but decided not to try to use it. There is more similarity between this implementation and the ExplodingInputStream so I think they are better candidates for future refactoring.

src/main/java/org/apache/commons/compress/compressors/lha/AbstractLhStaticHuffmanCompressorInputStream.java
fkjellberg: I do check for EOF when `codeOrSkipRange` returns -1 above? What am I missing?

src/main/java/org/apache/commons/compress/compressors/lha/AbstractLhStaticHuffmanCompressorInputStream.java
fkjellberg: Yes, I have that link in the javadoc for the archive parsing code in LhaArchiveInputStream. It only documents the archive format and not the LH5 compression method.

src/main/java/org/apache/commons/compress/archivers/lha/LhaArchiveInputStream.java
fkjellberg: I used `Charset` for this parameter initially but since ArchiveStreamFactory is using a String for this, I changed it. I think that we should use a Charset instead and let the client supply it instead of possibly throwing a UnsupportedCharsetException inside the constructor of the `ArchiveInputStream` class.

src/main/java/org/apache/commons/compress/archivers/lha/LhaArchiveInputStream.java
fkjellberg: Yes, except for level 3 header. Support for level 3 is not included in this PR, but I have it sitting in a branch for a future PR once I'm able to write some tests for it.

For level 0 and 1 (unless an extended header overrides it) headers, the file length is maximum 256 bytes.

The LHA archive format is a really old format mostly used on MS-DOS, Amiga and Atari back in the 80s and 90s. I think it makes sense to introduce a smaller hard limit here, e.g. max 2048 or 4096 bytes? What do you think @ppkarwasz ?

src/main/java/org/apache/commons/compress/archivers/lha/LhaArchiveInputStream.java
fkjellberg: Yes, makes sense. I changed all protected methods to package private instead.

src/main/java/org/apache/commons/compress/archivers/lha/LhaArchiveInputStream.java
garydgregory: @fkjellberg 
Yes, please follow the builder pattern, for example: `TarArchiveInputStream`.
Here's another example if you need a hierachy: `org.apache.commons.io.input.MessageDigestInputStream`

src/main/java/org/apache/commons/compress/archivers/lha/LhaArchiveInputStream.java
fkjellberg: For level 1, you don't know the size of all extended headers upfront. You need to read them one by one to figure out how many they are and the final total size of the header. I collect them in a list to be able to calculate the CRC of the entire header later on.

For level 2, you'll find the total size of the header once you have read the minimal header (22 bytes) to figure out which header level is used.

For level 3 (not supported in this PR), the extended headers can be much larger.

src/main/java/org/apache/commons/compress/archivers/lha/LhaArchiveInputStream.java
fkjellberg: Possibly, but I only find a `CharsetDecoder.decode(ByteBuffer)` that will decode the entire ByteBuffer and no way of selecting just a part of it? What am I missing?

src/main/java/org/apache/commons/compress/archivers/lha/LhaArchiveInputStream.java
ppkarwasz: The exact value of the limit doesn't really matter, as long as it is reasonable and does not cause OOMs on corrupted streams.

`libarchive` uses a 1 MiB limit on the pathname, which is also fine. With @garydgregory we thought about introducing a library-wide limit, maybe based on [`FileSystem#getMaxPathLength()`](https://commons.apache.org/proper/commons-io/apidocs/org/apache/commons/io/FileSystem.html#getMaxPathLength()).

src/main/java/org/apache/commons/compress/archivers/lha/LhaArchiveInputStream.java
ppkarwasz: Looking at the implementation, the decoder decodes the buffer from `position()` to `limit()`. We could use that with something like:

```java
// Decode the pathname and replace any 0xFF bytes with the file separator char
final CharBuffer charBuffer = getCharBuffer().clear();
final CharsetDecoder decoder = getCharset().newDecoder();
int start = buffer.position();
for (int i = start; i < limit; i++) {
    if (buffer.get(i) == (byte) 0xFF) {
        if (i > start) {
            // Decode the path segment into the char buffer using the specified charset
            decode(decoder, buffer.duplicate().position(start).limit(i), charBuffer);
            charBuffer.append(fileSeparatorChar);
        }
        start = i + 1; // Move start to the next segment
        buffer.position(start);
    }
}
```

where the `decode` helper checks for encoding errors:

```java
private void decode(final CharsetDecoder decoder, ByteBuffer in, CharBuffer out) throws ArchiveException {
    throwIfError(decoder.decode(in, out, true));
    throwIfError(decoder.flush(out));
}

private void throwIfError(final CoderResult result) throws ArchiveException {
    if (result.isError()) {
        try {
            result.throwException();
        } catch (final Exception e) {
            throw new ArchiveException("Invalid path segment", (Throwable) e);
        }
    }
}
```

ppkarwasz:
Hi @theobisproject,

Thanks! Since this PR introduces an entire new archive format, it will be a marathon, not a sprint.
It would be nice if you can also run the tests once some obvious errors will be removed.

src/main/java/org/apache/commons/compress/archivers/lha/LhaArchiveInputStream.java
fkjellberg: I added a check and will throw an error if the pathname is longer than 4096 bytes. I've downloaded a lot of files from Aminet and Funet for testing and the longest pathname I've seen so far is 89 characters.

src/main/java/org/apache/commons/compress/archivers/lha/LhaArchiveInputStream.java
fkjellberg: Hi @garydgregory , I added the builder pattern to the LhaArchiveInputStream class. I used the Charset class instead of charset name string for configuring the archive stream.

src/main/java/org/apache/commons/compress/archivers/lha/LhaArchiveInputStream.java
fkjellberg: I don't think that we should throw an error for decoding issues. There may be consumers that will never write the output to a file system but just want to consume the decompressed stream, e.g. Apache Tika or training AI models. Throwing an exception would force them to try a number of different encodings until finding the right one. The current code is replacing unknown characters with the unicode replacement character \uFFFD.

theobisproject:
Thanks for fixing the corrupt archives. I made a new run with the changes and found 6 more all attached in the zip. 
[lha-fuzzing-2.zip](https://github.com/user-attachments/files/22349371/lha-fuzzing-2.zip)

garydgregory:
@fkjellberg @ppkarwasz 
Please rebase on git master, resolve conflicts and we can bring this in.

fkjellberg:
Hi @garydgregory 

Thanks for pushing this PR forward. I rebased and aligned the code with the latest changes from the master branch.

I also replaced the CRC16 implementation with the new Crc16 class from commons-codec. Thanks for moving the CRC code over to commons-codec and adding support for several more CRC variants.

I asked Claude Code for a review of this PR and it found some issues that I think are worth fixing before merging. My plan is to update the PR again in the coming days.

garydgregory:
@fkjellberg 
Thank you the update. In the meantime, I triggered the builds.

fkjellberg:
@garydgregory I pushed some more commits for issues found by Claude when I asked for a review.

I decided to revert the change to move CircularBuffer to the public API to share it between zip and lha. It turned out that it had some implications for the existing use in ExplodingInputStream. Once this PR has been merged, I think we can revisit the refactoring of CircularBuffer in a separate PR.

fkjellberg:
Hi @garydgregory 

Thank you for merging and continuing improving and refactoring the LHA code. I think that we should keep the LhStaticHuffmanCompressorInputStream in the public API as there are other use cases for it other than lha archives. I've come across [TCompress](https://spis.co.nz/compress.htm) which is a Delphi component used to store LH5 compressed data in databases. This is one of the reasons I started to implement LH5 for commons-compress as I need to decompress data stored by a legacy Delphi application.

garydgregory:
Hello @fkjellberg 
Is such a file not readable via the current public API? I think this could be verified with a unit test.

fkjellberg:
Hello @garydgregory 
No it is not possible. TCompress is just implementing the lh5 compression and the compressed data is stored in the database as a blob by Delphi. There is no LHA archive envelope or anything. To be able to decompress such data blob, you need to create a Lh5CompressorInputStream from a byte array.

I will add a few more unit tests as part of my refactor to use the generic HuffmanDecoder instead of BinaryTree that will show how I need to read a TCompress lh5 stream.

PR #697, opened 2025-08-25
feat: Align DUMP archive block size with Linux `dump` utility

ppkarwasz:
This update standardizes the block size limit in `DumpArchiveInputStream` to match the value used by the Linux `dump` utility for ext2/3/4 filesystems.

By using the same limit as the reference implementation, we ensure consistent parsing behavior and improve compatibility with archives generated by standard `dump` tools.

- [x] I used AI to: proofread comments

### DIFF
--- a/src/main/java/org/apache/commons/compress/archivers/dump/DumpArchiveConstants.java
+++ b/src/main/java/org/apache/commons/compress/archivers/dump/DumpArchiveConstants.java
@@ -135,14 +135,31 @@ public final class DumpArchiveConstants {
     public static final int TP_SIZE = 1024;
 
     /**
-     * NTREC value {@value}.
+     * Minimum number of kilobytes per record: {@value}.
+     */
+    public static final int MIN_NTREC = 1;
+
+    /**
+     * Default number of kilobytes per record: {@value}.
      */
     public static final int NTREC = 10;
 
     /**
-     * HIGH_DENSITY_NTREC value {@value}.
+     * Number of kilobytes per record for high density tapes: {@value}.
      */
     public static final int HIGH_DENSITY_NTREC = 32;
+
+    /**
+     * Maximum number of kilobytes per record: {@value}.
+     * <p>
+     *     This limit matches the one used by the Linux ext2/3/4 dump/restore utilities.
+     *     For more details, see the
+     *     <a href="https://dump.sourceforge.io/">Linux dump/restore utilities documentation</a>
+     *     and the
+     *     <a href="https://manpages.debian.org/unstable/dump/dump.8.en.html#b,">dump(8) man page</a>.
+     * </p>
+     */
+    public static final int MAX_NTREC = 1024;
 
     /**
      * OFS_MAGIC value {@value}.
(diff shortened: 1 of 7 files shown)


src/main/java/org/apache/commons/compress/archivers/dump/DumpArchiveConstants.java
```diff
@@ -135,15 +135,32 @@ public static SEGMENT_TYPE find(final int code) {
     public static final int TP_SIZE = 1024;
 
     /**
-     * NTREC value {@value}.
+     * Minimum number of kilobytes per record: {@value}.
+     */
```
garydgregory, line 139: New public and protected elements need a Javadoc `@since` tag. The question really is: Does this really need to be public or can it be package-private or private?

src/main/java/org/apache/commons/compress/archivers/dump/DumpArchiveConstants.java
```diff
@@ -135,15 +135,32 @@ public static SEGMENT_TYPE find(final int code) {
...
 
+    /**
+     * Maximum number of kilobytes per record: {@value}.
+     * <p>
+     *     This limit matches the one used by the Linux ext2/3/4 dump/restore utilities.
+     *     For more details, see the
+     *     <a href="https://dump.sourceforge.io/">Linux dump/restore utilities documentation</a>
+     *     and the
+     *     <a href="https://manpages.debian.org/unstable/dump/dump.8.en.html#b,">dump(8) man page</a>.
+     * </p>
```
garydgregory, line 160: New public and protected elements need a Javadoc `@since` tag. The question really is: Does this really need to be public or can it be package-private or private?

garydgregory (changes requested):
Hi @ppkarwasz 
I commented on 2 small items.

src/main/java/org/apache/commons/compress/archivers/dump/DumpArchiveConstants.java
ppkarwasz: Fixed in https://github.com/apache/commons-compress/pull/697/commits/9e21d6adec7d56cd6f9f411969e4643d483da890, although I am wondering whether we need to make the constants public.

What do you think?

src/main/java/org/apache/commons/compress/archivers/dump/DumpArchiveConstants.java
ppkarwasz: Fixed in https://github.com/apache/commons-compress/pull/697/commits/9e21d6adec7d56cd6f9f411969e4643d483da890

src/main/java/org/apache/commons/compress/archivers/dump/DumpArchiveConstants.java
garydgregory: For now, I would say keep them private. We can make them public later but not the other way around.

ppkarwasz: Fixed in https://github.com/apache/commons-compress/pull/697/commits/5d443fd1c168979e0db35382219e895b26baae7a

PR #698, opened 2025-08-26
fix: Restore incremental file name handling

ppkarwasz:
The refactoring in #684 (never released) inadvertently dropped a critical part of long file name handling: names must be read incrementally to prevent excessive memory allocation if an entry is corrupted.

This change restores the previous safe behavior and adds two unit tests to ensure:

* POSIX and NTFS file names up to the maximum allowed length are handled correctly.
* Corrupted entries are detected and throw an `ArchiveException`.

- [x] I used AI to improve the unit tests.

### DIFF
--- a/src/main/java/org/apache/commons/compress/archivers/tar/TarUtils.java
+++ b/src/main/java/org/apache/commons/compress/archivers/tar/TarUtils.java
@@ -890,7 +890,7 @@ public class TarUtils {
         final long size = entry.getSize();
         // The encoding requires a byte array, whose size must be a positive int.
         if (size > Integer.MAX_VALUE) {
-            throw new ArchiveException("Invalid long name entry: size %d exceeds maximum allowed.", entry.getSize());
+            throw new ArchiveException("Invalid long name entry: size %,d exceeds maximum allowed.", entry.getSize());
         }
         // Read the long name incrementally to limit memory allocation in case of a corrupted entry.
         final BoundedInputStream boundedInput = BoundedInputStream.builder()
@@ -903,7 +903,7 @@ public class TarUtils {
                 .get();
         final long read = org.apache.commons.io.IOUtils.copyLarge(boundedInput, outputStream);
         if (read != size) {
-            throw new ArchiveException("Truncated long name entry: expected %d bytes, read %d bytes.", size, read);
+            throw new ArchiveException("Truncated long name entry: expected %,d bytes, read %,d bytes.", size, read);
         }
         final byte[] name = outputStream.toByteArray();
         int length = name.length;
--- a/src/main/java/org/apache/commons/compress/archivers/tar/TarUtils.java
+++ b/src/main/java/org/apache/commons/compress/archivers/tar/TarUtils.java
@@ -898,9 +898,7 @@ public class TarUtils {
                 .setMaxCount(size)
                 .setPropagateClose(false)
                 .get();
-        final UnsynchronizedByteArrayOutputStream outputStream = UnsynchronizedByteArrayOutputStream.builder()
-                .setBufferSize(org.apache.commons.io.IOUtils.DEFAULT_BUFFER_SIZE)
-                .get();
+        final UnsynchronizedByteArrayOutputStream outputStream = UnsynchronizedByteArrayOutputStream.builder().get();
         final long read = org.apache.commons.io.IOUtils.copyLarge(boundedInput, outputStream);
         if (read != size) {
             throw new ArchiveException("Truncated long name entry: expected %,d bytes, read %,d bytes.", size, read);
--- a/src/main/java/org/apache/commons/compress/archivers/tar/TarUtils.java
+++ b/src/main/java/org/apache/commons/compress/archivers/tar/TarUtils.java
@@ -38,7 +38,9 @@ import org.apache.commons.compress.archivers.zip.ZipEncodingHelper;
 import org.apache.commons.compress.utils.ArchiveUtils;
 import org.apache.commons.compress.utils.IOUtils;
 import org.apache.commons.compress.utils.ParsingUtils;
+import org.apache.commons.io.input.BoundedInputStream;
 import org.apache.commons.io.output.ByteArrayOutputStream;
+import org.apache.commons.io.output.UnsynchronizedByteArrayOutputStream;
 
 /**
  * This class provides static utility methods to work with byte streams.
@@ -883,22 +885,32 @@ public class TarUtils {
      * @throws IOException if an I/O error occurs or the entry is truncated.
      * @throws ArchiveException if the entry size is invalid.
      */
-    private static String readLongName(final InputStream input, final ZipEncoding encoding, final TarArchiveEntry entry)
+    static String readLongName(final InputStream input, final ZipEncoding encoding, final TarArchiveEntry entry)
             throws IOException {
         final long size = entry.getSize();
+        // The encoding requires a byte array, whose size must be a positive int.
         if (size > Integer.MAX_VALUE) {
-            throw new ArchiveException("Invalid long name size: " + entry.getSize());
+            throw new ArchiveException("Invalid long name size: %d.", entry.getSize());
         }
-        final int sizeInt = (int) size;
-        final byte[] buffer = new byte[sizeInt];
-        if (IOUtils.readFully(input, buffer, 0, sizeInt) < sizeInt) {
-            throw new ArchiveException("TAR entry is truncated.");
+        // Read the long name incrementally to limit memory allocation in case of a corrupted entry.
+        final BoundedInputStream boundedInput = BoundedInputStream.builder()
+                .setInputStream(input)
+                .setMaxCount(size)
+                .setPropagateClose(false)
+                .get();
+        final UnsynchronizedByteArrayOutputStream outputStream = UnsynchronizedByteArrayOutputStream.builder()
+                .setBufferSize(org.apache.commons.io.IOUtils.DEFAULT_BUFFER_SIZE)
+                .get();
+        final long read = org.apache.commons.io.IOUtils.copyLarge(boundedInput, outputStream);
+        if (read != size) {
+            throw new ArchiveException("Truncated long name entry: Expected %d bytes, read %d bytes.", size, read);
         }
-        int length = buffer.length;
-        while (length > 0 && buffer[length - 1] == 0) {
+        final byte[] name = outputStream.toByteArray();
+        int length = name.length;
+        while (length > 0 && name[length - 1] == 0) {
             length--;
         }
-        return encoding.decode(Arrays.copyOf(buffer, length));
+        return encoding.decode(Arrays.copyOf(name, length));
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/commons/compress/archivers/tar/TarUtils.java
```diff
@@ -883,22 +885,32 @@ private static long[] readLineOfNumberForPax1x(final InputStream inputStream) th
...
-            throw new ArchiveException("Invalid long name size: " + entry.getSize());
-        }
-        final int sizeInt = (int) size;
-        final byte[] buffer = new byte[sizeInt];
-        if (IOUtils.readFully(input, buffer, 0, sizeInt) < sizeInt) {
-            throw new ArchiveException("TAR entry is truncated.");
-        }
-        int length = buffer.length;
-        while (length > 0 && buffer[length - 1] == 0) {
+            throw new ArchiveException("Invalid long name entry: size %d exceeds maximum allowed.", entry.getSize());
```
garydgregory: `%d` -> use locale-specific grouping characters with `%,d`

src/main/java/org/apache/commons/compress/archivers/tar/TarUtils.java
```diff
@@ -883,22 +885,32 @@ private static long[] readLineOfNumberForPax1x(final InputStream inputStream) th
...
+                .setInputStream(input)
+                .setMaxCount(size)
+                .setPropagateClose(false)
+                .get();
+        final UnsynchronizedByteArrayOutputStream outputStream = UnsynchronizedByteArrayOutputStream.builder()
+                .setBufferSize(org.apache.commons.io.IOUtils.DEFAULT_BUFFER_SIZE)
+                .get();
+        final long read = org.apache.commons.io.IOUtils.copyLarge(boundedInput, outputStream);
+        if (read != size) {
+            throw new ArchiveException("Truncated long name entry: expected %d bytes, read %d bytes.", size, read);
```
garydgregory: `%d` -> use locale-specific grouping characters with `%,d`.

src/main/java/org/apache/commons/compress/archivers/tar/TarUtils.java
```diff
@@ -883,22 +885,32 @@ private static long[] readLineOfNumberForPax1x(final InputStream inputStream) th
...
+            throw new ArchiveException("Invalid long name entry: size %d exceeds maximum allowed.", entry.getSize());
+        }
+        // Read the long name incrementally to limit memory allocation in case of a corrupted entry.
+        final BoundedInputStream boundedInput = BoundedInputStream.builder()
+                .setInputStream(input)
+                .setMaxCount(size)
+                .setPropagateClose(false)
+                .get();
+        final UnsynchronizedByteArrayOutputStream outputStream = UnsynchronizedByteArrayOutputStream.builder()
+                .setBufferSize(org.apache.commons.io.IOUtils.DEFAULT_BUFFER_SIZE)
```
garydgregory: You don't need a FQCN here.

src/main/java/org/apache/commons/compress/archivers/tar/TarUtils.java
```diff
@@ -883,22 +885,32 @@ private static long[] readLineOfNumberForPax1x(final InputStream inputStream) th
...
+                .get();
+        final long read = org.apache.commons.io.IOUtils.copyLarge(boundedInput, outputStream);
```
garydgregory, line 902: You don't need a FQCN here.

src/test/java/org/apache/commons/compress/archivers/tar/TarUtilsTest.java
```diff
@@ -599,4 +607,121 @@ void testWriteNegativeBinary8Byte() {
...
+                Arguments.of("Empty (padded)", empty, paddedUtf8Bytes(empty)),
+                Arguments.of("NTFS", ntfsLongName, utf8Bytes(ntfsLongName)),
+                Arguments.of("NTFS (padded)", ntfsLongName, paddedUtf8Bytes(ntfsLongName)),
+                Arguments.of("POSIX", posixLongName, utf8Bytes(posixLongName)),
+                Arguments.of("POSIX (padded)", posixLongName, paddedUtf8Bytes(posixLongName)));
+    }
+
+    @ParameterizedTest(name = "{0} long name is read correctly")
+    @MethodSource
+    void readLongNameHandlesLimits(final String kind, final String expectedName, final byte[] data) throws IOException {
```
garydgregory: Prefix test methods with `test`, like all other test methods in this class.

src/test/java/org/apache/commons/compress/archivers/tar/TarUtilsTest.java
```diff
@@ -599,4 +607,121 @@ void testWriteNegativeBinary8Byte() {
...
+
+    static Stream<Arguments> readLongNameThrowsOnTruncation() {
+        return Stream.of(
+                Arguments.of(Integer.MAX_VALUE, "truncated long name"),
+                Arguments.of(Long.MAX_VALUE, "invalid long name"));
+    }
+
+    @ParameterizedTest(name = "readLongName of {0} bytes throws ArchiveException")
+    @MethodSource
+    void readLongNameThrowsOnTruncation(final long size, final CharSequence expectedMessage) throws IOException {
```
garydgregory: Prefix test methods with `test`, like all other test methods in this class.

src/test/java/org/apache/commons/compress/archivers/tar/TarUtilsTest.java
```diff
@@ -599,4 +607,121 @@ void testWriteNegativeBinary8Byte() {
...
+        final String prefix = "\\\\?\\C:\\";
+        final String extension = ".txt";
+
+        // U+2605 BLACK STAR (BMP, 1 UTF-16 unit, 3 UTF-8 bytes) => lets us pack 255 units per segment easily
+        final String segment = StringUtils.repeat("★", 255);
+        assertEquals(255, segment.length(), "Segment length should be 255 UTF-16 code units");
+
+        final StringBuilder sb = new StringBuilder(prefix);
+        while (sb.length() + extension.length() < totalUnits) {
+            sb.append(segment).append('\\');
```
garydgregory: WDYT about calling `append()` once if you build the argument outside the loop?

src/test/java/org/apache/commons/compress/archivers/tar/TarUtilsTest.java
```diff
@@ -599,4 +607,121 @@ void testWriteNegativeBinary8Byte() {
...
+    /**
+     * Builds an NTFS-style path (\\?\C:\...) up to a target total UTF-16 length, respecting 255-unit segments.
+     */
+    private static String createNtfsLongNameByUtf16Units(final int totalUnits) {
+        final String prefix = "\\\\?\\C:\\";
+        final String extension = ".txt";
+
+        // U+2605 BLACK STAR (BMP, 1 UTF-16 unit, 3 UTF-8 bytes) => lets us pack 255 units per segment easily
+        final String segment = StringUtils.repeat("★", 255);
+        assertEquals(255, segment.length(), "Segment length should be 255 UTF-16 code units");
```
garydgregory: I guess this assertion is a sanity check; the `255` magic number is the same as the line above, so it could be refactored into a local variable for clarity.

garydgregory (changes requested):
Hi @ppkarwasz 

I left a couple of comments throughout.
TY 😄

src/main/java/org/apache/commons/compress/archivers/tar/TarUtils.java
ppkarwasz: Fixed in https://github.com/apache/commons-compress/pull/698/commits/20f6b14207e744f02cbb2735ced07453780fb6b5

src/main/java/org/apache/commons/compress/archivers/tar/TarUtils.java
ppkarwasz: Fixed in https://github.com/apache/commons-compress/pull/698/commits/20f6b14207e744f02cbb2735ced07453780fb6b5

src/main/java/org/apache/commons/compress/archivers/tar/TarUtils.java
ppkarwasz: This class already imports `o.a.c.compress.utils.IOUtils`, so I need the FQCN to refer to `o.a.c.io.IOUtils`.

However, looking at the implementation of `UnsynchronizedByteArrayOutputStream`, I don't really need it since this is already the default buffer size and the same buffer size is used in `IOUtils.copyLarge`. Fixed in https://github.com/apache/commons-compress/pull/698/commits/2aafade4dd65e0a8d74a6a6e880cfa7c0434c459.

src/main/java/org/apache/commons/compress/archivers/tar/TarUtils.java
ppkarwasz: Here I really do, since all the `o.a.c.compress.utils.IOUtils.copy` methods are deprecated in favor of `o.a.c.io.IOUtils.copy` methods, which in turn use `copyLarge` under the hood.

src/test/java/org/apache/commons/compress/archivers/tar/TarUtilsTest.java
ppkarwasz: Fixed in https://github.com/apache/commons-compress/pull/698/commits/379e8d3a08172bf0a709625962276cd9b3b12569

src/test/java/org/apache/commons/compress/archivers/tar/TarUtilsTest.java
ppkarwasz: Fixed in https://github.com/apache/commons-compress/pull/698/commits/379e8d3a08172bf0a709625962276cd9b3b12569

src/test/java/org/apache/commons/compress/archivers/tar/TarUtilsTest.java
ppkarwasz: Good call! :100:
Fixed in https://github.com/apache/commons-compress/pull/698/commits/9f382cd120cc1d195f1041c411fd6e1b322c3fe9

src/test/java/org/apache/commons/compress/archivers/tar/TarUtilsTest.java
ppkarwasz: This assertion is a copied from the method below, were it is not obvious how many UTF-8 characters are necessary to encode 85 black stars. Fixed in https://github.com/apache/commons-compress/pull/698/commits/9f382cd120cc1d195f1041c411fd6e1b322c3fe9

src/main/java/org/apache/commons/compress/archivers/tar/TarUtils.java
garydgregory: Oh, I see we already import `org.apache.commons.compress.utils.IOUtils` in this class.

PR #699, opened 2025-08-26
feat: enforce 20-bit Huffman code-length limit consistently

ppkarwasz:
This PR improves `BZip2CompressorInputStream` by enforcing the mandated Huffman code-length bound and right-sizing related tables. It aligns behavior with the reference C implementation’s intent while avoiding its historical over-allocation.

## Why

The reference C code deliberately **over-sizes** its Huffman structures as a defensive programming technique: e.g., it defines `BZ_MAX_CODE_LEN = 23` and sizes length-indexed tables by `BZ_MAX_ALPHA_SIZE` rather than the code-length bound to provide safety margins. However, the bzip2 bitstream only permits code lengths ≤ 20 (and the 1.0.6 implementation effectively tightens this to 17), so carrying those defensive constants forward in our code obscures the true limit and keeps arrays larger than necessary.

## What’s changed

* **Set the true limit**: `MAX_CODE_LEN` now equals **20** (the format maximum).

* **Validate inputs**: code lengths are checked explicitly; any value **outside `[1, 20]`** triggers a clear exception early in decoding.

* **Right-size arrays**: Huffman tables and auxiliary arrays are sized to the minimum required for the 20-bit limit, reducing footprint and clarifying invariants.

* **Tests**: add a unit test that exercises the maximum alphabet size across boundary code lengths.

## Maintenance & performance

* Clearer invariants (the limit you see is the limit you enforce).
* Slightly lower memory usage due to smaller tables.
* Easier reasoning about bounds and indexing during review and future changes.

## References

* [bzip2/huffman.c](https://github.com/libarchive/bzip2/blob/master/huffman.c)
* [bzip2/bzlib_private.h](https://github.com/libarchive/bzip2/blob/master/bzlib_private.h)
 
The reference implementation uses over-sized constants for safety but enforces an effective 20-bit maximum.

### DIFF
--- a/src/changes/changes.xml
+++ b/src/changes/changes.xml
@@ -58,7 +58,7 @@ The <action> type attribute can be add,update,fix,remove.
       <action type="fix" dev="ggregory" due-to="Gary Gregory">Don't loose precision while reading folders from a SevenZFile.</action>
       <action type="fix" dev="ggregory" due-to="Roel van Dijk, Gary Gregory">Improve some exception messages in TarUtils and TarArchiveEntry.</action>
       <!-- FIX bzip2 -->      
-      <action type="fix" dev="ggregory" due-to="Tyler Nighswander, Gary Gregory">org.apache.commons.compress.compressors.bzip2.BZip2CompressorInputStream constructors now throws CompressorException instead of ArrayIndexOutOfBoundsException.</action>
+      <action type="fix" dev="pkarwasz" due-to="Tyler Nighswander, Piotr P. Karwasz, Gary Gregory">Enforce 20-bit Huffman code-length limit consistently in BZip2.</action>
       <action type="fix" dev="ggregory" due-to="Gary Gregory">org.apache.commons.compress.compressors.bzip2.BZip2CompressorInputStream throws CompressorException (an IOException subclass) instead of IOException when it encounters formatting problems.</action>
       <!-- FIX dump -->      
       <action type="fix" dev="pkarwasz" due-to="Tyler Nighswander">Align DUMP archive block size with Linux `dump` utility.</action>
--- a/src/main/java/org/apache/commons/compress/compressors/bzip2/BZip2CompressorInputStream.java
+++ b/src/main/java/org/apache/commons/compress/compressors/bzip2/BZip2CompressorInputStream.java
@@ -42,7 +42,8 @@ import org.apache.commons.io.input.CloseShieldInputStream;
  */
 public class BZip2CompressorInputStream extends CompressorInputStream implements BZip2Constants, InputStreamStatistics {
 
-    private static final class Data {
+    // package private for testing
+    static final class Data {
 
         // (with blockSize 900k)
         final boolean[] inUse = new boolean[256]; // 256 byte
@@ -56,8 +57,10 @@ public class BZip2CompressorInputStream extends CompressorInputStream implements
          */
         final int[] unzftab = new int[256]; // 1024 byte
 
-        final int[][] limit = new int[N_GROUPS][MAX_ALPHA_SIZE]; // 6192 byte
-        final int[][] base = new int[N_GROUPS][MAX_ALPHA_SIZE]; // 6192 byte
+        // Needs indexes from 0 to MAX_CODE_LEN inclusive.
+        final int[][] limit = new int[N_GROUPS][MAX_CODE_LEN + 1];
+        // Needs indexes from 0 to MAX_CODE_LEN + 1 inclusive.
+        final int[][] base = new int[N_GROUPS][MAX_CODE_LEN + 2];
         final int[][] perm = new int[N_GROUPS][MAX_ALPHA_SIZE]; // 6192 byte
         final int[] minLens = new int[N_GROUPS]; // 24 byte
 
@@ -157,7 +160,7 @@ public class BZip2CompressorInputStream extends CompressorInputStream implements
      * Called by createHuffmanDecodingTables() exclusively.
      */
     private static void hbCreateDecodeTables(final int[] limit, final int[] base, final int[] perm, final char[] length, final int minLen, final int maxLen,
-            final int alphaSize) throws IOException {
+            final int alphaSize) {
         for (int i = minLen, pp = 0; i <= maxLen; i++) {
             for (int j = 0; j < alphaSize; j++) {
                 if (length[j] == i) {
@@ -171,7 +174,6 @@ public class BZip2CompressorInputStream extends CompressorInputStream implements
         }
         for (int i = 0; i < alphaSize; i++) {
             final int len = length[i] + 1;
-            checkBounds(len, MAX_ALPHA_SIZE, "length");
             base[len]++;
         }
         for (int i = 1, b = base[0]; i < MAX_CODE_LEN; i++) {
@@ -298,8 +300,7 @@ public class BZip2CompressorInputStream extends CompressorInputStream implements
     /**
      * Called by recvDecodingTables() exclusively.
      */
-    private void createHuffmanDecodingTables(final int alphaSize, final int nGroups) throws IOException {
-        final Data dataShadow = this.data;
+    static void createHuffmanDecodingTables(final int alphaSize, final int nGroups, final Data dataShadow) throws IOException {
         final char[][] len = dataShadow.temp_charArray2d;
         final int[] minLens = dataShadow.minLens;
         final int[][] limit = dataShadow.limit;
@@ -307,8 +308,8 @@ public class BZip2CompressorInputStream extends CompressorInputStream implements
         final int[][] perm = dataShadow.perm;
 
         for (int t = 0; t < nGroups; t++) {
-            int minLen = 32;
-            int maxLen = 0;
+            int minLen = MAX_CODE_LEN;
+            int maxLen = 1;
             final char[] len_t = len[t];
             for (int i = alphaSize; --i >= 0;) {
                 final char lent = len_t[i];
@@ -751,12 +752,17 @@ public class BZip2CompressorInputStream extends CompressorInputStream implements
                 while (bsGetBit(bin)) {
                     curr += bsGetBit(bin) ? -1 : 1;
                 }
+                // Same condition as in bzip2
+                if (curr < 1 || curr > MAX_CODE_LEN) {
+                    throw new CompressorException(
+                            "Corrupted input, code length value out of range [%d, %d]: %d", 1, MAX_CODE_LEN, curr);
+                }
                 len_t[i] = (char) curr;
             }
         }
 
         // finally create the Huffman tables
-        createHuffmanDecodingTables(alphaSize, nGroups);
+        createHuffmanDecodingTables(alphaSize, nGroups, dataShadow);
     }
 
     private int setupBlock() throws IOException {
--- a/src/test/java/org/apache/commons/compress/compressors/bzip2/BZip2CompressorInputStreamTest.java
+++ b/src/test/java/org/apache/commons/compress/compressors/bzip2/BZip2CompressorInputStreamTest.java
@@ -18,6 +18,7 @@
  */
 package org.apache.commons.compress.compressors.bzip2;
 
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
@@ -35,6 +36,7 @@ import org.apache.commons.compress.archivers.ArchiveException;
 import org.apache.commons.compress.archivers.ArchiveInputStream;
 import org.apache.commons.compress.archivers.ArchiveStreamFactory;
 import org.apache.commons.compress.compressors.CompressorException;
+import org.apache.commons.compress.compressors.bzip2.BZip2CompressorInputStream.Data;
 import org.apache.commons.io.IOUtils;
 import org.junit.jupiter.api.Test;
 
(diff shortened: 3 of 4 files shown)


src/test/java/org/apache/commons/compress/compressors/bzip2/BZip2CompressorInputStreamTest.java
```diff
@@ -149,4 +160,103 @@ void testSingleByteReadConsistentlyReturnsMinusOneAtEof() throws IOException {
...
+     *     <li>Selectors: all three encode j=1 (unary "10").</li>
+     *     <li>Huffman code lengths for 2 groups over alphabet size 3 (RUNA, RUNB, EOB) are all equal to {@code codeLength}.</li>
+     * </ul>
+     * <p>
+     *     <strong>Note:</strong> The values are chosen to keep everything byte-aligned.
+     * </p>
+     * @param codeLength the code length to use for each symbol in each group; must be in [0, 31]
+     */
+    private BitInputStream prepareDecodingTables(final int codeLength) {
+        assertTrue(0 <= codeLength && codeLength <= 31, "codeLength must be between 0 and 31");
```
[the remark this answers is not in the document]
ppkarwasz: This method is actually used to create both valid and **invalid** tables, so the limit between `0` and `31` is only due to the simplified implementation.

In reality we could have an implementation that generates code lengths of any length (both positive and negative), but `0` is enough to test for values below the allowed `[1, 20]` range and `21` is enough to test values above that range.

src/main/java/org/apache/commons/compress/compressors/bzip2/BZip2CompressorInputStream.java
```diff
@@ -298,17 +300,16 @@ private boolean complete() throws IOException {
...
         final char[][] len = dataShadow.temp_charArray2d;
         final int[] minLens = dataShadow.minLens;
         final int[][] limit = dataShadow.limit;
         final int[][] base = dataShadow.base;
         final int[][] perm = dataShadow.perm;
 
         for (int t = 0; t < nGroups; t++) {
-            int minLen = 32;
-            int maxLen = 0;
+            int minLen = MAX_CODE_LEN;
```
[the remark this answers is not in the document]
ppkarwasz: This suggestion doesn't make any difference: any value ≥ 20 would do fine.

However I see how this can hinder reasoning if we iterate over an empty set, so I initialized the values to the length of the first symbol in https://github.com/apache/commons-compress/pull/699/commits/7d1f6af54fd11a65f7a36f9602e843f7f8233d5c. There are at least 2 symbols at this point in the execution.

src/main/java/org/apache/commons/compress/compressors/bzip2/BZip2CompressorInputStream.java
```diff
@@ -171,7 +175,6 @@ private static void hbCreateDecodeTables(final int[] limit, final int[] base, fi
         }
         for (int i = 0; i < alphaSize; i++) {
             final int len = length[i] + 1;
-            checkBounds(len, MAX_ALPHA_SIZE, "length");
             base[len]++;
         }
         for (int i = 1, b = base[0]; i < MAX_CODE_LEN; i++) {
```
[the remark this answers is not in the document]
ppkarwasz: Nice catch! :100:

We actually need a bit more here, the loop should iterate through the full length of `base`. The reasoning is:

* At this stage we fill `base` with the cumulative counts Σ<sub>< len</sub> count\[len], where `count[len]` is the number of symbols of a given length. Later, when computing `limit[len]`, we need to access `count[len]` values all the way up to `count[MAX_CODE_LEN]`, which requires having the next `MAX_CODE_LEN + 1` index available. I commented the code in b3331775b813e0bf1e12e05f641c655928e31413 to make this more clear.
* In e217afe0ac33638caf7d866daecd0eb81953429f I replaced bound checks against theoretical maximums with checks against the actual array length.

src/test/java/org/apache/commons/compress/compressors/bzip2/BZip2CompressorInputStreamTest.java
```diff
@@ -149,4 +160,103 @@ void testSingleByteReadConsistentlyReturnsMinusOneAtEof() throws IOException {
...
+        try (BitInputStream tables = prepareDecodingTables(codeLength)) {
+            final Data data = new Data(1);
+
+            final CompressorException ex = assertThrows(
+                    CompressorException.class,
+                    () -> BZip2CompressorInputStream.recvDecodingTables(tables, data),
+                    "Expected CompressorException for invalid code length " + codeLength);
+
+            final String msg = ex.getMessage();
+            assertAll(
```
garydgregory: Please don't use `assertAll()`, it's just clutter IMO. The code is clearer and easier to maintain without using assertAll(). Should I add an assertion to an existing assertAll(), outside of it, in a separate assertAll()? It's crazy making IMO. We don't use assertAll() anywhere else.

Every single line of code is not expected to throw an exception; it's a given in a unit test. If you want to document something, use an inline `//` comment.

We do use `assertThrows()` when we _expect_ exceptions, of course. Everything else is expected not to throw.

src/main/java/org/apache/commons/compress/compressors/bzip2/BZip2CompressorInputStream.java
```diff
@@ -155,38 +159,46 @@ private static void checkBounds(final int checkVal, final int limitExclusive, fi
...
                 }
             }
         }
-        for (int i = MAX_CODE_LEN; --i > 0;) {
-            base[i] = 0;
-            limit[i] = 0;
-        }
+        // Ensure the arrays were not reused.
+        Arrays.fill(base, 0);
+        Arrays.fill(limit, 0);
```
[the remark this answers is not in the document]
ppkarwasz: That is actually useful, thanks!

src/test/java/org/apache/commons/compress/compressors/bzip2/BZip2CompressorInputStreamTest.java
```diff
@@ -149,4 +160,103 @@ void testSingleByteReadConsistentlyReturnsMinusOneAtEof() throws IOException {
...
+     *     <li>Selectors: all three encode j=1 (unary "10").</li>
+     *     <li>Huffman code lengths for 2 groups over alphabet size 3 (RUNA, RUNB, EOB) are all equal to {@code codeLength}.</li>
+     * </ul>
+     * <p>
+     *     <strong>Note:</strong> The values are chosen to keep everything byte-aligned.
+     * </p>
+     * @param codeLength the code length to use for each symbol in each group; must be in [0, 31]
+     */
+    private BitInputStream prepareDecodingTables(final int codeLength) {
+        assertTrue(0 <= codeLength && codeLength <= 31, "codeLength must be between 0 and 31");
```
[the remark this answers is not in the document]
ppkarwasz: See https://github.com/apache/commons-compress/pull/699#discussion_r2303281425

garydgregory (changes requested):
Hi @ppkarwasz 

Please see my comment.

src/main/java/org/apache/commons/compress/compressors/bzip2/BZip2CompressorInputStream.java
```diff
@@ -42,10 +42,12 @@
  */
 public class BZip2CompressorInputStream extends CompressorInputStream implements BZip2Constants, InputStreamStatistics {
 
-    private static final class Data {
+    // package private for testing
+    static final class Data {
 
         // (with blockSize 900k)
         final boolean[] inUse = new boolean[256]; // 256 byte
+        private int inUseCount;
```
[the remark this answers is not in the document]
ppkarwasz: Looks like a good clarification. I initialized the value to `2` (its minimal) value to improve code readability in https://github.com/apache/commons-compress/pull/699/commits/cd028373331626843741066dd63c723fdc7a4bd8.

src/test/java/org/apache/commons/compress/compressors/bzip2/BZip2CompressorInputStreamTest.java
ppkarwasz: Fixed in https://github.com/apache/commons-compress/pull/699/commits/3b02d35eb8a92db9882849f629fff6464170e730

ppkarwasz:
**Note**: I plan to extract the Huffman table build/decoding into a small reusable class. The fact we’re testing package-private helpers suggests the abstraction belongs elsewhere.

That refactor needs a bit more work: it only pays off if DEFLATE64 uses it too. Today `deflate64.HuffmanDecoder` mixes generic Huffman logic with DEFLATE64-specific handling, so I’ll address that in a follow-up.

src/changes/changes.xml
```diff
@@ -58,7 +58,7 @@ The <action> type attribute can be add,update,fix,remove.
       <action type="fix" dev="ggregory" due-to="Gary Gregory">Don't loose precision while reading folders from a SevenZFile.</action>
       <action type="fix" dev="ggregory" due-to="Roel van Dijk, Gary Gregory">Improve some exception messages in TarUtils and TarArchiveEntry.</action>
       <!-- FIX bzip2 -->      
-      <action type="fix" dev="ggregory" due-to="Tyler Nighswander, Gary Gregory">org.apache.commons.compress.compressors.bzip2.BZip2CompressorInputStream constructors now throws CompressorException instead of ArrayIndexOutOfBoundsException.</action>
+      <action type="fix" dev="pkarwasz" due-to="Tyler Nighswander, Piotr P. Karwasz, Gary Gregory">Enforce 20-bit Huffman code-length limit consistently in BZip2.</action>
```
garydgregory: Hello @ppkarwasz 
We lost some useful information in this edit. "Enforce" doesn't mean anything unless you mention the consequence, which was previously included and is still true after this PR: "org.apache.commons.compress.compressors.bzip2.BZip2CompressorInputStream constructors now throws CompressorException instead of ArrayIndexOutOfBoundsException"

I would leave the original `action` and add a new one, something like: "org.apache.commons.compress.compressors.bzip2.BZip2CompressorInputStream constructors throw a CompressorException when input exceeds the 20-bit Huffman code-length limit."

ppkarwasz: I think both actions are talking about the same problem so we should merge them.

What about: “Enforce 20-bit Huffman code-length limit in `BZip2CompressorInputStream` constructor“? All class names are unique, so I don't find it useful to provide the FQCN.

garydgregory: No, because the original description is for a broader set of changes where the exception is now thrown from more than once place; the new description is for a single case, the 20-bit limit.

ppkarwasz: Sorry, since those changes were spread across multiple commits, it’s a bit tricky to pin down exactly which ones you’re referring to.

The commit that originally introduced this changelog entry (e0a9a57d611df84f1fdacf917e28a97c8c33b17b) updated the bound check here:

https://github.com/apache/commons-compress/blob/7a0ae07dd139175b3ed8a58d370cb80f882066cf/src/main/java/org/apache/commons/compress/compressors/bzip2/BZip2CompressorInputStream.java#L171-L175

so that it correctly validated the index before dereferencing:

https://github.com/apache/commons-compress/blob/e0a9a57d611df84f1fdacf917e28a97c8c33b17b/src/main/java/org/apache/commons/compress/compressors/bzip2/BZip2CompressorInputStream.java#L172-L176

That change effectively turned an `ArrayIndexOutOfBoundsException` into a `CompressorException`.

In this PR, however, that particular bound check is removed, since the caller already ensures that `length` is in the `[0, MAX_CODE_LEN]` range:

https://github.com/apache/commons-compress/blob/3b02d35eb8a92db9882849f629fff6464170e730/src/main/java/org/apache/commons/compress/compressors/bzip2/BZip2CompressorInputStream.java#L180-L183

As far as I can tell, that was the only array access with an incorrect bound check, so the only realistic source of `ArrayIndexOutOfBoundsException`. For that reason, I don’t think it’s worth explicitly mentioning AOOB in the changelog.

That said, the same commit (e0a9a57d611df84f1fdacf917e28a97c8c33b17b) also switched the type of exception thrown by bound checks in general from `IOException` to `CompressorException`. To reflect that more accurately, I’d suggest replacing the single changelog entry with two clearer ones:

* BZip2 input streams now throw `CompressorException` (instead of `IOException`) for invalid or corrupted data.
* BZip2 input streams treat Huffman codes longer than 20 bits as corrupted data, matching the behavior of the reference implementation.

src/changes/changes.xml
```diff
@@ -58,8 +58,8 @@ The <action> type attribute can be add,update,fix,remove.
...
       <action type="fix" dev="ggregory" due-to="Roel van Dijk, Gary Gregory">Improve some exception messages in TarUtils and TarArchiveEntry.</action>
       <!-- FIX bzip2 -->      
-      <action type="fix" dev="ggregory" due-to="Tyler Nighswander, Gary Gregory">org.apache.commons.compress.compressors.bzip2.BZip2CompressorInputStream constructors now throws CompressorException instead of ArrayIndexOutOfBoundsException.</action>
-      <action type="fix" dev="ggregory" due-to="Gary Gregory">org.apache.commons.compress.compressors.bzip2.BZip2CompressorInputStream throws CompressorException (an IOException subclass) instead of IOException when it encounters formatting problems.</action>
+      <action type="fix" dev="ggregory" due-to="Tyler Nighswander, Gary Gregory">BZip2 input streams now throw CompressorException (a subclass of IOException) for invalid or corrupted data, providing more specific error reporting.</action>
+      <action type="fix" dev="pkarwasz" due-to="Tyler Nighswander, Piotr P. Karwasz">BZip2 input streams treat Huffman codes longer than 20 bits as corrupted data, matching the behavior of the reference implementation.</action>
```
ppkarwasz, line 62: @garydgregory,

I consolidated the BZip2 changelog entries into these two, which I think strike a good balance between clarity and user relevance across the 5 commits and this PR:

* They explain that BZip2 streams now distinguish between I/O errors (`IOException`) and content-related errors (`CompressorException`), which helps users decide whether to retry or report a clearer error.
* They highlight the Huffman code length restriction, which may matter for anyone relying on custom-encoded data.

I didn’t include the removal of unchecked exceptions (e.g., `ArrayIndexOutOfBoundsException`), since those are more indicative of internal bugs than user-facing changes. Do you think it’s worth adding an explicit note like:

> “Fixed `ArrayIndexOutOfBoundsException` for archives with non-compliant Huffman codes.”

or should we keep that detail implicit in the general error handling change?

src/changes/changes.xml
ppkarwasz: Let's continue the discussion in this comment: https://github.com/apache/commons-compress/pull/699#discussion_r2304735466

PR #701, opened 2025-08-29
Add reusable Huffman decoder

ppkarwasz:
This PR adds a reusable Huffman decoder for canonical Huffman codes.
The decoder can be shared across multiple algorithms, including DEFLATE, DEFLATE64, and BZip2.

By centralizing Huffman code handling, the implementation reduces duplication and makes compressor code easier to maintain and extend.

**Work in progress:**

* [x] Refactor BZip2 implementation
* [x] Refactor DEFLATE64 implementation
* [ ] (optional) Refactor DEFLATE implementation: currently relies on `InflaterInputStream` from the JDK


garydgregory:
Hello @ppkarwasz 

Are the implementations in DEFLATE, DEFLATE64, and BZip2 identical? Do we know the specs for these formats well enough to have confidence that we won't end up back where we started, accommodating tweaks, features, or oddities that are format-specific? We'll only know some of this as this refactoring proceeds. 

I don't think we need a new package called "alg", especially with a single class in it. If this new class is designed for reuse among compressors, then is can go in "compressors". Or "hufman" if we end up with more classes.

ppkarwasz:
> Are the implementations in DEFLATE, DEFLATE64, and BZip2 identical? Do we know the specs for these formats well enough to have confidence that we won't end up back where we started, accommodating tweaks, features, or oddities that are format-specific? We'll only know some of this as this refactoring proceeds.

They aren’t identical end-to-end, but they share the same **canonical Huffman decoding primitive**: given an array of code lengths, map an LSB/MSB bitstream to symbols. This PR extracts just that primitive into a reusable `HuffmanDecoder`. Format-specific parts remain outside (block structure, alphabets, extra bits, selectors, etc.).

I have also refactored the DEFLATE64 implementation, so this PR is ready for a review.

> I don't think we need a new package called "alg", especially with a single class in it. If this new class is designed for reuse among compressors, then is can go in "compressors". Or "hufman" if we end up with more classes.

Agreed. I moved the class to `compressors.support` for now. If we don’t want it as public API, I can relocate to an `internal` package (and mark as non-API in module/OSGi metadata).

src/main/java/org/apache/commons/compress/compressors/support/HuffmanDecoder.java
```diff
@@ -0,0 +1,268 @@
...
+        return bit;
+    }
+
+    private static int readBitsFully(final BitInputStream in, final int numBits) throws IOException {
+        final int code = (int) in.readBits(numBits);
+        if (code < 0) {
+            throw new EOFException("Truncated Huffman bitstream");
+        }
+        // Adjust for bit order
+        return in.getByteOrder() == ByteOrder.BIG_ENDIAN ? code : Integer.reverse(code) >>> (32 - numBits);
```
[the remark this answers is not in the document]
ppkarwasz: `in` is a parameter passed by the caller, so it can not be cached.

src/main/java/org/apache/commons/compress/compressors/support/package-info.java
```diff
@@ -0,0 +1,23 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+/**
+ * Algorithms reused by various compressors.
+ * @since 1.29.0
+ */
+package org.apache.commons.compress.compressors.support;
```
garydgregory, line 23: Adding a package is unnecessary; we already have common utilities in `org.apache.commons.compress.compressors`. I can pull this up post-merge.

ppkarwasz: Having a separate package for components of a compression/decompression pipeline seems cleaner to me.
Granted: we need to have more than one class to do that.

Since this PR doesn't fix any bug, we can delay this (and #690) to version `1.30.0`, so we can have time to find additional compression pipeline elements to refactor. What do you think?

garydgregory:
Hi @ppkarwasz 
I'm OK with waiting on this one. Are you thinking that #690 should come in first? For 1.29.0 or 1.30.0?

ppkarwasz:
Hi @garydgregory,

Yes, I believe #690 is more valuable and should come in first. It gives us a good opportunity to restructure the compressor code into more reusable and maintainable components. Right now much of the compressor logic is still a direct Java translation of defensive C code, which I don’t fully understand myself. That has led to duplication across classes implementing the same algorithms with C-style optimizations that don’t fit well in Java.

LHA introduces four additional compression methods, some of which may already overlap with what we have (e.g. in SevenZ). But I don’t currently have the time to review it as thoroughly as it deserves.

So my proposal would be:

* For **1.29.0**: complete the cycle of archiver improvements we’ve been working on in recent months.
* For **1.30.0**: revisit the compression algorithms to identify reusable patterns, reduce duplication, and add LHA support.

@fkjellberg: does this plan sound reasonable to you? You have much deeper expertise in compression than we do, so your perspective would be really valuable here. And thank you for your patience. I know your PR has been waiting for review longer than it should, and we’ve been tied up finishing the work from our previous roadmap. I really appreciate the time and effort you’ve already put into it.

fkjellberg:
Hi @ppkarwasz ,

The plan to postpone this and my LHA PR until 1.30 sounds good to me.

Like you, I also noticed that there are a lot of overlapping code in different compressors that can be refactored for better reuse and to make the code easier to understand. My plan was to create some refactoring PRs once my LHA implementation was merged because I wanted to avoid doing too much refactoring of existing code as part of my PR.

The LHA compressor is very similar to the `ExplodingInputStream` in the zip package and I reused the `BinaryTree` and `CircularBuffer` from that package and more refactoring could possibly be done to align both implementations even more.

The deflate64 that you started refactoring in this PR is also a good candidate for further refactoring, e.g. the `HuffmanDecoder.DecodingMemory` could possibly be replaced with `CircularBuffer` instead.

I think this PR is a good first step in the refactoring process. Let's continue working on refactoring after the 1.29 release.

garydgregory:
@ppkarwasz 
PR https://github.com/apache/commons-compress/pull/792 is a new version of this PR without conflicts.

garydgregory:
Closing, merged https://github.com/apache/commons-compress/pull/792 instead.

PR #702, opened 2025-08-30
extract duplicate code 

Raeps:
_growAtEnd_
_growAtFront_
_growForInsert_
These three methods share exactly the same logic to calculate  _increment_

extracting the logic into _calculateIncrement_ removes duplicate.

### DIFF
--- a/src/main/java/org/apache/commons/compress/harmony/pack200/IntList.java
+++ b/src/main/java/org/apache/commons/compress/harmony/pack200/IntList.java
@@ -17,7 +17,7 @@
  * under the License.
  */
 package org.apache.commons.compress.harmony.pack200;
-
+import org.apache.commons.lang3.math.NumberUtils;
 import java.util.Arrays;
 
 /**
@@ -122,7 +122,7 @@ public class IntList {
 
 
 
-    private int calculateIncrement(final int size, final int required) {
+    private static int calculateIncrement(final int size, final int required) {
         return NumberUtils.max(size / 2, required, 12);
     }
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/commons/compress/harmony/pack200/IntList.java
```diff
@@ -120,6 +120,19 @@ public int get(final int location) {
...
+
+    private int calculateIncrement(final int size, final int required) {
+        int increment = size / 2;
+        if (required > increment) {
+            increment = required;
+        }
+        if (increment < 12) {
+            increment = 12;
+        }
+        return increment;
```
ppkarwasz: You can use `NumberUtils` for this:
```suggestion
        return NumberUtils.max(size / 2, required, 12);
```

Additionally I would:

- Make the method `static`.
- Optionally extract the constant 12 to a `MIN_INCREMENT` constant, although for this single usage, it is not really necessary.

ppkarwasz (changes requested):
Hi @Raeps,

Nice catch! :100:

src/main/java/org/apache/commons/compress/harmony/pack200/IntList.java
Raeps: yeah 
the use of NumberUtils would be more clear for this case

src/main/java/org/apache/commons/compress/harmony/pack200/IntList.java
```diff
@@ -120,6 +122,12 @@ public int get(final int location) {
         throw new IndexOutOfBoundsException("" + location);
     }
 
+
+
```
garydgregory: You only need one blank line to separate methods, not 3.

Raeps: You're right, I omitted the codestyle, now it's fixed.

ppkarwasz: The `checkstyle:check` execution did not catch this, so it is on us! :wink:

PR #703, opened 2025-09-03
[COMPRESS-708] Fix ZstdCompressorInputStream closing underlying stream

YuWeiBoy:
I fixed the issue where ZstdCompressorInputStream closes the underlying stream.

1. The bug occurred because the ZstdInputStream used in ZstdCompressorInputStream would close the underlying InputStream during GC, which in turn caused errors like "stream closed" for other wrapper streams (such as ZipArchiveInputStream) that depend on this underlying stream.
2.  The principle to fix this bug is to wrap the InputStream passed to ZstdInputStream with CloseShieldInputStream.
3. I wrote the testZstdInputStreamErrorCloseWhenGc method for testing, and in this method, I called System.gc(); to quickly trigger Java's GC to reproduce the bug.


garydgregory:
Tracking https://issues.apache.org/jira/browse/COMPRESS-708

garydgregory:
Closing: See my comment in https://issues.apache.org/jira/browse/COMPRESS-708

PR #706, opened 2025-09-12
feat: Introduce builders for all ArchiveInputStream implementations

ppkarwasz:
This PR introduces builders for all `ArchiveInputStream` implementations, starting from a new generic builder that captures options shared across implementations. Existing code has been refactored to use the new builders instead of direct constructor calls, improving consistency, readability, and maintainability.

## Changes

* **Generic builder** for `ArchiveInputStream` to handle shared options.
* **New builders** for specific implementations:

  * `ArArchiveInputStream`
  * `ArjArchiveInputStream`
  * `CpioArchiveInputStream`
  * `DumpArchiveInputStream`
  * `JarArchiveInputStream`
  * `ZipArchiveInputStream`

## Motivation

* Simplifies client code by replacing complex or repetitive constructor calls.
* Provides a unified and extensible API for configuring archive input streams.
* Improves readability and makes future changes to constructors less disruptive.

### DIFF
--- a/src/main/java/org/apache/commons/compress/archivers/ArchiveInputStream.java
+++ b/src/main/java/org/apache/commons/compress/archivers/ArchiveInputStream.java
@@ -99,6 +99,9 @@ public abstract class ArchiveInputStream<E extends ArchiveEntry> extends FilterI
     public abstract static class AbstractBuilder<T extends ArchiveInputStream<?>, B extends AbstractBuilder<T, B>>
             extends AbstractStreamBuilder<T, B> {
 
+        /**
+         * Constructs a new instance.
+         */
         protected AbstractBuilder() {
             // empty
         }
--- a/src/main/java/org/apache/commons/compress/archivers/ArchiveInputStream.java
+++ b/src/main/java/org/apache/commons/compress/archivers/ArchiveInputStream.java
@@ -26,6 +26,7 @@ import java.util.Iterator;
 import java.util.Objects;
 
 import org.apache.commons.io.Charsets;
+import org.apache.commons.io.build.AbstractStreamBuilder;
 import org.apache.commons.io.function.IOConsumer;
 import org.apache.commons.io.function.IOIterator;
 import org.apache.commons.io.input.NullInputStream;
@@ -86,6 +87,20 @@ public abstract class ArchiveInputStream<E extends ArchiveEntry> extends FilterI
             return null;
         }
 
+    }
+
+    /**
+     * Generic builder for ArchiveInputStream instances.
+     *
+     * @param <T> The type of {@link ArchiveInputStream} to build.
+     * @param <B> The type of the concrete Builder.
+     * @since 1.29.0
+     */
+    public abstract static class Builder<T extends ArchiveInputStream<?>, B extends Builder<T, B>> extends AbstractStreamBuilder<T, B> {
+
+        protected Builder() {
+            // empty
+        }
     }
 
     private static final int BYTE_MASK = 0xFF;
@@ -294,4 +309,5 @@ public abstract class ArchiveInputStream<E extends ArchiveEntry> extends FilterI
     public synchronized void reset() throws IOException {
         // noop
     }
+
 }
--- a/src/main/java/org/apache/commons/compress/archivers/ArchiveInputStream.java
+++ b/src/main/java/org/apache/commons/compress/archivers/ArchiveInputStream.java
@@ -93,12 +93,13 @@ public abstract class ArchiveInputStream<E extends ArchiveEntry> extends FilterI
      * Generic builder for ArchiveInputStream instances.
      *
      * @param <T> The type of {@link ArchiveInputStream} to build.
-     * @param <B> The type of the concrete Builder.
+     * @param <B> The type of the concrete AbstractBuilder.
      * @since 1.29.0
      */
-    public abstract static class Builder<T extends ArchiveInputStream<?>, B extends Builder<T, B>> extends AbstractStreamBuilder<T, B> {
+    public abstract static class AbstractBuilder<T extends ArchiveInputStream<?>, B extends AbstractBuilder<T, B>>
+            extends AbstractStreamBuilder<T, B> {
 
-        protected Builder() {
+        protected AbstractBuilder() {
             // empty
         }
     }
--- a/src/main/java/org/apache/commons/compress/archivers/ar/ArArchiveInputStream.java
+++ b/src/main/java/org/apache/commons/compress/archivers/ar/ArArchiveInputStream.java
@@ -39,6 +39,30 @@ import org.apache.commons.lang3.ArrayUtils;
  */
 public class ArArchiveInputStream extends ArchiveInputStream<ArArchiveEntry> {
 
+    /**
+     * Builds a new {@link ArArchiveInputStream}.
+     * <p>
+     *     For example:
+     * </p>
+     * <pre>{@code
+     * ArArchiveInputStream in = ArArchiveInputStream.builder()
+     *     .setPath(inputPath)
+     *     .get();
+     * }</pre>
+     * @since 1.29.0
+     */
+    public static final class Builder extends ArchiveInputStream.Builder<ArArchiveInputStream, Builder> {
+
+        private Builder() {
+            setCharset(StandardCharsets.US_ASCII);
+        }
+
+        @Override
+        public ArArchiveInputStream get() throws IOException {
+            return new ArArchiveInputStream(this);
+        }
+    }
+
     // offsets and length of meta data parts
     private static final int NAME_OFFSET = 0;
     private static final int NAME_LEN = 16;
@@ -113,6 +137,16 @@ public class ArArchiveInputStream extends ArchiveInputStream<ArArchiveEntry> {
         return ArrayUtils.startsWith(buffer, ArArchiveEntry.HEADER_BYTES);
     }
 
+    /**
+     * Creates a new builder.
+     *
+     * @return A new builder
+     * @since 1.29.0
+     */
+    public static Builder builder() {
+        return new Builder();
+    }
+
     private boolean closed;
 
     /*
@@ -137,7 +171,15 @@ public class ArArchiveInputStream extends ArchiveInputStream<ArArchiveEntry> {
      * @param inputStream the ar input stream
      */
     public ArArchiveInputStream(final InputStream inputStream) {
-        super(inputStream, StandardCharsets.US_ASCII.name());
+        this(inputStream, builder());
+    }
+
+    private ArArchiveInputStream(final Builder builder) throws IOException {
+        this(builder.getInputStream(), builder);
+    }
+
+    private ArArchiveInputStream(final InputStream inputStream, final Builder builder) {
+        super(inputStream, builder.getCharset());
     }
 
     private int asInt(final byte[] byteArray, final int offset, final int len, final boolean treatBlankAsZero) throws IOException {
--- a/src/main/java/org/apache/commons/compress/archivers/ar/ArArchiveInputStream.java
+++ b/src/main/java/org/apache/commons/compress/archivers/ar/ArArchiveInputStream.java
@@ -42,7 +42,7 @@ public class ArArchiveInputStream extends ArchiveInputStream<ArArchiveEntry> {
     /**
      * Builds a new {@link ArArchiveInputStream}.
      * <p>
-     *     For example:
+     * For example:
      * </p>
      * <pre>{@code
      * ArArchiveInputStream in = ArArchiveInputStream.builder()
--- a/src/main/java/org/apache/commons/compress/archivers/ar/ArArchiveInputStream.java
+++ b/src/main/java/org/apache/commons/compress/archivers/ar/ArArchiveInputStream.java
@@ -51,7 +51,7 @@ public class ArArchiveInputStream extends ArchiveInputStream<ArArchiveEntry> {
      * }</pre>
      * @since 1.29.0
      */
-    public static final class Builder extends ArchiveInputStream.Builder<ArArchiveInputStream, Builder> {
+    public static final class Builder extends AbstractBuilder<ArArchiveInputStream, Builder> {
 
         private Builder() {
             setCharset(StandardCharsets.US_ASCII);
--- a/src/main/java/org/apache/commons/compress/archivers/zip/ZipArchiveInputStream.java
+++ b/src/main/java/org/apache/commons/compress/archivers/zip/ZipArchiveInputStream.java
@@ -478,6 +478,13 @@ public class ZipArchiveInputStream extends ArchiveInputStream<ZipArchiveEntry> i
                         .setSkipSplitSignature(skipSplitSignature));
     }
 
+    /**
+     * Creates an instance from the given builder.
+     *
+     * @param builder The builder used to configure and create the stream.
+     * @throws IOException If the builder fails to create the underlying {@link InputStream}.
+     * @since 1.29.0
+     */
     protected ZipArchiveInputStream(final AbstractBuilder<?, ?> builder) throws IOException {
         this(builder.getInputStream(), builder);
     }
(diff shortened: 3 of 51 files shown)


src/main/java/org/apache/commons/compress/archivers/ArchiveInputStream.java
```diff
@@ -88,6 +89,20 @@ public Iterator<E> unwrap() {
...
     }
 
+    /**
+     * Generic builder for ArchiveInputStream instances.
+     *
+     * @param <T> The type of {@link ArchiveInputStream} to build.
+     * @param <B> The type of the concrete Builder.
+     * @since 1.29.0
+     */
+    public abstract static class Builder<T extends ArchiveInputStream<?>, B extends Builder<T, B>> extends AbstractStreamBuilder<T, B> {
```
garydgregory: If a Builder class is abstract, then please prefix its name with "Abstract" (like its superclass in this case).

src/main/java/org/apache/commons/compress/archivers/ArchiveInputStream.java
```diff
@@ -88,6 +89,20 @@ public Iterator<E> unwrap() {
...
+
+        protected Builder() {
```
garydgregory: New protected methods should also have a Javadoc, in this case: "Constructs a new instance."

src/main/java/org/apache/commons/compress/archivers/ar/ArArchiveInputStream.java
```diff
@@ -39,6 +39,30 @@
  */
 public class ArArchiveInputStream extends ArchiveInputStream<ArArchiveEntry> {
 
+    /**
+     * Builds a new {@link ArArchiveInputStream}.
+     * <p>
+     *     For example:
```
garydgregory: There is no need to indent text in paragraph blocks, especially since these spaces will be stripped out.

src/main/java/org/apache/commons/compress/archivers/zip/ZipArchiveInputStream.java
```diff
@@ -77,6 +77,112 @@
...
+         * <p>See <a href="https://issues.apache.org/jira/browse/COMPRESS-555">COMPRESS-555</a>
+         * for details.</p>
+         *
+         * <p>This feature is disabled by default.</p>
+         *
+         * @param allowStoredEntriesWithDataDescriptor {@code true} to read STORED entries with data descriptors,
+         *                                             {@code false} to stop at the compressed size.
+         * @return this
+         */
+
```
garydgregory: Remove extra blank line.

src/main/java/org/apache/commons/compress/archivers/zip/ZipArchiveInputStream.java
```diff
@@ -77,6 +77,112 @@
...
+         * issues in certain cases, such as JARs embedded in WAR files.</p>
+         *
+         * <p>See <a href="https://issues.apache.org/jira/browse/COMPRESS-555">COMPRESS-555</a>
+         * for details.</p>
+         *
+         * <p>This feature is disabled by default.</p>
+         *
+         * @param allowStoredEntriesWithDataDescriptor {@code true} to read STORED entries with data descriptors,
+         *                                             {@code false} to stop at the compressed size.
+         * @return this
```
garydgregory: Other builders we have are less terse for these types of setters: `@return {@code this} instance.`

src/main/java/org/apache/commons/compress/archivers/zip/ZipArchiveInputStream.java
```diff
@@ -77,6 +77,112 @@
...
+         */
+
+        public B setAllowStoredEntriesWithDataDescriptor(final boolean allowStoredEntriesWithDataDescriptor) {
+            this.allowStoredEntriesWithDataDescriptor = allowStoredEntriesWithDataDescriptor;
+            return asThis();
+        }
+
+        protected boolean isAllowStoredEntriesWithDataDescriptor() {
```
garydgregory: Javadoc

src/main/java/org/apache/commons/compress/archivers/zip/ZipArchiveInputStream.java
```diff
@@ -77,6 +77,112 @@
...
+            return allowStoredEntriesWithDataDescriptor;
+        }
+
+        /**
+         * Configures whether the stream should skip the ZIP split signature
```
garydgregory: The convention I've been using is that a setter method "Sets...:", a getter method "Gets...", and an "isFoo()" method "Tests..." or "Tests whether...".

src/main/java/org/apache/commons/compress/archivers/zip/ZipArchiveInputStream.java
```diff
@@ -77,6 +77,112 @@
...
+         * declared compressed size. Instead, it will continue until a data descriptor
+         * is encountered (by detecting the Data Descriptor Signature). This may cause
+         * issues in certain cases, such as JARs embedded in WAR files.</p>
+         *
+         * <p>See <a href="https://issues.apache.org/jira/browse/COMPRESS-555">COMPRESS-555</a>
+         * for details.</p>
+         *
+         * <p>This feature is disabled by default.</p>
+         *
+         * @param allowStoredEntriesWithDataDescriptor {@code true} to read STORED entries with data descriptors,
```
garydgregory: The "allow" prefix seems redundant to me and "with" is kinda wierd (to me). How about "storedDescribedEntries"?

src/main/java/org/apache/commons/compress/archivers/zip/ZipArchiveInputStream.java
```diff
@@ -77,6 +77,112 @@
...
+         *
+         * @param skipSplitSig {@code true} to skip the ZIP split signature, {@code false} otherwise
+         * @return this
+         */
+        public B setSkipSplitSig(final boolean skipSplitSig) {
+            this.skipSplitSig = skipSplitSig;
+            return asThis();
+        }
+
+        protected boolean isSkipSplitSig() {
```
garydgregory: Using an abbreviation like "Sig" could be confusing; I'd just say "Signtature", assuming that is what we're talking about here. Also: Javadoc which would remove any confusion at all.

src/main/java/org/apache/commons/compress/archivers/zip/ZipArchiveInputStream.java
```diff
@@ -320,14 +445,32 @@ public ZipArchiveInputStream(final InputStream inputStream, final String encodin
...
+        this(
+                inputStream,
+                builder()
+                        .setCharset(encoding)
+                        .setUseUnicodeExtraFields(useUnicodeExtraFields)
+                        .setAllowStoredEntriesWithDataDescriptor(allowStoredEntriesWithDataDescriptor)
+                        .setSkipSplitSig(skipSplitSig));
+    }
+
+    protected ZipArchiveInputStream(final AbstractBuilder<?, ?> builder) throws IOException {
```
garydgregory, line 488: Javadoc.

src/main/java/org/apache/commons/compress/archivers/zip/ZipArchiveInputStream.java
```diff
@@ -77,6 +77,112 @@
...
+         *
+         * @param useUnicodeExtraFields If {@code true} Unicode Extra Fields should be used.
+         * @return this
+         */
+        public B setUseUnicodeExtraFields(final boolean useUnicodeExtraFields) {
+            this.useUnicodeExtraFields = useUnicodeExtraFields;
+            return asThis();
+        }
+
+        protected boolean isUseUnicodeExtraFields() {
```
garydgregory, line 118: Javadoc.

src/main/java/org/apache/commons/compress/archivers/zip/ZipArchiveInputStream.java
```diff
@@ -77,6 +77,112 @@
...
+     * @param <B> The type of the builder itself.
+     * @since 1.29.0
+     */
+    public abstract static class AbstractBuilder<T extends ZipArchiveInputStream, B extends AbstractBuilder<T, B>>
+            extends ArchiveInputStream.Builder<T, B> {
+        private boolean useUnicodeExtraFields = true;
+        private boolean allowStoredEntriesWithDataDescriptor;
+        private boolean skipSplitSig;
+
+        protected AbstractBuilder() {
```
garydgregory, line 96: Javadoc.

src/main/java/org/apache/commons/compress/archivers/zip/ZipArchiveInputStream.java
```diff
@@ -77,6 +77,112 @@
...
+            setCharset(StandardCharsets.UTF_8);
+        }
+
+        /**
+         * Controls whether to use InfoZIP Unicode Extra Fields (if present) to set the file names.
```
garydgregory: Javadoc: "Sets..."

garydgregory (changes requested):
Hi @ppkarwasz 
Thank you for the (giant!) PR.

I've left comments throughout the PR, most apply to more than one location.

src/main/java/org/apache/commons/compress/archivers/ArchiveInputStream.java
ppkarwasz: Fixed in https://github.com/apache/commons-compress/pull/706/commits/dfbc2f0132cffc0c66857dae23c5b895af6d10fc

src/main/java/org/apache/commons/compress/archivers/ArchiveInputStream.java
ppkarwasz: Fixed in https://github.com/apache/commons-compress/pull/706/commits/8cf58ad7cb397857b5fc9d5fd172af692e1ce7c4

src/main/java/org/apache/commons/compress/archivers/ar/ArArchiveInputStream.java
ppkarwasz: How about introducing a shared Javadoc formatter in `commons-parent`? The indentation here came from IntelliJ IDEA, and since Javadoc is essentially HTML, I don’t have a strong preference as long as it’s easy to stay consistent.

Since each IDE formats differently and whitespace debates aren’t very productive, could you export your formatting configuration and set up a Spotless task in `commons-parent`? That way we can all apply the same rules from the CLI, without needing to change editors.

src/main/java/org/apache/commons/compress/archivers/zip/ZipArchiveInputStream.java
ppkarwasz: I fixed it in https://github.com/apache/commons-compress/pull/706/commits/8cf58ad7cb397857b5fc9d5fd172af692e1ce7c4.

As with the comment above, this happened because I try to preserve the original formatting of the file instead of rewrapping everything with my editor’s formatter. Here I reformatted only the Javadoc, but not the method that followed, so the extra blank line wasn’t collapsed.

src/main/java/org/apache/commons/compress/archivers/zip/ZipArchiveInputStream.java
ppkarwasz: Fixed in https://github.com/apache/commons-compress/pull/706/commits/8cf58ad7cb397857b5fc9d5fd172af692e1ce7c4

src/main/java/org/apache/commons/compress/archivers/zip/ZipArchiveInputStream.java
ppkarwasz: Fixed in https://github.com/apache/commons-compress/pull/706/commits/8cf58ad7cb397857b5fc9d5fd172af692e1ce7c4

src/main/java/org/apache/commons/compress/archivers/zip/ZipArchiveInputStream.java
ppkarwasz: Fixed in https://github.com/apache/commons-compress/pull/706/commits/8cf58ad7cb397857b5fc9d5fd172af692e1ce7c4

src/main/java/org/apache/commons/compress/archivers/zip/ZipArchiveInputStream.java
ppkarwasz: Fixed in https://github.com/apache/commons-compress/pull/706/commits/e6d2e40fd8b570d15c380c33f99f5e5ff9e72eb3

src/main/java/org/apache/commons/compress/archivers/zip/ZipArchiveInputStream.java
ppkarwasz: Fixed in https://github.com/apache/commons-compress/pull/706/commits/8cf58ad7cb397857b5fc9d5fd172af692e1ce7c4

src/main/java/org/apache/commons/compress/archivers/zip/ZipArchiveInputStream.java
ppkarwasz: Fixed in https://github.com/apache/commons-compress/pull/706/commits/8cf58ad7cb397857b5fc9d5fd172af692e1ce7c4

src/main/java/org/apache/commons/compress/archivers/zip/ZipArchiveInputStream.java
ppkarwasz: Fixed in https://github.com/apache/commons-compress/pull/706/commits/8cf58ad7cb397857b5fc9d5fd172af692e1ce7c4

src/main/java/org/apache/commons/compress/archivers/zip/ZipArchiveInputStream.java
ppkarwasz: I fixed the Javadoc in https://github.com/apache/commons-compress/pull/706/commits/8cf58ad7cb397857b5fc9d5fd172af692e1ce7c4 and renamed the method (and all associated private fields) to `isSkipSplitSignature` in https://github.com/apache/commons-compress/pull/706/commits/bb742953c40f2efb8306a8c98dc6747edb9c7b84

I left the precise Javadoc on the builder setters, since those are `public` and will be used by users. These getters were introduced to be reused by implementors of derived classes.

src/main/java/org/apache/commons/compress/archivers/zip/ZipArchiveInputStream.java
ppkarwasz: I renamed this to `supportStoredEntryDataDescriptor` in https://github.com/apache/commons-compress/pull/706/commits/bb742953c40f2efb8306a8c98dc6747edb9c7b84.

I agree that `with` wasn’t helpful. I hesitated on `storedDescribedEntry`, though: to me it suggests “an entry that is STORED and described somehow,” whereas the intent is specifically about entries that include a trailing data descriptor.

garydgregory:
I now realize (post-merge) that the refactoring of the tests was a bit too aggressive because we now have _many_ constructors that are no longer tested 😞

ppkarwasz:
BTW: the tests were intentionally rewritten to use the new builder pattern rather than the legacy constructors. The idea was to hold our own code to the same standard that many of our users apply to theirs: some projects now **fail** the build if deprecated APIs are used. By doing this, we make sure every deprecation is **deliberate** and **considered**.

However, personally, I wouldn’t deprecate any constructors in `1.29.0`. My preference would be to first announce the plan in the release notes, then introduce the deprecations in `1.30.0`.

To smooth the migration, we could also add Error Prone’s `@InlineMe` annotations. These are already supported by tools like Refaster and OpenRewrite, so users would be able to automatically rewrite call sites away from deprecated methods.

What do you think about this approach?

ppkarwasz:
I added the missing tests in #708.

> I don't see any point in delaying deprecations. In general, when we introduce a new API that we know replaces an old way of doing things, we deprecate the old APIs. It feels easier to understand this way, in with the new, out with the old.

After digging some more, this is also how the `@InlineMe` annotation should be used: Error Prone gives a warning if we add it to a non deprecated method. So I am convinced: let's deprecated the methods in `1.29.0`, however we should think of some way to allow users to remove deprecations automatically in the future.

PR #709, opened 2025-09-23
Add `ArchiveFile` abstraction for file-based archives

ppkarwasz:
This PR adds a new `ArchiveFile` interface to unify the handling of file-based archive utilities such as `SevenZFile`, `TarFile`, and `ZipFile`.

Although these classes target different archive formats, they share several core characteristics:

* All are `Closeable`.
* Each provides the same method to open an `InputStream` for a given entry (`InputStream getInputStream(T)` where `T extends ArchiveEntry`).
* Historically, their `getEntries()` methods returned incompatible types. This PR introduces a common `List<? extends T> entries()` method, aligning with `java.util.zip.ZipFile` in name but offering a modern return value.
* The `ZipFile#stream()` method (added in 1.28.0) is now part of this abstraction.

### DIFF
--- a/src/main/java/org/apache/commons/compress/archivers/ArchiveFile.java
+++ b/src/main/java/org/apache/commons/compress/archivers/ArchiveFile.java
@@ -87,9 +87,7 @@ public interface ArchiveFile<T extends ArchiveEntry> extends Closeable, IOIterab
 
     @Override
     default Iterable<T> unwrap() {
-        // Commons IO 2.21.0:
-        // return asIterable();
-        return null;
+        return asIterable();
     }
 }
 


src/test/java/org/apache/commons/compress/archivers/tar/TarFileTest.java
```diff
@@ -40,15 +40,14 @@
 import java.util.List;
 import java.util.zip.GZIPInputStream;
 
-import org.apache.commons.compress.AbstractTest;
+import org.apache.commons.compress.archivers.AbstractArchiveFileTest;
 import org.apache.commons.compress.archivers.ArchiveException;
 import org.apache.commons.io.IOUtils;
+import org.apache.commons.lang3.StringUtils;
```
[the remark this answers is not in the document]
ppkarwasz: The change is intentional, using the shaded version wasn't.

src/main/java/org/apache/commons/compress/archivers/tar/TarFile.java
```diff
@@ -240,7 +238,11 @@ private TarFile(final Builder builder) throws IOException {
         this.recordBuffer = ByteBuffer.allocate(this.recordSize);
         this.blockSize = builder.getBlockSize();
         this.lenient = builder.isLenient();
-        forEach(entries::add);
+        // Read all entries eagerly
+        TarArchiveEntry entry;
+        while ((entry = getNextTarEntry()) != null) {
+            entries.add(entry);
+        }
```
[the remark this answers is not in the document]
ppkarwasz: The previous implementation wasn't lazy either, although we should probably implement `stream` in a lazy way.

src/main/java/org/apache/commons/compress/archivers/tar/TarFile.java
```diff
@@ -240,7 +238,11 @@ private TarFile(final Builder builder) throws IOException {
         this.recordBuffer = ByteBuffer.allocate(this.recordSize);
         this.blockSize = builder.getBlockSize();
         this.lenient = builder.isLenient();
-        forEach(entries::add);
```
garydgregory: This look like a merge mistake.

garydgregory (changes requested):
It looks like there was a merge error. I'll wait to review. TY.

src/main/java/org/apache/commons/compress/archivers/tar/TarFile.java
ppkarwasz: Reverting to the old code here is deliberate.

When I dropped your custom `iterator()` implementation in favor of the default from `ArchiveFile`, it introduced a loop:

* `forEach` in `IOIterable` calls `iterator()`.
* The default `iterator()` calls `entries()`:
  https://github.com/apache/commons-compress/blob/f5dd39c8bcb1abb43270c75167008ba0330c6818/src/main/java/org/apache/commons/compress/archivers/ArchiveFile.java#L82-L86
* `entries()` calls `stream()`:
  https://github.com/apache/commons-compress/blob/f5dd39c8bcb1abb43270c75167008ba0330c6818/src/main/java/org/apache/commons/compress/archivers/ArchiveFile.java#L54-L58
* `stream()` iterates over the `entries` collection, the very collection we’re trying to populate:
  https://github.com/apache/commons-compress/blob/f5dd39c8bcb1abb43270c75167008ba0330c6818/src/main/java/org/apache/commons/compress/archivers/tar/TarFile.java#L464-L467

So `forEach(entries::add)` ends up iterating an empty collection and adding nothing, effectively a no-op. Restoring the original `iterator()` avoids this self-referential loop.

ppkarwasz:
I will retarget this PR to `commons_io_2_21_0`, once #721 is ready.

src/main/java/org/apache/commons/compress/archivers/tar/TarFile.java
ppkarwasz: I added a comment in https://github.com/apache/commons-compress/pull/709/commits/9b16aeb26c3f20ccf18cc7f978e242360f9e70e9

PR #710, opened 2025-09-25
feat: Add configurable maximum entry name length

ppkarwasz:
This change introduces a configurable limit on archive entry names.

Although formats like **AR**, **CPIO**, and **TAR** permit arbitrarily long path names, real operating systems and file systems impose much stricter limits:

* Individual path segments (file names) are typically limited to 255 bytes.
* Full paths are usually capped at a few KiB (e.g. 1024 bytes on macOS via `MAX_PATH`).

#### What’s new

* Added a new common builder, `AbstractArchiveBuilder`, inserted in the hierarchy between archive stream builders and `AbstractStreamBuilder`.
* Introduced a new configuration option: `setMaxEntryNameLength`.
* Default value is `Short.MAX_VALUE`, which is higher than any realistic OS limit.
* Enforced the limit across:

  * All `ArchiveInputStream` implementations
  * `SevenZFile`, `TarFile`, and `ZipFile`
* Added a dedicated test suite to verify:

  * Entry names up to `Short.MAX_VALUE` are handled correctly.
  * Entries exceeding a lowered limit result in an exception.

#### Exception usage

This PR applies a coherent strategy on how exceptions are thrown when parsing archive input, which also required updates to unit tests:

* **`EOFException`**
  Thrown when the stream ends *unexpectedly* (e.g., while reading a structure with a declared length).
  This signals a likely truncated archive. Users may choose to re-fetch or regenerate the archive.

* **`MemoryLimitException`**
  Thrown when the implementation determines it cannot handle the archive under the current JVM heap settings.

  * For large **and** truncated archives, either `EOFException` or `MemoryLimitException` may occur depending on available memory.
  * The test suite currently adapts to `Runtime.totalMemory()`, but this value fluctuates between `-Xms` and `-Xmx`.
  * To improve reproducibility, I propose:

    * Switching to `Runtime.maxMemory()` checks (directly tied to `-Xmx`).
    * Adding a Surefire execution with constrained heap (e.g. `-Xmx256m`) to simulate low-memory environments and can be achieved on most developer machines.

* **`ArchiveException`**
  Thrown when archive data is *structurally invalid*.
  Examples:

  * A PAX header declares key/value pairs longer than the header itself.
  * An entry path exceeds `maxEntryNameLength`.
  
  These errors cannot be resolved by simply retrying with the same archive and options. However, it may be worth distinguishing cases where the archive itself is invalid from cases where the failure results from an incompatibility between the archive and the chosen options. The latter is conceptually similar to `MemoryLimitException`, which indicates that an archive is not compatible with the JVM’s heap settings.


src/main/java/org/apache/commons/compress/archivers/ArchiveInputStream.java
```diff
@@ -114,37 +95,56 @@ public Iterator<E> unwrap() {
...
+
+    /**
+     * Constructs a new instance from a builder.
+     *
+     * @param builder The builder.
      * @since 1.29.0
      */
-    protected ArchiveInputStream(final InputStream inputStream, final Charset charset) {
-        super(inputStream);
-        this.charset = Charsets.toCharset(charset);
```
ppkarwasz: This method (to be introduced in `1.29.0`) was removed, which confuses GitHub diff algorithm.

src/main/java/org/apache/commons/compress/archivers/dump/DumpArchiveInputStream.java
```diff
@@ -547,7 +550,8 @@ private void readDirectoryEntry(DumpArchiveEntry entry) throws IOException {
 
                 final byte type = blockBuffer[i + 6];
 
-                final String name = DumpArchiveUtil.decode(zipEncoding, blockBuffer, i + 8, blockBuffer[i + 7]);
+                final String name =
+                        DumpArchiveUtil.decode(zipEncoding, blockBuffer, i + 8, Byte.toUnsignedInt(blockBuffer[i + 7]));
```
ppkarwasz: This change fixes an unrelated type conversion problem. I created a separate PR #711 for this, but I need the fix for the unit tests to succeed.

src/main/java/org/apache/commons/compress/archivers/sevenz/SevenZFile.java
```diff
@@ -1353,7 +1365,10 @@ private void readFilesInfo(final ByteBuffer header, final Archive archive) throw
                 for (int i = 0; i < namesLength; i += 2) {
                     if (names[i] == 0 && names[i + 1] == 0) {
                         computeIfAbsent(fileMap, nextFile);
-                        fileMap.get(nextFile).setName(new String(names, nextName, i - nextName, UTF_16LE));
+                        // Entry name length in UTF-16LE characters (not bytes)
+                        final int entryNameLength =
+                                ArchiveUtils.checkEntryNameLength((i - nextName) / 2, maxEntryNameLength, "7z");
+                        fileMap.get(nextFile).setName(new String(names, nextName, 2 * entryNameLength, UTF_16LE));
```
ppkarwasz: I documented `maxEntryNameLength` as limiting the path length in **bytes**, based on the character set provided through `setCharset()`.

In this case, however, the 7z format mandates **UTF-16LE** for file names, so the limit is enforced in *characters* rather than raw bytes.

This prevents the effective maximum length from being cut in half: UTF-16LE always uses 2 bytes per character, whereas most other archive formats rely on encodings where ASCII fits in 1 byte. Applying the documented `maxEntryNameLength` directly could therefore be surprising to users, while recalculating the exact byte size with `getCharset()` would add unnecessary overhead.

src/test/java/org/apache/commons/compress/archivers/cpio/CpioArchiveInputStreamTest.java
```diff
@@ -90,33 +89,6 @@ void testCpioUnarchiveMultibyteCharName() throws Exception {
         assertEquals(2, count);
     }
 
-    @Test
-    void testEndOfFileInEntry_c_namesize_0x7FFFFFFF() throws Exception {
-        // CPIO header with c_namesize = 0x7FFFFFFF
```
ppkarwasz, line 109: A similar test it now included in `MaxNameEntryLengthTest`, but with improve semantics: a size of `0x7FFF_FFFF` will never work, due to the limitation on Java arrays and `MemoryLimitException` will always be thrown.

The new test case instead, lowers the size to `0x7FFF_FFF7` and adapts its expectation depending on whether the system has 2 GiB of memory available

src/main/java/org/apache/commons/compress/archivers/tar/TarFile.java
```diff
@@ -507,7 +511,14 @@ private TarArchiveEntry getNextTarEntry() throws IOException {
             lastWasSpecial = TarUtils.isSpecialTarRecord(currEntry);
             if (lastWasSpecial) {
                 // Handle PAX, GNU long name, or other special records
-                TarUtils.handleSpecialTarRecord(getInputStream(currEntry), zipEncoding, currEntry, paxHeaders, sparseHeaders, globalPaxHeaders,
```
garydgregory, line 511: Please don't change the formatting to this newspaper column style.

src/main/java/org/apache/commons/compress/archivers/AbstractArchiveBuilder.java
```diff
@@ -0,0 +1,77 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.commons.compress.archivers;
+
+import org.apache.commons.io.build.AbstractStreamBuilder;
+
+/**
+ * Base builder for all archive formats.
```
garydgregory: Nice idea to extract this into a class that's reused from streams and non-streams.

"Base builder for all archive formats." -> "Abstracts the builder pattern for all implementations that need to implement a maximum file name length."

src/main/java/org/apache/commons/compress/archivers/ar/ArArchiveInputStream.java
```diff
@@ -244,7 +246,8 @@ public void close() throws IOException {
      * @since 1.3
      */
     private String getBSDLongName(final String bsdLongName) throws IOException {
-        final int nameLen = ParsingUtils.parseIntValue(bsdLongName.substring(BSD_LONGNAME_PREFIX_LEN));
+        final int nameLen =
```
garydgregory: This formatting is awkward, put it all on one line. Our Checkstyle rules allows for lines up to 160.

src/main/java/org/apache/commons/compress/archivers/sevenz/SevenZFile.java
```diff
@@ -677,14 +680,23 @@ public SevenZFile(final SeekableByteChannel channel, final String fileName, fina
         this(channel, fileName, password, false, SevenZFileOptions.DEFAULT);
     }
 
-    private SevenZFile(final SeekableByteChannel channel, final String fileName, final byte[] password, final boolean closeOnError, final int maxMemoryLimitKiB,
-            final boolean useDefaultNameForUnnamedEntries, final boolean tryToRecoverBrokenArchives) throws IOException {
+    private SevenZFile(
```
garydgregory: Don't change the formatting please, I don't think any of us are on 80 character TTY terminals 😉

src/main/java/org/apache/commons/compress/archivers/tar/TarArchiveInputStream.java
```diff
@@ -512,7 +513,15 @@ public TarArchiveEntry getNextEntry() throws IOException {
             lastWasSpecial = TarUtils.isSpecialTarRecord(currEntry);
             if (lastWasSpecial) {
                 // Handle PAX, GNU long name, or other special records
-                TarUtils.handleSpecialTarRecord(this, zipEncoding, currEntry, paxHeaders, sparseHeaders, globalPaxHeaders, globalSparseHeaders);
+                TarUtils.handleSpecialTarRecord(
```
garydgregory: Don't change the formatting please, I don't think any of us are on 80 character TTY terminals 😉

src/main/java/org/apache/commons/compress/archivers/tar/TarFile.java
```diff
@@ -65,7 +66,7 @@ private final class BoundedTarEntryInputStream extends BoundedArchiveInputStream
         BoundedTarEntryInputStream(final TarArchiveEntry entry, final SeekableByteChannel channel) throws IOException {
             super(entry.getDataOffset(), entry.getRealSize());
             if (channel.size() - entry.getSize() < entry.getDataOffset()) {
-                throw new ArchiveException("Entry size exceeds archive size");
+                throw new EOFException("Truncated TAR archive: entry size exceeds archive size");
```
garydgregory, line 66: I'm not sure about this ... This is not a real EOF. Why mix in this change in an large PR?

src/main/java/org/apache/commons/compress/archivers/tar/TarUtils.java
```diff
@@ -354,34 +354,44 @@ public static void formatUnsignedOctalString(final long value, final byte[] buff
...
      * @param paxHeaders the map to update with PAX headers.
      * @param sparseHeaders the list to update with sparse headers.
      * @param globalPaxHeaders the map to update with global PAX headers.
      * @param globalSparseHeaders the list to update with global sparse headers.
      * @throws IOException if an I/O error occurs while reading the entry.
      */
-    static void handleSpecialTarRecord(final InputStream input, final ZipEncoding encoding, final TarArchiveEntry entry, final Map<String, String> paxHeaders,
-            final List<TarArchiveStructSparse> sparseHeaders, final Map<String, String> globalPaxHeaders,
-            final List<TarArchiveStructSparse> globalSparseHeaders) throws IOException {
+    static void handleSpecialTarRecord(
```
garydgregory: Don't change the formatting please, I don't think any of us are on 80 character TTY terminals 😉

src/main/java/org/apache/commons/compress/archivers/tar/TarUtils.java
```diff
@@ -767,15 +813,24 @@ protected static Map<String, String> parsePaxHeaders(final InputStream inputStre
...
+                            // Validate entry length
+                            // 1. Ignore empty keywords
                             if (restLen <= 1) { // only NL
                                 headers.remove(keyword);
+                            // 2. Entry length exceeds header size
                             } else if (headerSize >= 0 && restLen > headerSize - totalRead) {
                                 throw new ArchiveException("PAX header value size %,d exceeds size of header record.", restLen);
                             } else {
+                                // 3. Entry length exceeds configurable file and link name limits
+                                if ("path".equals(keyword) || "linkpath".equals(keyword)) {
```
garydgregory: It might be worth extracting these magic strings into private constants.

src/main/java/org/apache/commons/compress/archivers/tar/TarUtils.java
```diff
@@ -767,15 +813,24 @@ protected static Map<String, String> parsePaxHeaders(final InputStream inputStre
...
+                                    ArchiveUtils.checkEntryNameLength(restLen - 1, maxEntryPathLength, "TAR");
+                                }
                                 final byte[] rest = IOUtils.readRange(inputStream, restLen);
                                 final int got = rest.length;
                                 if (got != restLen) {
-                                    throw new ArchiveException("Failed to read PAX header: Expected %,d bytes, read %,d.", restLen, got);
+                                    throw new EOFException(String.format(
```
garydgregory: No need to change the formatting or the exception IMO since this is not a real EOF.

garydgregory (changes requested):
Hi @ppkarwasz 

TY for the PR.

I stopped reviewing after a while, as there are two major changes in here, which makes the PR too big as a mix of too many topics on top of the bizarre newspaper column formatting changes (please, no).

RE the exception changes, these don't belong in the main theme of the PR and I'm not sure they makes sense in the 1st place: The `ArchiveException` says to a call site that there's a problem parsing, and `EOFException` says we've hit the end of the file/stream when reading, which doesn't appear to the actually happening. We've really hit a parsing issue from what I can tell.

Maybe `ArchiveException` should have been called `ArchiveFormatException` or `ArchiveParsingException` but `ArchiveException`  is what we have now. If you think we need a `ArchiveParsingException` to make it crystal clear, then we can do that separately.

src/main/java/org/apache/commons/compress/archivers/tar/TarFile.java
ppkarwasz: I realize this bothers you, so I’ve been correcting it whenever I noticed line breaks you don’t like.

As I mentioned earlier, the root cause is that we’re not using the same editors or sharing a common configuration. IntelliJ IDEA and Eclipse apply different formatting rules, and even after setting IDEA’s right margin to 160 characters I noticed that it sometimes still exceeded the limit until I tweaked additional options.

In https://github.com/apache/commons-compress/pull/710/commits/94c8f6ee0ce2999015b540f548414a85779b691b I adjusted the style to be closer to what you expect. That said, until we adopt a **deterministic formatter** (one that rewrites line breaks consistently and leaves only blank lines to user choice), we’ll continue to run into these inconsistencies.

src/main/java/org/apache/commons/compress/archivers/ar/ArArchiveInputStream.java
ppkarwasz: Fixed in https://github.com/apache/commons-compress/pull/710/commits/94c8f6ee0ce2999015b540f548414a85779b691b

src/main/java/org/apache/commons/compress/archivers/sevenz/SevenZFile.java
ppkarwasz: Fixed in https://github.com/apache/commons-compress/pull/710/commits/94c8f6ee0ce2999015b540f548414a85779b691b

src/main/java/org/apache/commons/compress/archivers/tar/TarArchiveInputStream.java
ppkarwasz: Fixed in https://github.com/apache/commons-compress/pull/710/commits/94c8f6ee0ce2999015b540f548414a85779b691b

src/main/java/org/apache/commons/compress/archivers/tar/TarUtils.java
ppkarwasz: Fixed in https://github.com/apache/commons-compress/pull/710/commits/94c8f6ee0ce2999015b540f548414a85779b691b

src/main/java/org/apache/commons/compress/archivers/tar/TarUtils.java
ppkarwasz: Nice point! Fixed in https://github.com/apache/commons-compress/pull/710/commits/7a3d21cbf94d012f769cf1cab7ac4cd629e51685

src/main/java/org/apache/commons/compress/archivers/AbstractArchiveBuilder.java
ppkarwasz: Fixed in https://github.com/apache/commons-compress/pull/710/commits/7599a8916fc8866629811e95b4c8a7caa082b8a6

src/main/java/org/apache/commons/compress/archivers/tar/TarFile.java
ppkarwasz: I switched the exception to `EOFException` to make `TarFile` and `TarArchiveInputStream` behave consistently, since they share the same test suite in `MaxNameEntryLengthTest`.

However, you are right: this is not a “real” EOF. With a genuine EOF you can usually still read the non-truncated part of the entry. Here the check is performed eagerly in the constructor and prevents reading anything at all.

This raises the question: should we remove this upfront check and instead let truncation errors surface naturally while reading downstream? I generally favor eager validation, but only when we know the stream will be fully consumed before handing control back to the caller. In this case the check applies unconditionally: it triggers both when we consume entries internally (e.g. PAX headers) and when users obtain the stream themselves.

One more caveat: `SeekableByteChannel#size()` is not stable by contract. It may change if the underlying file is modified, so even if this check fails here it might have succeeded if performed at a later stage.

src/main/java/org/apache/commons/compress/archivers/tar/TarUtils.java
ppkarwasz: Here I beg to differ: to me this **is** a textbook case for `EOFException`.

We know the length of a PAX header value and attempt to read it. If the stream ends before delivering the expected number of bytes, that’s exactly the scenario `EOFException` is meant to signal. It is very similar to how `DataInputStream.readLong()` throws `EOFException` if only three bytes are available.

However, in practice this branch is never hit:

* If the TAR entry itself is truncated, `TarArchiveInputStream` or `TarFile.BoundedTarEntryInputStream` already throw as soon as EOF is reached before the declared entry size (line 815)
* If a PAX header entry claims a length larger than what remains in the TAR entry, the earlier length check (line 809) fails first.

So this code is effectively unreachable except in unit tests. Which means the choice of exception here doesn’t affect real-world behavior, but I still believe `EOFException` is semantically the most accurate signal for “expected N bytes, got fewer.”

https://github.com/apache/commons-compress/blob/7a3d21cbf94d012f769cf1cab7ac4cd629e51685/src/main/java/org/apache/commons/compress/archivers/tar/TarUtils.java#L808-L819

src/main/java/org/apache/commons/compress/archivers/tar/TarFile.java
garydgregory: I was hoping that we could distinguish between low-level "real" `IOException`s  (reading issues from the underlying stream) and higher-level formatting and parsing issues using `ArchiveException`. Perhaps this was misguided. Since `ArchiveException` is a kind of `IOException`, call sites may choose to catch `IOException` anyway and leave it to readers of stack traces to figure out the difference.

In this case, and probably others in Compress, eager validation is better I think. If a call site cares about preserving the state of a stream they can always use a `PushbackInputStream`.

> I generally favor eager validation, but only when we know the stream will be fully consumed before handing control back to the caller.

That one doesn't make sense to me. That means that before throwing an exception, we'd have to read all the way to the end of an input stream? That sounds like a DoS waiting to happen. Even if you are saying that we should read to the end only if we throw `EOFException`, then it's still a bad idea for the same reason IMO.

All of that to say:
- This instance is not an EOF, it's a formatting error.
- It doesn't seem like changing exceptions in this PR is a good idea: It's thematically different, and the PR is already 42 files, making it harder to review.

src/main/java/org/apache/commons/compress/archivers/ArchiveInputStream.java
```diff
@@ -114,37 +95,56 @@ public Iterator<E> unwrap() {
     /** The number of bytes read in this stream. */
     private long bytesRead;
 
-    private Charset charset;
+    private final Charset charset;
+
+    private final int maxEntryNameLength;
```
garydgregory, line 100: In the PR comment, you write:

> Default value is Short.MAX_VALUE, which is higher than any realistic OS limit.

So shouldn't this be a `short`?

src/main/java/org/apache/commons/compress/archivers/ArchiveInputStream.java
garydgregory: I think we should (separately) deprecated `ArchiveInputStream.ArchiveInputStream(InputStream, String)`. It has no callers anyway.

src/main/java/org/apache/commons/compress/archivers/ArchiveInputStream.java
ppkarwasz: The previous _de facto_ limit was the size of a Java `String`, so I don't want to introduce incompatibilities with previous versions of Commons Compress.

Most users will be satisfied with `Short.MAX_VALUE`, but some users can always put `Integer.MAX_VALUE` instead.

garydgregory: Check.

src/main/java/org/apache/commons/compress/archivers/ArchiveInputStream.java
ppkarwasz: Sure, `(InputStream, String)` should be deprecated. As of `(InputStream, Charset)` I took the liberty to remove it right away, since we haven't published it yet.

src/main/java/org/apache/commons/compress/archivers/tar/TarFile.java
ppkarwasz: I see `EOFException` a bit differently: it’s not a “low-level” `IOException` but always a higher-level signal that fewer bytes were delivered than expected. That’s why I initially thought it fit here. However, I’m not strongly attached to using it, especially since some users (e.g. in socket scenarios) treat `EOFException` as a benign “end of stream”.

What I do want, like you, is for users to be able to distinguish causes. My angle is slightly different: I’d like them to be able to tell **truncation** from **corruption**, rather than just archive-layer vs. I/O-layer. One way to make that distinction clearer would be to introduce a dedicated `TruncatedArchiveException` for archive-level truncation cases. This could subclass either `EOFException` (to align with the semantics of “not enough data”) or `ArchiveException`.

> That one doesn't make sense to me. That means that before throwing an exception, we'd have to read all the way to the end of an input stream? That sounds like a DoS waiting to happen.

This check applies to both special TAR entries (such as PAX headers) and regular file entries. For special entries that are consumed internally, eager validation is acceptable (and essentially neutral from a security perspective, since we already use the same code paths for both `InputStream` of unknown size and `SeekableByteChannel`). For regular file entries, however, the exception should be deferred so that **users** can read all data actually available before being told the entry is incomplete.

Take a backup TAR written to a nearly full filesystem that gets truncated: ideally, the user should still be able to recover the intact portion of the last entry before being told it is incomplete. Throwing in the `BoundedTarEntryInputStream` constructor prevents this; surfacing the error lazily in `read()` preserves recoverability.

There’s also a less common but illustrative case: a TAR archive being read while another process is appending to it. This is inherently racy, but users would reasonably expect the exception to occur only when the reader overtakes the writer.

**TL;DR**

* I’m fine with replacing `EOFException` with a custom exception, but in unit tests I’d still like to:

  * Distinguish **truncation** from **corruption** cases.
  * *(Nitpick)* Verify that the exception message clearly reports how many bytes were missing (e.g., `Integer.MAX_VALUE` in a PAX header suggests corruption instead).
* I’d drop the eager exception from `BoundedTarEntryInputStream`.

ppkarwasz: This check was removed i #715, so I am closing this discussion for now.

src/main/java/org/apache/commons/compress/archivers/tar/TarUtils.java
ppkarwasz: I removed the size check in 8756b7493dd812b9901dd08a0909f09b7abfc95a.

As explained above, that check was effectively unreachable. In the unlikely event of a future programming error, `IOUtils.toByteArray` will still throw an appropriate exception.

PR #711, opened 2025-09-25
fix: Handling of long file names in `DumpArchiveInputStream`

ppkarwasz:
`DumpArchiveInputStream` previously failed to handle file names longer than `Byte.MAX_VALUE` (127) due to an incorrect type conversion.

This change fixes the interpretation of the file name length field so that it is treated as an *unsigned* 8-bit integer, allowing file names up to 255 bytes as specified by the dump format.


garydgregory:
TY @ppkarwasz ! Merged.

PR #712, opened 2025-09-25
remove: make `TarUtils` final and clean up internal methods

ppkarwasz:
`TarUtils` already had a private constructor and was only used within the `o.a.c.compress.archivers.tar` package. This PR makes the class explicitly final and simplifies its internal API:

* Mark `TarUtils` as `final`.
* Change `protected` methods to package-private (they were never usable outside the package due to the private constructor).
* Remove deprecated non-`public` methods.
* Simplify `parsePaxHeaders`:

  * Require a non-negative `headerSize`. The special value `-1` (“unlimited”) was only used in tests.
  * Update tests to supply a valid `headerSize`.
  * Standardize error handling for invalid PAX 0.0 sparse records: all parsing errors now throw `ArchiveException` (previously one case threw `IOException`).

A changelog entry is included even though these are internal changes, to give users a reference point in case any unintended side effects arise.

This change is linked with #710, which needs to add a parameter to the currently protected `parsePaxHeaders` method.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -300,6 +300,12 @@ Brotli, Zstandard and ar, cpio, jar, tar, zip, dump, 7z, arj.
                 <exclude>org.apache.commons.compress.harmony.pack200.SegmentMethodVisitor</exclude>
                 <exclude>org.apache.commons.compress.harmony.pack200.SegmentAnnotationVisitor</exclude>
                 <exclude>org.apache.commons.compress.harmony.pack200.SegmentFieldVisitor</exclude>
+                <!-- Compress 1.28.0 -> 1.29.0: make TarUtils final -->
+                <exclude>org.apache.commons.compress.archivers.tar.TarUtils#parseFromPAX01SparseHeaders(java.lang.String)</exclude>
+                <exclude>org.apache.commons.compress.archivers.tar.TarUtils#parsePAX01SparseHeaders(java.lang.String)</exclude>
+                <exclude>org.apache.commons.compress.archivers.tar.TarUtils#parsePAX1XSparseHeaders(java.io.InputStream,int)</exclude>
+                <exclude>org.apache.commons.compress.archivers.tar.TarUtils#parsePaxHeaders(java.io.InputStream,java.util.List,java.util.Map)</exclude>
+                <exclude>org.apache.commons.compress.archivers.tar.TarUtils#parsePaxHeaders(java.io.InputStream,java.util.List,java.util.Map,long)</exclude>
               </excludes>
             </parameter>
           </configuration>
--- a/pom.xml
+++ b/pom.xml
@@ -74,8 +74,6 @@ Brotli, Zstandard and ar, cpio, jar, tar, zip, dump, 7z, arj.
     <slf4j.version>2.0.16</slf4j.version>
     <!-- project.build.outputTimestamp is managed by Maven plugins, see https://maven.apache.org/guides/mini/guide-reproducible-builds.html -->
     <project.build.outputTimestamp>2025-07-29T22:23:16Z</project.build.outputTimestamp>
-    <!-- spdx 0.6.0 can require Java 11 depending on undocumented behavior which kicks in for us here. -->
-    <commons.spdx.version>0.5.5</commons.spdx.version>
     <!-- JaCoCo: Don't make code coverage worse than: -->
     <commons.jacoco.haltOnFailure>true</commons.jacoco.haltOnFailure>
     <commons.jacoco.classRatio>0.96</commons.jacoco.classRatio>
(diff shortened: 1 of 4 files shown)


src/main/java/org/apache/commons/compress/archivers/tar/TarUtils.java
```diff
@@ -48,8 +48,7 @@
  *
  * @Immutable
  */
-// CheckStyle:HideUtilityClassConstructorCheck OFF (bc)
-public class TarUtils {
+public final class TarUtils {
```
garydgregory, line 51: You cannot make these changes and maintain binary compatibility. Or am I misunderstanding the JLS and japicmp?

pom.xml
```diff
@@ -300,7 +300,26 @@ Brotli, Zstandard and ar, cpio, jar, tar, zip, dump, 7z, arj.
...
+                <exclude>org.apache.commons.compress.archivers.tar.TarUtils#parsePaxHeaders(java.io.InputStream,java.util.List,java.util.Map)</exclude>
+                <exclude>org.apache.commons.compress.archivers.tar.TarUtils#parsePaxHeaders(java.io.InputStream,java.util.List,java.util.Map,long)</exclude>
               </excludes>
+              <!--
+                ~ JapiCmp does not support ignoring a specific compatibility change for a single class.
+                ~ As a workaround, we override the change globally.
+                ~
+                ~ TODO: Remove after release 1.29.0.
+                -->
+              <overrideCompatibilityChangeParameters>
```
garydgregory: Obvious -1: We can't break binary compatibility for a trivial style/design change.

garydgregory (commented):
Obvious -1: We can't break binary compatibility for a trivial style/design changes.

src/main/java/org/apache/commons/compress/archivers/tar/TarUtils.java
ppkarwasz: I wouldn’t propose a change that breaks binary compatibility for users.

You’re correct that in general marking a class `final` is a binary-incompatible change, but in this specific case the rule only applies if the class **can be subclassed**. Since `TarUtils` has a **private constructor** and no nested subclasses, it cannot be extended, so making it `final` does not remove any legal use and therefore does not break compatibility.

This is a subtle edge case: tools like **BND** recognize it correctly, while **JapiCmp** currently does not.

pom.xml
ppkarwasz: The reason I temporarily disabled `CLASS_NOW_FINAL` instead of excluding `o.a.c.compress.archivers.tar.TarUtils` from the JapiCmp check is that I considered it more likely we’d make further adjustments to `TarUtils` before the release than that we’d mark additional classes as `final`.

That said, I can certainly roll back the `final` modifier on `TarUtils` and remove this exclusion. The most important part of this PR for me is the change of many methods from `protected` to package-private (and the removal of deprecated methods). This way we avoid being locked into keeping obsolete overloads whenever we need to add an extra parameter, as in #710.

garydgregory: OK, that makes sense; let's see what JApiCmp says about the ticket I created: https://github.com/siom79/japicmp/issues/426 

It would be nicer and better to bring in this change without messing with JApiCmp, so if they agree and fix the issue, we can pick up a new version.

FWIW, the JLS' definition of binary compatible for final classes has changed from version to version, which I didn't bother to talk about in the ticket.

ppkarwasz:
I targeted this PR to the `commons_io_2_21_0`, to limit conflicts, when that branch is merged back into `master`.

ppkarwasz:
To simplify the release I bumped the version of `japicmp` in this PR, overriding the value inherited from `commons-parent`.

Since the release of `commons-io` and `commons-parent` can be in parallel though, we can probably use both new version in the `commons_io_2_21_0` branch.

pom.xml
```diff
@@ -89,6 +87,12 @@ Brotli, Zstandard and ar, cpio, jar, tar, zip, dump, 7z, arj.
     <checkstyle.config.file>${basedir}/src/conf/checkstyle/checkstyle.xml</checkstyle.config.file>
     <checkstyle.suppress.file>${basedir}/src/conf/checkstyle/checkstyle-suppressions.xml</checkstyle.suppress.file>
     <checkstyle.resourceExcludes>LICENSE.txt, NOTICE.txt, **/maven-archiver/pom.properties</checkstyle.resourceExcludes>
+    <!--
```
garydgregory: There is no need to make this PR noisier by moving comments and changing their format to whatever this "~" is.

ppkarwasz: I moved all the overwritten versions together, so that we can more easily check each of those overrides, when we upgrade `commons-parent`.

I minimized the changes in https://github.com/apache/commons-compress/pull/712/commits/adcb46b6e785fac63153f3bb604c831d9f0bb37e

ppkarwasz:
@garydgregory,

In 333e11f8db47161efe69fdd56e385c9ae761dc6d, I added a Maven Enforcer rule intended to remind us to remove the `commons.japicmp.version` override with the next `commons-parent` upgrade.

Unfortunately, the rule fails on CI while passing locally, and I was unable to reproduce the CI failure despite trying:

* Using the same Maven version as CI (`3.9.11`)
* Clearing the local `~/.m2/repository` cache
* Running the build with `CI=true`
* Removing all repository caches

Since the build consistently fails only on CI, I dropped the rule in 854affde56ca12830a7a7353ce46108439e7c11b.

garydgregory:
PR merged, thank you @ppkarwasz 👍

PR #715, opened 2025-09-29
Improve sparse file handling performance

ppkarwasz:
Previously, sparse files were processed recursively. On highly fragmented files, this led to deep recursion and significant inefficiency.

This change replaces the recursive approach with an iterative strategy, which scales better for files with many fragments. It also introduces generated tests that simulate sparse files with very high fragmentation to ensure correctness and performance under stress.

### DIFF
--- a/src/main/java/org/apache/commons/compress/archivers/tar/ComposedTarInputStream.java
+++ b/src/main/java/org/apache/commons/compress/archivers/tar/ComposedTarInputStream.java
@@ -1,0 +1,103 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.commons.compress.archivers.tar;
+
+import java.io.EOFException;
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Iterator;
+
+import org.apache.commons.compress.utils.ArchiveUtils;
+
+final class ComposedTarInputStream extends InputStream {
+
+    final Iterator<? extends InputStream> streams;
+    private final long size;
+    private InputStream current;
+    private long position;
+
+    ComposedTarInputStream(final Iterable<? extends InputStream> streams, final long size) {
+        this.streams = streams.iterator();
+        this.size = size;
+        this.current = this.streams.hasNext() ? this.streams.next() : null;
+        this.position = 0;
+    }
+
+    @Override
+    public void close() throws IOException {
+        while (current != null) {
+            current.close();
+            current = streams.hasNext() ? streams.next() : null;
+        }
+    }
+
+    @Override
+    public int read() throws IOException {
+        if (position >= size) {
+            return -1;
+        }
+        while (current != null) {
+            final int ret = current.read();
+            if (ret != -1) {
+                position++;
+                return ret;
+            }
+            nextStream();
+        }
+        throw new EOFException(String.format("Truncated TAR archive: expected %d bytes, but got only %d bytes", size, position));
+    }
+
+    @Override
+    public int read(final byte[] b, final int off, final int len) throws IOException {
+        ArchiveUtils.checkFromIndexSize(b, off, len);
+        if (len == 0) {
+            return 0;
+        }
+        if (position >= size) {
+            return -1;
+        }
+
+        final int toRead = (int) Math.min(size - position, len);
+        int remaining = toRead;
+        int dst = off;
+
+        while (current != null && remaining > 0) {
+            final int n = current.read(b, dst, remaining);
+            if (n == -1) {
+                nextStream();
+                continue;
+            }
+            position += n;
+            dst += n;
+            remaining -= n;
+        }
+
+        if (remaining == 0) {
+            return toRead;
+        }
+        throw new EOFException(String.format("Truncated TAR archive: expected %d bytes, but got only %d bytes", size, position));
+    }
+
+    private void nextStream() throws IOException {
+        if (current != null) {
+            current.close();
+        }
+        current = streams.hasNext() ? streams.next() : null;
+    }
+}
--- a/src/main/java/org/apache/commons/compress/archivers/tar/TarArchiveInputStream.java
+++ b/src/main/java/org/apache/commons/compress/archivers/tar/TarArchiveInputStream.java
@@ -25,6 +25,7 @@
 package org.apache.commons.compress.archivers.tar;
 
 import java.io.ByteArrayOutputStream;
+import java.io.EOFException;
 import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
@@ -149,20 +150,14 @@ public class TarArchiveInputStream extends ArchiveInputStream<TarArchiveEntry> {
     /** True if stream is at EOF. */
     private boolean atEof;
 
-    /** Size of the current. */
-    private long entrySize;
-
     /** How far into the entry the stream is at. */
     private long entryOffset;
 
-    /** Input streams for reading sparse entries. **/
-    private List<InputStream> sparseInputStreams;
-
-    /** The index of current input stream being read when reading sparse entries. */
-    private int currentSparseInputStreamIndex;
-
     /** The meta-data about the current entry. */
     private TarArchiveEntry currEntry;
+
+    /** The current input stream. */
+    private InputStream currentInputStream;
 
     /** The encoding of the file. */
     private final ZipEncoding zipEncoding;
@@ -332,8 +327,7 @@ public class TarArchiveInputStream extends ArchiveInputStream<TarArchiveEntry> {
      * </p>
      */
     private void buildSparseInputStreams() throws IOException {
-        currentSparseInputStreamIndex = -1;
-        sparseInputStreams = new ArrayList<>();
+        final List<InputStream> sparseInputStreams = new ArrayList<>();
         final List<TarArchiveStructSparse> sparseHeaders = currEntry.getOrderedSparseHeaders();
         // Stream doesn't need to be closed at all as it doesn't use any resources
         final InputStream zeroInputStream = new TarArchiveSparseZeroInputStream(); // NOSONAR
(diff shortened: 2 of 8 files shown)


src/main/java/org/apache/commons/compress/archivers/tar/ComposedTarInputStream.java
```diff
@@ -0,0 +1,103 @@
...
+    final Iterator<? extends InputStream> streams;
+    private final long size;
+    private InputStream current;
+    private long position;
+
+    ComposedTarInputStream(final Iterable<? extends InputStream> streams, final long size) {
+        this.streams = streams.iterator();
+        this.size = size;
+        this.current = this.streams.hasNext() ? this.streams.next() : null;
+        this.position = 0;
```
garydgregory: Don't initialize instance variables to their default values. But why not use (or extend)  `java.io.SequenceInputStream`?

src/main/java/org/apache/commons/compress/utils/ArchiveUtils.java
```diff
@@ -261,6 +262,23 @@ public static String toString(final ArchiveEntry entry) {
...
+    /**
+     * Checks that the specified range is within the bounds of an array of the specified length.
+     *
+     * @param b           the array
+     * @param off         the starting offset of the range
+     * @param len         the length of the range
+     * @throws IndexOutOfBoundsException if {@code off} is negative, or {@code len} is negative, or {@code off + len} is greater than {@code arrayLength}
+     * @since 1.29.0
+     */
+    public static void checkFromIndexSize(final byte[] b, final int off, final int len) {
```
garydgregory: Let's not add this as a public method since it is planned to be in Commons IO https://github.com/apache/commons-io/pull/790

src/test/java/org/apache/commons/compress/archivers/TestArchiveGenerator.java
```diff
@@ -0,0 +1,390 @@
...
+        }
+        return baos.toByteArray();
+    }
+
+    private static void writePaxKeyValue(final String key, final String value, final PrintWriter out) {
+        final String entry = ' ' + key + "=" + value + "\n";
+        // Guess length: length of length + space + entry
+        final int length = String.valueOf(entry.length()).length() + entry.length();
+        // Recompute if number of digits changes
+        out.print(String.valueOf(length).length() + entry.length());
```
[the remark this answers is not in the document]
ppkarwasz: No need for an iteration: `length` is equal for length of the entire PAX header (length as decimal integer + `entry`), except the rare case where `entry.length()` and `length` have a different number of digits. In this case a second iteration is required:

- If `entry` is `9` characters, `length` is `10` (a 2 digit number) and the real length of the PAX header becomes `11`.
- If `entry` is `98` characters, `length` is `100` (a 3 digit number) and the real length of the PAX header becomes `101`.

src/main/java/org/apache/commons/compress/utils/ArchiveUtils.java
```diff
@@ -261,6 +262,23 @@ public static String toString(final ArchiveEntry entry) {
...
+        // TODO: replace with IOUtils.checkFromIndexSize, after upgrading to Commons IO 2.21.0
+        Objects.requireNonNull(b, "byte array");
+        if ((off | len) < 0 || b.length - len < off) {
+            throw new IndexOutOfBoundsException(String.format("Range [%s, %<s + %s) out of bounds for length %s", off, len, b.length));
```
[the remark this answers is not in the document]
ppkarwasz: `%<s` references the previous and first argument: `off`.

src/main/java/org/apache/commons/compress/archivers/tar/TarArchiveInputStream.java
```diff
@@ -532,11 +513,20 @@ public TarArchiveEntry getNextEntry() throws IOException {
...
-        entrySize = currEntry.getSize();
         return currEntry;
     }
 
+    private void afterRead(final int read) throws IOException {
+        // Count the bytes read
+        count(read);
+        // Check for truncated entries
+        if (read == -1 && entryOffset < currEntry.getSize()) {
+            throw new EOFException(String.format("Truncated TAR archive: entry '%s' expected %d bytes, but got %d", currEntry.getName(), currEntry.getSize(),
```
garydgregory: Use `%,d` instead of `%d` to use the locale-dependent thousand separator. 

Use `actual` instead of `but got` which makes the vocabulary parallel unit tests. I've been trying to be consitent on this elsewhere in Commons.

FWIW, I agree here that this is an actual EOF since we read a -1.

src/main/java/org/apache/commons/compress/archivers/tar/TarArchiveInputStream.java
```diff
@@ -797,21 +718,11 @@ public long skip(final long n) throws IOException {
...
-        long skipped;
-        if (!currEntry.isSparse()) {
-            skipped = IOUtils.skip(in, numToSkip);
-            // for non-sparse entry, we should get the bytes actually skipped bytes along with
-            // inputStream.available() if inputStream is instance of FileInputStream
-            skipped = getActuallySkipped(availableOfInputStream, skipped, numToSkip);
-        } else {
-            skipped = skipSparse(numToSkip);
+        if (currEntry == null || currentInputStream == null) {
+            throw new IllegalStateException("No current tar entry");
```
garydgregory, line 742: Is this indicative of a programming error (which grants the use of ISE) or an archive format exception, in which case an `ArchiveException` would be justified.

src/main/java/org/apache/commons/compress/archivers/tar/TarFile.java
```diff
@@ -57,85 +61,30 @@ private final class BoundedTarEntryInputStream extends BoundedArchiveInputStream
...
-            }
+            requireNonNull(buf, "ByteBuffer");
+            // The caller ensures that [pos, pos + buf.remaining()] is within [start, end]
+            channel.position(pos);
+            final int totalRead = channel.read(buf);
             if (totalRead == -1) {
-                if (buf.array().length > 0) {
-                    throw new ArchiveException("Truncated TAR archive");
+                if (buf.remaining() > 0) {
+                    throw new EOFException(String.format("Truncated TAR archive: expected at least %d bytes, but got only %d bytes",
```
garydgregory, line 84: FWIW, I agree here that this is an actual EOF since we read a -1.

src/main/java/org/apache/commons/compress/archivers/tar/TarArchiveInputStream.java
```diff
@@ -820,37 +731,15 @@ public long skip(final long n) throws IOException {
...
-            return in.skip(n);
-        }
-        long bytesSkipped = 0;
-        while (bytesSkipped < n && currentSparseInputStreamIndex < sparseInputStreams.size()) {
-            final InputStream currentInputStream = sparseInputStreams.get(currentSparseInputStreamIndex);
-            bytesSkipped += currentInputStream.skip(n - bytesSkipped);
-            if (bytesSkipped < n) {
-                currentSparseInputStreamIndex++;
+            if (skipped != padding) {
+                throw new EOFException(String.format("Truncated TAR archive: failed to skip record padding for entry '%s'", currEntry.getName()));
```
garydgregory, line 760: This looks to me like we didn't read a -1 so it might be better as an `ArchiveException`.

src/main/java/org/apache/commons/compress/archivers/tar/ComposedTarInputStream.java
ppkarwasz: Nice catch! :100:

I removed the class in https://github.com/apache/commons-compress/pull/715/commits/38b1ecb9953c8af09ef3ab18b08d014418b3bfb0.

src/main/java/org/apache/commons/compress/utils/ArchiveUtils.java
ppkarwasz: The new GitHub diff view is behaving strangely.

I removed the method in https://github.com/apache/commons-compress/pull/715/commits/f3a4f8029d6670838016adf66535715c3f31fac7, i.e. before your review.

src/main/java/org/apache/commons/compress/archivers/tar/TarArchiveInputStream.java
ppkarwasz: Fixed in https://github.com/apache/commons-compress/pull/715/commits/c164798e7ea96181a2f87b99a96766b8400fa921

src/main/java/org/apache/commons/compress/archivers/tar/TarArchiveInputStream.java
ppkarwasz: This situation occurs when `skip()` is called **before** `getNextEntry()`, so it represents a programming error rather than a malformed archive, which justifies using `IllegalStateException`.

However, `java.util.ZipInputStream` silently returns `0` in the same scenario. We could consider aligning with that behavior across all archive formats, but that would be best handled in a separate PR.

src/main/java/org/apache/commons/compress/archivers/tar/TarArchiveInputStream.java
ppkarwasz: If the number of padding bytes doesn’t match what we expect, that still suggests a **truncated archive**, so I’d lean toward keeping it as an `EOFException`.

That said, we could also decide that an archive that isn’t properly closed is still *technically valid* and just ignore the mismatch. What do you think?

src/main/java/org/apache/commons/compress/archivers/tar/TarFile.java
```diff
@@ -49,91 +53,45 @@
...
 
-        private long entryOffset;
-
-        private int currentSparseInputStreamIndex;
-
-        BoundedTarEntryInputStream(final TarArchiveEntry entry, final SeekableByteChannel channel) throws IOException {
-            super(entry.getDataOffset(), entry.getRealSize());
-            if (channel.size() - entry.getSize() < entry.getDataOffset()) {
-                throw new ArchiveException("Entry size exceeds archive size");
-            }
```
ppkarwasz, line 67: Removing this check lets users read as much of the TAR entry as is actually available, even if the archive is truncated.

The `read()` method below already detects a short stream and throws an `EOFException` if the data ends before the declared entry size.

src/main/java/org/apache/commons/compress/archivers/tar/TarArchiveInputStream.java
garydgregory: > If the number of padding bytes doesn’t match what we expect,

Hm, but does that mean we are actually very near the end of the stream/file, or, could it be that we are in the middle of it, or even toward the start, and that the archive is completely corrupted?

If we know we are at or near the end, then EOFE seems OK, but not if we are in the middle of the file and we are really in a broken file situation. 

Not that big of a deal, but it's nice to know the EOF really means "I fell off the end".

ppkarwasz: `IOUtils.skip` reads until EOF, so this exception truly means the stream is exhausted.

The only question is whether we should be tolerant (according to Postel's law) and allow EOF within the padding area. In fact, we already do so:

* When the archive omits the two zero blocks (non-compliant but tolerated). In fact we don't even throw an exception if EOF occurs while reading a record, which is probably a bug.
* When the archive isn’t padded to a full block multiple (non-required but common).

ppkarwasz:
@garydgregory,

It's done, should I merge this?

PR #716, opened 2025-10-02
Use `IOUtils.checkIndexFromSize` for argument validation

ppkarwasz:
This change replaces custom argument checks in all `InputStream` and `OutputStream` implementations with the common utility method `IOUtils.checkIndexFromSize`. This ensures consistent validation logic across the codebase.

Some implementations are intentionally left unchanged: cases where the arguments are immediately delegated to methods such as `InputStream#read`, `OutputStream#write`, or `ByteBuffer#wrap`, which already perform equivalent checks internally.


garydgregory:
PR merged, thank you @ppkarwasz.

PR #721, opened 2025-10-03
Add `AbstractArchiveBuilder` for unified archiver support

ppkarwasz:
This PR introduces a new `AbstractArchiveBuilder`, which serves as a shared foundation for all archive input streams, including `SevenZFile`, `TarFile`, and `ZipFile`.

This refactoring also leverages the new `ChannelOrigin` from Commons IO 2.21.0, streamlining the builder implementations for `SevenZFile`, `TarFile`, and `ZipFile`.

### Additional improvements

* **Constructor unification:** All constructors now delegate to a single constructor that accepts a builder.
* **Resource safety:** If an error occurs during construction (or within the builder’s `get()` method), the underlying resource is now always closed. Previously, resources provided via `SeekableByteChannel` were not closed on error, potentially leaving channels in an incoherent state.
* **Deprecation cleanup:** All references to deprecated constructors and methods have been removed. Compatibility is verified through `LegacyConstructorsTest`, which ensures the builder pattern behaves equivalently to the removed constructors.


garydgregory (changes requested):
Hi @ppkarwasz,
We have some build failures on Windows.

ppkarwasz:
@garydgregory,

I fixed the failure, which was due to a typo in a test: the default encoding of Zip is **always** UTF-8 and not the default encoding of the system.

PR #723, opened 2025-10-06
ARJ: correct byte accounting and truncation errors

ppkarwasz:
In the current implementation:

* `getBytesRead()` could drift from the actual archive size after a full read.
* Exceptions on truncation errors were inconsistent or missing.
* `DataInputStream` (big-endian) forced ad-hoc helpers for ARJ’s little-endian fields.

This PR introduces:

* **Accurate byte accounting:** count all consumed bytes across main/file headers, variable strings, CRCs, extended headers, and file data. `getBytesRead()` now matches the archive length at end-of-stream.
* **Consistent truncation handling:**

  * Truncation in the **main (archive) header**, read during construction, now throws an `ArchiveException` **wrapping** an `EOFException` (cause preserved).
  * Truncation in **file headers or file data** is propagated as a plain `EOFException` from `getNextEntry()`/`read()`.
* **Endianness refactor:** replace `DataInputStream` with `EndianUtils`, removing several bespoke helpers and making intent explicit.

* Add assertion that `getBytesRead()` equals the archive size after full consumption.
* Parameterized truncation tests at key boundaries (signature, basic/fixed header sizes, end of fixed/basic header, CRC, extended-header length, file data) verifying the exception contract above.

### DIFF
--- a/src/main/java/org/apache/commons/compress/archivers/arj/ArjArchiveInputStream.java
+++ b/src/main/java/org/apache/commons/compress/archivers/arj/ArjArchiveInputStream.java
@@ -31,7 +31,6 @@ import org.apache.commons.compress.archivers.ArchiveEntry;
 import org.apache.commons.compress.archivers.ArchiveException;
 import org.apache.commons.compress.archivers.ArchiveInputStream;
 import org.apache.commons.compress.utils.ArchiveUtils;
-import org.apache.commons.compress.utils.IOUtils;
 import org.apache.commons.io.EndianUtils;
 import org.apache.commons.io.input.BoundedInputStream;
 import org.apache.commons.io.input.ChecksumInputStream;
@@ -396,15 +395,6 @@ public class ArjArchiveInputStream extends ArchiveInputStream<ArjArchiveEntry> {
         }
 
         return header;
-    }
-
-    private byte[] readRange(final InputStream in, final int len) throws IOException {
-        final byte[] b = IOUtils.readRange(in, len);
-        count(b.length);
-        if (b.length < len) {
-            throw new EOFException();
-        }
-        return b;
     }
 
     private ByteArrayOutputStream readString(final InputStream dataIn) throws IOException {


src/main/java/org/apache/commons/compress/archivers/arj/ArjArchiveInputStream.java
```diff
@@ -225,164 +235,174 @@ public int read(final byte[] b, final int off, final int len) throws IOException
...
+     */
     private byte[] readHeader() throws IOException {
-        boolean found = false;
-        byte[] basicHeaderBytes = null;
-        do {
+        byte[] basicHeaderBytes;
+        // TODO: Explain why we are scanning for a valid ARJ header
+        //       and don't throw, when an invalid/corrupted header is found,
+        //       which might indicate a corrupted archive.
+        while (true) {
```
ppkarwasz, line 270: The current code scans for the ARJ magic anywhere, ignoring invalid or corrupted entries. This favors the reading of self-extracting archives over strict validation of normal archives.

In a follow-up PR, I would like to:

* Use scanning **only** to locate the Main Archive Header (validate magic, lengths, CRC).
* After AMH, parse **strictly and contiguously** to end-of-archive; no skipping of invalid entries mid-stream.
* Add a builder option (e.g., `lenient` or `selfExtracting`) to enable AMH scanning, while by default assume that the ARJ archive starts at offset 0.

@garydgregory, what do you think?

ppkarwasz: A prototype of the `selfExtracting` option is available in branch [feat/arj-strict-validation](https://github.com/apache/commons-compress/tree/feat/arj-strict-validation)

src/main/java/org/apache/commons/compress/archivers/arj/ArjArchiveInputStream.java
```diff
@@ -128,6 +125,8 @@ private ArjArchiveInputStream(final InputStream inputStream, final Builder build
             if ((mainHeader.arjFlags & MainHeader.Flags.VOLUME) != 0) {
                 throw new ArchiveException("Multi-volume ARJ files are unsupported");
             }
+        } catch (final ArchiveException e) {
+            throw e;
         } catch (final IOException e) {
             throw new ArchiveException(e.getMessage(), (Throwable) e);
```
garydgregory: Since `ArchiveException` is an `IOException`, should we let this propagate as is, reflecting a real, low-level error? Then, there would be no need to rethrow the `ArchiveException` above.

src/main/java/org/apache/commons/compress/archivers/arj/ArjArchiveInputStream.java
garydgregory: Hi @ppkarwasz 

> DataInputStream (big-endian) forced ad-hoc helpers for ARJ’s little-endian fields.

I don't understand what the above means.

```java
        // TODO: Explain why we are scanning for a valid ARJ header
        //       and don't throw, when an invalid/corrupted header is found,
        //       which might indicate a corrupted archive.
```
Yes, please do explain it, or perhaps this was meant to be a draft PR? 😉 

I'm concerned that this is switching behavior around so much that we should instead introduce a `STRICT`/`LENIENT` `enum` option in the root builder that different classes can implement as they each best see fit. This would allow us to keep existing behavior or at least have users opt-in to whether they want strict or lenient behavior, no matter what the default is, where we can then decide when what defaults should be changed. Either way, if we do change the default behavior to strict and break some users' oddball files, they can ask for lenient processing.

WDYT?

ppkarwasz: > `DataInputStream` (big-endian) forced ad-hoc helpers for ARJ’s little-endian fields.

ARJ writes numbers in little-endian format, so I found that using `DataInputStream` is not very helpful, since we still need to inverse the order of the bytes after. Instead I used `EndianUtils` on a “normal” input stream, which in my opinion simplifies the class.

> Yes, please do explain it, or perhaps this was meant to be a draft PR? 😉

The `ArjArchiveInputStream` is the only input stream in Commons Compress that does not throw, when an invalid entry is found (e.g. mismatching CRC). The invalid entry is simply skipped.

Since we don't have any tests that use this feature, my guess is that it was meant to handle self-extracting archives, which contain an executable stub and a regular ARJ archive. If an “invalid” entry is found in the executable stub, it should be skipped, but an “invalid” entry in the data part of the executable should throw.

This PR does **not** modify the existing behavior. I opened #728 for that.

src/main/java/org/apache/commons/compress/archivers/arj/ArjArchiveInputStream.java
ppkarwasz: That’s an excellent point! :100:

The new `Builder.get()` method declares `throws Exception`, but many of the older or deprecated constructors either don’t declare any checked exceptions or use `ArchiveException`. To avoid **source** incompatibility, we had to add extra “helper” constructors.

Since we’ve already introduced some **source-incompatible** changes elsewhere in Commons Compress, maybe it’s simpler to just have the old constructor declare `throws IOException`. What do you think?

garydgregory: I'm OK with changing the signature to throw IOException, it's binary compatible. Go for it 😄

ppkarwasz: I did it for all archivers in #731.

We probably need to sleep on this, since a binary compatible, but source incompatible change will still break many user applications.

garydgregory: Check on #731, merged!

ppkarwasz:
@garydgregory,

I resolved the conflicts and sorted the methods alphabetically.

PR #754, opened 2025-12-05
[COMPRESS-714] Internal IllegalArgumentException in ZipFile and TarFile creation is not caught

garydgregory:
Fixes Jira https://issues.apache.org/jira/browse/COMPRESS-714


src/main/java/org/apache/commons/compress/archivers/tar/TarFile.java
```diff
@@ -191,13 +191,16 @@ private TarFile(final Builder builder) throws IOException {
             while ((entry = getNextTarEntry()) != null) {
                 entries.add(entry);
             }
-        } catch (final IOException ex) {
+        } catch (final IOException | IllegalArgumentException e) {
```
ppkarwasz: I’m wondering whether extending the catch clause from `IOException` to `IOException | IllegalArgumentException` is the right boundary. If the intention is to shield users from **any** unexpected internal failures we should rather catch `Exception`.

However, I see unchecked exceptions here as symptoms of underlying bugs. Masking them in the constructor can improve user experience, but only hides the bugs, instead of solving them.

In the specific case of COMPRESS-714, the `IllegalArgumentException` comes from a bug in `SeekableInMemoryByteChannel`. According to the `position()` contract, providing a position larger than the channel size should not result in an exception, so the current implementation violates that contract:

https://github.com/apache/commons-compress/blob/b83da24c57ec32cf4c4ee7dcb58b09861c9c17ce/src/main/java/org/apache/commons/compress/utils/SeekableInMemoryByteChannel.java#L114-L121

ppkarwasz (changes requested):

src/main/java/org/apache/commons/compress/archivers/tar/TarFile.java
garydgregory: Hm good catch, here's a fix: https://github.com/apache/commons-compress/pull/756 (now merged).

src/main/java/org/apache/commons/compress/archivers/sevenz/SevenZFile.java
```diff
@@ -629,9 +629,7 @@ private SevenZFile(final Builder builder) throws IOException {
             archive = readHeaders(password);
             this.password = password != null ? Arrays.copyOf(password, password.length) : null;
         } catch (final ArithmeticException | IllegalArgumentException e) {
-            final ArchiveException archiveException = new ArchiveException(e);
-            IOUtils.close(channel, archiveException::addSuppressed);
-            throw archiveException;
+            throw IOUtils.closeQuietly(channel, new ArchiveException(e));
         }
```
ppkarwasz, line 640: Honestly, I wouldn't delay the release of `commons-compress` version `1.29.0` more than it already has been to wait for a new version of `commons-io`.

Let's use what we have in `2.21.0`:

```java
        } catch (final Exception e) {
            final IOException ioe = e instanceof IOException ? (IOException) e : new ArchiveException(e);
            IOUtils.closeQuietly(channel, ioe::addSuppressed);
            throw ioe;
        }
```

When we **do** release Commons IO version `2.22.0`, we'll update this call site and **all** the other call sites that might benefit from the new method. There is no chance to forget that, since `closeQuietly` becomes ambiguous in IO `2.22.0`.

src/main/java/org/apache/commons/compress/archivers/zip/ZipFile.java
```diff
@@ -756,12 +756,8 @@ private ZipFile(final Builder builder) throws IOException {
...
             fillNameMap();
         } catch (final IOException e) {
-            final ArchiveException archiveException = e instanceof ArchiveException
-                    ? (ArchiveException) e
-                    : new ArchiveException("Error reading Zip content from " + builder.getName(), (Throwable) e);
             this.closed = true;
-            IOUtils.close(archive, archiveException::addSuppressed);
-            throw archiveException;
+            throw IOUtils.closeQuietly(archive, e);
         }
```
ppkarwasz, line 761: Same as above: I wouldn't rely on an unpublished IO method and use what is available now.

```java
        } catch (final Exception e) {
            this.closed = true;
            final IOException ioe = e instanceof IOException ? (IOException) e : new ArchiveException(e);
            IOUtils.closeQuietly(archive, ioe::addSuppressed);
            throw ioe;
        }
```

ppkarwasz (changes requested):

jtattermusch:
> @@aisle-analyzer experiment.

it's supposed to be @aisle-analyzer, can you please try again?

ideally in a new comment.

PR #762, opened 2026-03-08
[COMPRESS-720] Integrate OSS-Fuzz fuzzers and enable CIFuzz

vishalcoc44:
JIRA Ticket: [COMPRESS-720](https://issues.apache.org/jira/browse/COMPRESS-720)
Currently, our fuzz testing lives externally in the google/oss-fuzz repository. While effective, this creates a gap between development and security testing. and also the fact that google is increasingly wanting their fuzzers to live in the upstream repositories.

This integration ensures that:
Fuzzers evolve with the code: No more "bit-rot" when internal APIs change.

Instant Feedback: The new CIFuzz workflow automatically stress-tests every Pull Request before it's merged.

Developer Empowerment: Any contributor can now run these security tests locally with a single Maven command.

Maven Integration: * Added jazzer-junit as a test-scoped dependency.

Introduced a fuzz Maven profile. This keeps the fuzzers tucked away during standard mvn test runs but makes them easy to trigger via mvn test -Pfuzz.

17 Fuzzer Targets: * We’ve integrated coverage for all major formats: Zip, Tar, 7z, Ar, Arj, Cpio, Dump, and several compressors (BZip2, Gzip, LZ4, Snappy, Z, etc.).

Automated Workflow: * Added .github/workflows/cifuzz.yml to tap into Google's cifuzz actions for continuous security monitoring. (this workflow will run a mini cifuzz test everytime someone changes something in the repo.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -219,12 +219,6 @@ Brotli, Zstandard and ar, cpio, jar, tar, zip, dump, 7z, arj.
       <artifactId>junit-pioneer</artifactId>
       <scope>test</scope>
     </dependency>
-    <dependency>
-      <groupId>com.code-intelligence</groupId>
-      <artifactId>jazzer-junit</artifactId>
-      <version>0.24.0</version>
-      <scope>test</scope>
-    </dependency>
   </dependencies>
   <scm>
     <connection>scm:git:https://gitbox.apache.org/repos/asf/commons-compress.git</connection>
@@ -411,6 +405,15 @@ Brotli, Zstandard and ar, cpio, jar, tar, zip, dump, 7z, arj.
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-checkstyle-plugin</artifactId>
+      </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <configuration>
+          <testExcludes>
+            <testExclude>org/apache/commons/compress/fuzz/**</testExclude>
+          </testExcludes>
+        </configuration>
       </plugin>
     </plugins>
   </build>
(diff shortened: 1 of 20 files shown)


pom.xml
```diff
@@ -219,6 +219,12 @@ Brotli, Zstandard and ar, cpio, jar, tar, zip, dump, 7z, arj.
       <artifactId>junit-pioneer</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.code-intelligence</groupId>
+      <artifactId>jazzer-junit</artifactId>
+      <version>0.24.0</version>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
```
raboof: I find this dependency, which is not trivial to build and includes native code, difficult to trust.

vishalcoc44: @raboof can you take a look at the new changes, i've removed jazzer-junit entirely. The fuzzer targets are now vanilla Java classes with no external requirements.

src/test/java/org/apache/commons/compress/fuzz/BaseTests.java
```diff
@@ -0,0 +1,53 @@
...
+import org.apache.commons.compress.archivers.ArchiveInputStream;
+import org.apache.commons.compress.compressors.CompressorInputStream;
+
+import java.io.IOException;
+import java.util.logging.LogManager;
+
+/**
+ * Class with common functionality shared among fuzzing harnesses.
+ */
+public class BaseTests {
```
garydgregory: Hello @vishalcoc44 

If this class is meant to be an abstract class, then please declare it as such and use the prefix "Abstract" in its name instead of "Base". Otherwise it looks like this class could be instantiated.

garydgregory (changes requested):

src/test/java/org/apache/commons/compress/fuzz/BaseTests.java
vishalcoc44: Hello @garydgregory 
I've rectified that now.

pom.xml
```diff
@@ -406,6 +406,15 @@ Brotli, Zstandard and ar, cpio, jar, tar, zip, dump, 7z, arj.
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-checkstyle-plugin</artifactId>
       </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <configuration>
+          <testExcludes>
+            <testExclude>org/apache/commons/compress/fuzz/**</testExclude>
```
garydgregory, line 414: Do we really need to NOT compile this code? It seems like a typo away from breaking something.

PR #763, opened 2026-03-10
docs: document security model

raboof:

- [ ] Write unit tests that match behavioral changes, where the tests fail if the changes to the runtime are not applied. This may not always be possible, but it is a best practice.
- [ ] Each commit in the pull request should have a meaningful subject line and body. Note that a maintainer may squash commits during the merge process.

### DIFF
--- a/src/site/xdoc/security.xml
+++ b/src/site/xdoc/security.xml
@@ -31,6 +31,28 @@
         of the Commons project</a>.</p>
       </section>
 
+      <section name="Security Model">
+        <p>Commons Compress may be used to process untrusted input data.
+        Input may be read from the network when this is explicitly requested,
+        but the processing of the compressed payload should not trigger
+        additional network access, unbounded deserialization
+        or code execution. It may require processing power proportional to
+        the size of the input. An input causing an OutOfMemoryError,
+        StackOverflowError, or another Error may be considered a DoS
+        vulnerability when it is unexpected based on the size of the input
+        and the resources (RAM, disk, etc) allocated to the process.</p>
+
+        <p>Input causing any other RuntimeException is expected and not a
+        vulnerability, but may be a regular bug when a more appropriate
+        Exception is available</p>
+
+        <p>Regular API parameters are typically designed to work with input
+        that is either trusted or validated/sanitized by the application
+        using the library unless otherwise specified, in accordance with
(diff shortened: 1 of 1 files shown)


garydgregory (changes requested):
I think we need to be more specific on "network" access because you can use a builder's setPath(...), setURI(...), or constructor's InputStream to cause access to a network drive ("network" as in LAN or WAN) or the internet at large. So I think we should skip this part or document that if you don't want that, then you need to guard for it.

WDYT?

raboof:
> I think we need to be more specific on "network" access because you can use a builder's setPath(...), setURI(...), or constructor's InputStream to cause access to a network drive ("network" as in LAN or WAN) or the internet at large. So I think we should skip this part or document that if you don't want that, then you need to guard for it.

I've now worded this as: "Input may be read from the network when this is explicitly requested, but the processing of the compressed payload should not trigger additional network access, unbounded deserialization or code execution.". WDYT?

garydgregory:
> unbounded deserialization

Much better :) But don't you mean "unbounded decompression" instead of "unbounded deserialization"?

raboof:
> > unbounded deserialization
> 
> Much better :) But don't you mean "unbounded decompression" instead of "unbounded deserialization"?

I really did mean "unbounded deserialization" (i.e. deserialization allowing instantiation of arbitrary classes) - perhaps "unrestricted deserialization" would be clearer?

Taking a step back here, to be honest I'm still somewhat on the fence on if we should promise the absence of DoS conditions even for small inputs. I added that because that's what I understood the consensus to be from the past advisories and the discussions on the list, but as I've been getting more involved as a volunteer I'm not entirely sure myself that this is something we're well-prepared to promise. I'm starting to lean towards only promising this for a few well-understood components, like Zip and Tar, and not promise it for 'all of them'.

How would you feel about that?

garydgregory:
I agree that we should not state anything in absolute terms like "guarantee", which you don't use anyway.

Anything "unbounded" is too loose I think in light of the discussion currently going on in https://github.com/apache/commons-compress/pull/777

Whether decompression or deserialization, in general, it's up to the app to configure our code or have the app guard against that case. We could state exceptions around proportional input size, and that's still loose since we're not going to state a number.

AFK here, let's keep cycling on this...

PR #764, opened 2026-03-11
COMPRESS-721 -- avoid reflection when possible in unpack200

tballison:

- [*] Read the [contribution guidelines](CONTRIBUTING.md) for this project.
- [ *] Read the [ASF Generative Tooling Guidance](https://www.apache.org/legal/generative-tooling.html) if you use Artificial Intelligence (AI).
- [* ] I used AI to create any part of, or all of, this pull request. Which AI tool was used to create this pull request, and to what extent did it contribute?
Claude Opus 4.6 significantly
- [* ] Run a successful build using the default [Maven](https://maven.apache.org/) goal with `mvn`; that's `mvn` on the command line by itself.
- [ ] Write unit tests that match behavioral changes, where the tests fail if the changes to the runtime are not applied. This may not always be possible, but it is a best practice.
- [* ] Write a pull request description that is detailed enough to understand what the pull request does, how, and why.
- [ *] Each commit in the pull request should have a meaningful subject line and body. Note that a maintainer may squash commits during the merge process.

### DIFF
--- a/src/changes/changes.xml
+++ b/src/changes/changes.xml
@@ -92,6 +92,7 @@ The <action> type attribute can be add,update,fix,remove.
       <action type="fix" dev="pkarwasz" due-to="Piotr P. Karwasz">Correct byte accounting and truncation errors in ARJ input stream.</action>
       <action type="fix" dev="pkarwasz" due-to="Piotr P. Karwasz">Add strict header validation in ARJ input stream and `selfExtracting` option.</action>
       <!-- FIX unpack200 -->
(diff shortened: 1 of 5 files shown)


garydgregory (changes requested):
Hello @tballison 

-1: These tests don't test anything new. They always pass, whether or not the changes to `main` are applied, inviting a regression to whatever you think you're fixing.

tballison:
@garydgregory this is the go bigger option with fully removing reflection. Pack200ModuleAccessTest fails when the fix is reverted.

tballison:
K, I moved this to draft to do a bit more research. I think this is ready for review now.

My belief is that the band-length OOM guard added in a3474d158 (unpack200 hardening) makes the BoundedInputStream's Codec.check() tripwire redundant...checkIntArray runs whether or not the stream is bounded.

This means that we could remove the BoundedInputStream entirely, but that would touch more of the code and would be a bigger PR.

Let me know what you think.

PR #765, opened 2026-03-28
[COMPRESS-722] Introduce handling of entries compressed by a specific method

uvoigt:
#### Content
This introduces the mode `autoCompress` of `ZipArchiveOutputStream`. If enabled, the stream uses factories of the implemented methods to create a compressing `OutputStream` per entry. When using this mode, it is no longer necessary to explicitly set the uncompressed size of an entry.

#### Additional modifications:
- It is no longer necessary to close explicitly close the last archive entry.

---

- [ x Write a pull request description that is detailed enough to understand what the pull request does, how, and why.

### DIFF
--- a/src/main/java/org/apache/commons/compress/archivers/zip/ZipArchiveOutputStream.java
+++ b/src/main/java/org/apache/commons/compress/archivers/zip/ZipArchiveOutputStream.java
@@ -87,11 +87,6 @@ public class ZipArchiveOutputStream extends ArchiveOutputStream<ZipArchiveEntry>
         protected long zipSplitSize;
 
         /**
-         * The options specifying how the file is opened.
-         */
-        protected OpenOption[] options;
-
-        /**
          * Whether this stream automatically compresses entries using registered compressor factories.
          */
         protected boolean autoCompress;
@@ -101,19 +96,6 @@ public class ZipArchiveOutputStream extends ArchiveOutputStream<ZipArchiveEntry>
          */
         protected AbstractBuilder() {
             setCharset(StandardCharsets.UTF_8);
-        }
-
-        /**
-         * Sets the options specifying how the file is opened.
-         *
-         * <p>Null by default.</p>
-         *
-         * @param options the opem options.
-         * @return {@code this} instance.
-         */
-        public B setOptions(OpenOption... options) {
-            this.options = options;
-            return asThis();
         }
 
         /**
@@ -180,10 +162,12 @@ public class ZipArchiveOutputStream extends ArchiveOutputStream<ZipArchiveEntry>
      * and writes it to the underlying StreamCompressor as raw (already-compressed) bytes.
      */
     private class CompressorBridgeOutputStream extends OutputStream {
+        private final byte[] oneByte = new byte[1];
+
         @Override
         public void write(final int b) throws IOException {
-            final byte[] buf = { (byte) b };
-            write(buf, 0, 1);
+            oneByte[0] = (byte) (b & ZipConstants.BYTE_MASK);
+            write(oneByte, 0, 1);
         }
 
         @Override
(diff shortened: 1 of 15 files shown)


src/main/java/org/apache/commons/compress/archivers/zip/ZipArchiveOutputStream.java
```diff
@@ -929,7 +1134,7 @@ public void finish() throws IOException {
         }
 
         if (entry != null) {
-            throw new ArchiveException("This archive contains unclosed entries.");
+            closeArchiveEntry();
```
mehmet-karaman, line 1116: To keep the API compatible with older clients, wouldn’t it be better to execute this only when the isAutoCompressed flag is set and otherwise still throw the exception?

src/test/java/org/apache/commons/compress/archivers/ArchiveOutputStreamTest.java
```diff
@@ -61,10 +61,15 @@ private void doCallSequence(final String archiveType) throws Exception {
         aos2.putArchiveEntry(aos2.createArchiveEntry(dummy, "dummy"));
         aos2.write(dummy);
 
-        // TODO check if second putArchiveEntry() can follow without closeAE?
```
mehmet-karaman, line 64: Ah.. sorry didn't saw that there is a todo.. Might be also a good time to fix this..

src/main/java/org/apache/commons/compress/archivers/zip/ZipArchiveOutputStream.java
mehmet-karaman: We can close this.. There was a todo in the ArchiveOutputStreamTest which is resolved by this change, so it was somehow planned to do..

src/test/java/org/apache/commons/compress/archivers/ArchiveOutputStreamTest.java
mehmet-karaman: This one can also be closed.

src/main/java/org/apache/commons/compress/archivers/zip/ZipArchiveOutputStream.java
uvoigt: It is tested in [testAutoCompressZstd(ZipMethod zipMethod)](https://github.com/apache/commons-compress/pull/765/changes#diff-540c440c3632bdafdb31449fccac357441aae5f8fd8689e8a17922a4e6caf7bdR43) for `auto-compress` and in [testAutoCompressFalsePreservesLegacyBehavior()](https://github.com/apache/commons-compress/pull/765/changes#diff-540c440c3632bdafdb31449fccac357441aae5f8fd8689e8a17922a4e6caf7bd) for user specified compressing.

mehmet-karaman:
Hi @garydgregory,

I hope you’re doing well. Just a quick friendly reminder before this gets out of scope :)

Thanks in advance!

garydgregory:
@uvoigt 

It's not forgotten. There are other tasks that are higher priority than new features 😉

mehmet-karaman:
Hi @garydgregory,

I just wanted to kindly follow up on this 🙂. Is there any chance we could move this PR forward?

Best regards,  
Mehmet

garydgregory:
Hello @mehmet-karaman 
Thank you for your patience. This is on my list. I am evaluating other PRs and issues. TY!

src/main/java/org/apache/commons/compress/archivers/zip/ZipArchiveOutputStream.java
```diff
@@ -66,6 +71,111 @@
...
+         *
+         * @param autoCompress {@code true} to automatically compress entries, {@code false} otherwise.
+         * @return {@code this} instance.
+         */
+        public B setAutoCompress(final boolean autoCompress) {
+            this.autoCompress = autoCompress;
+            return asThis();
+        }
+
+        @Override
```
garydgregory, line 127: @uvoigt 
Why is this method needed if all it does is call super?

uvoigt: @garydgregory 
This is because the super method is declared as `protected` and is therefore inaccessible within the `ZipArchiveOutputStream` constructor. Overriding it (without changing the access modifiers) makes it accessible there.

PR #770, opened 2026-04-23
[COMPRESS-723] TAR PAX header parsing: enforce memory bound to mitigate resource exhaustion from oversized headers

subbudvk:
Enforce a configurable memory bound on PAX header parsing via a new `maxPaxHeaderSize` builder option on `TarArchiveInputStream` and `TarFile`. The default is 10 MB (`TarConstants.DEFAULT_MAX_PAX_HEADER_SIZE`), enforced through the existing `MemoryLimitException.checkBytes()` mechanism. This closes the last unbounded allocation surface in the TAR parsing pipeline and follows the same defense-in-depth pattern already established for entry names and 7z headers.


subbudvk:
@garydgregory

PR #775, opened 2026-06-17
[COMPRESS-724] Fix TAR PAX 1.x sparse header alignment

OldTruckDriver:
## Summary

Fix `TarUtils.parsePAX1XSparseHeaders(InputStream, int)` so it does not skip an extra tar record when GNU PAX 1.x sparse headers end exactly on a record boundary.

## Details

Previously, the padding calculation used `recordSize - bytesRead % recordSize`. When `bytesRead` was already aligned to `recordSize`, this computed a full-record skip instead of `0`, advancing the stream past real entry data.

This change uses modulo padding calculation so aligned sparse headers leave the stream positioned at the first data byte.

## Tests

- `mvn -q -Dtest=org.apache.commons.compress.archivers.tar.TarUtilsTest test`
- `mvn -q`

### DIFF
--- a/src/test/java/org/apache/commons/compress/archivers/tar/TarUtilsTest.java
+++ b/src/test/java/org/apache/commons/compress/archivers/tar/TarUtilsTest.java
@@ -109,6 +109,16 @@ class TarUtilsTest extends AbstractTest {
         return TarUtils.parsePaxHeaders(new ByteArrayInputStream(data), globalPaxHeaders, data.length, Short.MAX_VALUE, sparseHeaders);
     }
 
+    private static byte[] pax1xSparseHeaderAlignedToSingleRecord() {
+        final StringBuilder header = new StringBuilder("100\n");
+        for (int i = 0; i < 200; i++) {
+            header.append(i < 108 ? "00\n" : "0\n");
+        }
+        final byte[] bytes = header.toString().getBytes(UTF_8);
+        assertEquals(512, bytes.length);
+        return bytes;
+    }
+
     static Stream<Arguments> testReadLongNameHandlesLimits() {
         final String empty = "";
         final String ntfsLongName = createNtfsLongNameByUtf16Units(32767);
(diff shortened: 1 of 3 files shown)


garydgregory (changes requested):
Hello @OldTruckDriver 

Thank you for the PR.


The tests contain too many magic numbers.

For example:
```java
        final byte[] fileData = new byte[513];
        Arrays.fill(fileData, 0, 512, (byte) 'A');
        fileData[512] = 'B';
```
Above 513 and 512 are clearly related.

In `pax1xSparseHeaderAlignedToSingleRecord`, why 200 and 108. Either use local vars with clear names or write comments.

TY!

src/test/java/org/apache/commons/compress/archivers/tar/TarUtilsTest.java
```diff
@@ -328,6 +338,19 @@ void testParsePAX1XSparseHeaders() throws Exception {
         }
     }
 
+    @Test
+    void testParsePAX1XSparseHeadersDoesNotSkipAlignedDataRecord() throws Exception {
+        final byte[] fileData = new byte[513];
+        Arrays.fill(fileData, 0, 512, (byte) 'A');
```
garydgregory: Hello @OldTruckDriver 

What are all these magic numbers?

Is 512 supposed to be `TarConstants.DEFAULT_RCDSIZE` or something else?

Are all the 512 values the same as a coincidence?

Is 513 supposed to be `TarConstants.DEFAULT_RCDSIZE + 1`?

You get the idea...

Please use meaningful local variable names or references.

garydgregory (changes requested):
@OldTruckDriver 
Please see my comment.

garydgregory:
@OldTruckDriver ping 🔔 See comments from myself and copiolot.

OldTruckDriver:
Thanks for the review @garydgregory. You're right, those numbers should be named. To clarify the intent:

1. 512 is TarConstants.DEFAULT_RCDSIZE and 513 is DEFAULT_RCDSIZE + 1 — the extra byte is what exposes the bug: with the old formula, the reader skips a whole extra record and lands on that trailing byte instead of the first data byte.
2. 100 is the sparse entry count, 200 is the resulting number of offset/size lines, and 108 is only there to pad the header to exactly one record so it's already aligned.
I'll push a commit replacing these with named constants/comments shortly. Sorry for the late reply.

garydgregory:
@OldTruckDriver Thank you, merged 🚀

PR #776, opened 2026-06-28
[COMPRESS-727] Provide a symlink-resistant and concurrency-safe archive Extractor

tomasilluminati:
Adds a security-first Extractor under archivers. extractor with a SecureDirectoryStream-based race-safe path on Linux and a best-effort base extractor elsewhere. Symlink and special-file policies, safe hard-link resolution, tests and a changes.xml entry are included.

### DIFF
--- a/src/main/java/org/apache/commons/compress/archivers/extractor/Extractor.java
+++ b/src/main/java/org/apache/commons/compress/archivers/extractor/Extractor.java
@@ -237,6 +237,9 @@ public class Extractor {
         if (resolved.equals(rootDirectory)) {
             return null;
         }
+        // A component of only dots or spaces (for example "...") is stripped by Windows and can alias the root, which this
+        // lexical equality does not catch. The entry still stays within the root and its no-follow CREATE_NEW write fails
+        // closed rather than replacing the root, so this is a known corner case rather than a boundary escape.
         if (!isWithinRoot(resolved)) {
             throw new ArchiveException("Entry '%s' would escape the extraction root", name);
         }
@@ -258,9 +261,14 @@ public class Extractor {
             final BasicFileAttributes attrs;
             try {
                 attrs = Files.readAttributes(current, BasicFileAttributes.class, LinkOption.NOFOLLOW_LINKS);
-            } catch (final NoSuchFileException e) {
-                Files.createDirectory(current);
-                continue;
+            } catch (final IOException e) {
+                // readAttributes documents only IOException, and NoSuchFileException is an optional specific exception, so
+                // confirm the component is genuinely missing before creating it and rethrow any other I/O error.
+                if (e instanceof NoSuchFileException || Files.notExists(current, LinkOption.NOFOLLOW_LINKS)) {
+                    Files.createDirectory(current);
+                    continue;
+                }
+                throw e;
             }
             if (attrs.isSymbolicLink()) {
                 throw new ArchiveException("Refusing to traverse symbolic link: %s", current);
--- a/src/main/java/org/apache/commons/compress/archivers/extractor/Extractor.java
+++ b/src/main/java/org/apache/commons/compress/archivers/extractor/Extractor.java
@@ -1,0 +1,388 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.commons.compress.archivers.extractor;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.nio.channels.Channels;
+import java.nio.file.DirectoryStream;
+import java.nio.file.Files;
+import java.nio.file.InvalidPathException;
+import java.nio.file.LinkOption;
+import java.nio.file.NoSuchFileException;
+import java.nio.file.Path;
+import java.nio.file.SecureDirectoryStream;
+import java.nio.file.StandardOpenOption;
+import java.nio.file.attribute.BasicFileAttributes;
+import java.util.Enumeration;
+import java.util.Objects;
+
+import org.apache.commons.compress.archivers.ArchiveEntry;
+import org.apache.commons.compress.archivers.ArchiveException;
+import org.apache.commons.compress.archivers.ArchiveInputStream;
+import org.apache.commons.compress.archivers.tar.TarArchiveEntry;
+import org.apache.commons.compress.archivers.tar.TarFile;
+import org.apache.commons.compress.archivers.zip.ZipArchiveEntry;
+import org.apache.commons.compress.archivers.zip.ZipFile;
+import org.apache.commons.io.IOUtils;
+
+/**
+ * Safely materializes an archive into a target directory, including symbolic links, while resisting both symlink-slip from
+ * the archive itself and concurrent symlink races from third parties.
+ * <p>
+ * Obtain an instance with {@link #newExtractor(Path)}, which returns the strongest implementation the platform supports: a
+ * race-safe extractor backed by {@link SecureDirectoryStream} where the file system provides one (Linux), otherwise this
+ * best-effort base implementation. The base implementation resolves every path component with
+ * {@link LinkOption#NOFOLLOW_LINKS} and writes regular files {@code CREATE_NEW}, which stops symlink-slip from the archive and
+ * silent overwrites, but it cannot fully close a concurrent time-of-check-to-time-of-use race; that residual is documented
+ * and only the secure implementation removes it.
+ * </p>
+ * <p>
+ * Defaults are conservative: symbolic-link entries are rejected ({@link SymlinkPolicy#REJECT}), special entries are skipped
+ * ({@link SpecialFilePolicy#SKIP}), and existing files are not overwritten. Instances are not thread-safe: configure an
+ * instance before extracting and do not share it across concurrent extractions.
+ * </p>
+ *
+ * @since 1.29.0
+ */
+public class Extractor {
+
+    enum EntryType {
+        FILE, DIRECTORY, SYMLINK, HARD_LINK, SPECIAL
+    }
+
+    /**
+     * Creates an extractor for {@code targetDirectory}, returning a race-safe implementation when the platform's file system
+     * provider exposes a {@link SecureDirectoryStream} (Linux), otherwise a best-effort implementation.
+     *
+     * @param targetDirectory an existing directory into which archives are extracted; resolved to its real path.
+     * @return an extractor bound to {@code targetDirectory}.
+     * @throws IOException if {@code targetDirectory} does not exist or is not a directory.
+     * @since 1.29.0
+     */
+    public static Extractor newExtractor(final Path targetDirectory) throws IOException {
+        final Path root = Objects.requireNonNull(targetDirectory, "targetDirectory").toRealPath();
+        if (!Files.isDirectory(root)) {
+            throw new IOException("Target is not a directory: " + targetDirectory);
+        }
+        try (DirectoryStream<Path> stream = Files.newDirectoryStream(root)) {
+            if (stream instanceof SecureDirectoryStream) {
+                return new SecureExtractor(root);
+            }
+        }
+        return new Extractor(root);
+    }
+
+    /** The validated, canonical extraction root. */
+    final Path rootDirectory;
+
+    private SymlinkPolicy symlinkPolicy = SymlinkPolicy.REJECT;
+
+    private SpecialFilePolicy specialFilePolicy = SpecialFilePolicy.SKIP;
+
+    private boolean overwrite;
+
+    private Runnable beforeLeafWrite;
+
+    Extractor(final Path rootDirectory) {
+        this.rootDirectory = rootDirectory;
+    }
+
+    private static EntryType classify(final ArchiveEntry entry) {
+        if (entry instanceof TarArchiveEntry) {
+            final TarArchiveEntry tar = (TarArchiveEntry) entry;
+            if (tar.isSymbolicLink()) {
+                return EntryType.SYMLINK;
+            }
+            if (tar.isLink()) {
+                return EntryType.HARD_LINK;
+            }
+            if (tar.isBlockDevice() || tar.isCharacterDevice() || tar.isFIFO()) {
+                return EntryType.SPECIAL;
+            }
+            return tar.isDirectory() ? EntryType.DIRECTORY : EntryType.FILE;
+        }
+        if (entry instanceof ZipArchiveEntry && ((ZipArchiveEntry) entry).isUnixSymlink()) {
+            return EntryType.SYMLINK;
+        }
+        return entry.isDirectory() ? EntryType.DIRECTORY : EntryType.FILE;
+    }
+
+    /**
+     * Extracts the entries of an {@link ArchiveInputStream}.
+     * <p>
+     * Zip unix symbolic links are recognized only through {@link #extract(ZipFile)}: their unix mode lives in the central
+     * directory, which a stream does not expose, so a zip symbolic link presented through a stream is materialized as a
+     * regular file (no-follow, create-new) rather than a link.
+     * </p>
+     *
+     * @param archive the archive to extract; not closed by this method.
+     * @throws IOException if extraction fails or a policy rejects an entry.
+     * @since 1.29.0
+     */
+    public void extract(final ArchiveInputStream<?> archive) throws IOException {
+        ArchiveEntry entry;
+        while ((entry = archive.getNextEntry()) != null) {
+            if (!archive.canReadEntryData(entry)) {
+                continue;
+            }
+            final EntryType type = classify(entry);
+            String linkTarget = null;
+            if (type == EntryType.SYMLINK || type == EntryType.HARD_LINK) {
+                // Only tar carries link metadata through a stream: a zip unix symbolic link is recognized solely from the
+                // central directory (see extract(ZipFile)), which a stream does not expose, so classify never reports a
+                // streamed entry as a link unless it is a TarArchiveEntry.
+                linkTarget = ((TarArchiveEntry) entry).getLinkName();
+            }
+            process(entry.getName(), type, linkTarget, type == EntryType.FILE ? archive : null);
+        }
+    }
+
+    /**
+     * Extracts the entries of a {@link TarFile}.
+     *
+     * @param archive the archive to extract; not closed by this method.
+     * @throws IOException if extraction fails or a policy rejects an entry.
+     * @since 1.29.0
+     */
+    public void extract(final TarFile archive) throws IOException {
+        for (final TarArchiveEntry entry : archive.getEntries()) {
+            final EntryType type = classify(entry);
+            if (type == EntryType.FILE) {
+                try (InputStream content = archive.getInputStream(entry)) {
+                    process(entry.getName(), type, null, content);
+                }
+            } else {
+                final String linkTarget = type == EntryType.SYMLINK || type == EntryType.HARD_LINK ? entry.getLinkName() : null;
+                process(entry.getName(), type, linkTarget, null);
+            }
+        }
+    }
+
+    /**
+     * Extracts the entries of a {@link ZipFile}.
+     *
+     * @param archive the archive to extract; not closed by this method.
+     * @throws IOException if extraction fails or a policy rejects an entry.
+     * @since 1.29.0
+     */
+    public void extract(final ZipFile archive) throws IOException {
+        final Enumeration<ZipArchiveEntry> entries = archive.getEntries();
+        while (entries.hasMoreElements()) {
+            final ZipArchiveEntry entry = entries.nextElement();
+            if (!archive.canReadEntryData(entry)) {
+                continue;
+            }
+            final EntryType type = classify(entry);
+            if (type == EntryType.FILE) {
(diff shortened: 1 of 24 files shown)


tomasilluminati:
@ppkarwasz PR Opened.

tomasilluminati:
@ppkarwasz Fixed the Windows Test Errors.

garydgregory (commented):
Hello @tomasilluminati 
This is interesting and I am hoping we can extract something more generic on this theme.

The background nugget is hidden in Apache Commons Text in the package-private class  [PathFence](https://github.com/apache/commons-text/blob/master/src/main/java/org/apache/commons/text/lookup/PathFence.java).

I extracted the class for addition in Apache Commons IO in [PR 805](https://github.com/apache/commons-io/pull/805) but it might not be good enough. The IO PR is something I had started to iterate with @ppkarwasz but other things popped up since and now seems like a good time to revist.

There are two concerns: resolving relative paths and dealing with various types of file system links. Handling the relative path issue is "simple", the links less so. At first glance it looks like you've covered many cases here.

I'm not sure if both concerns should be intermingled in one class or composed.

Can you see a PathFence class as something that can be improved on and reused here?

The PathFence could also be reused in Apache Commons Configuration for example.

Looking forward to your thoughts.

TY!

tomasilluminati:
@garydgregory 
Thanks, this is a great direction, I'm keen to reuse PathFence instead of carrying my own containment code.

The syntactic part is a clean fit. toAbsolutePath().normalize() plus the within-root check is generic, and Compress, Configuration and Text can all share it as-is.

Links are the tricky bit, and I'd lean toward composing rather than folding them in. Part of why they resist living in a path predicate is that during extraction the entry doesn't exist yet, so I can't just toRealPath() the target. The real protection ends up being operational, resolve each component no-follow, create with CREATE_NEW so an existing link is never followed, and on Linux write relative to the directory handle (SecureDirectoryStream) to close the TOCTOU window. That's bound to the act of creating files, not to judging a path.

Where @ppkarwasz point lands well is the already-on-disk case, like a trusted config dir. There a resolving check (toRealPath then the same within-root test) makes sense, and that could be an optional mode in PathFence. The extractor would lean on the syntactic decision and keep its no-follow creation layer on top.

So my vote is composed. PathFence owns the containment decision, extraction-specific link safety sits around it, and PathFence stays light enough for Configuration to reuse.

Happy to dig into the IO PR with you and Piotr whenever it's a good time, and to bring over the link cases I already cover as tests. Thanks.

src/main/java/org/apache/commons/compress/archivers/extractor/Extractor.java
```diff
@@ -0,0 +1,388 @@
...
+     * target that resolves outside the root is rejected; under {@link SymlinkPolicy#ALLOW_ALL} the link is created verbatim.
+     */
+    void writeSymbolicLink(final Path leaf, final String linkTarget) throws IOException {
+        if (symlinkPolicy == SymlinkPolicy.ALLOW_WITHIN_ROOT) {
+            final Path parent = leaf.getParent();
+            final Path resolved = (parent != null ? parent : rootDirectory).resolve(linkTarget).normalize();
+            if (!resolved.startsWith(rootDirectory)) {
+                throw new ArchiveException("Symbolic link target escapes the extraction root: %s -> %s", leaf, linkTarget);
+            }
+        }
```
Marcono1234, line 417: It looks like this `ALLOW_WITHIN_ROOT` validation can be circumvented, but fixing this might be pretty difficult or impossible.
See https://issues.apache.org/jira/browse/COMPRESS-727?focusedCommentId=18092597#comment-18092597

src/main/java/org/apache/commons/compress/archivers/extractor/Extractor.java
```diff
@@ -0,0 +1,388 @@
...
+    }
+
+    /**
+     * Resolves {@code name} against the extraction root and rejects any result that escapes it (the lexical zip-slip guard).
+     */
+    private Path resolveWithinRoot(final String name) throws ArchiveException {
+        final Path resolved = rootDirectory.resolve(name).normalize();
+        if (!resolved.startsWith(rootDirectory)) {
+            throw new ArchiveException("Entry '%s' would escape the extraction root", name);
+        }
```
Marcono1234, line 245: Maybe this should also handle the case where `resolved.equals(rootDirectory)` (e.g. for entry names like `/`, `.`, `a/..`).

There are multiple ways in which this could go wrong / could be exploitable:
- Depending on the `Files` API being used, this could lead to `rootDirectory` being deleted and replaced with a regular file instead
- If this is ever extended to also apply file permissions to created files and directories, it could change the permissions of `rootDirectory`

Currently this implementation is not affected by this, but it is conceivable that it becomes affected in case of future refactoring.
Therefore would be good to explicitly handle this case.

Might be best to throw an exception as well?
Except for `/` which might redundantly exist in some non-malicious ZIP file? But in that case the caller of `resolveWithinRoot` would have to skip processing. Maybe could use a `Optional` / nullable return type here and use _empty_ / null for `/` and let the caller skip it.

tomasilluminati:
@Marcono1234 thanks, both are good calls.

On ALLOW_WITHIN_ROOT, you're right. The lexical normalize() check can't model how the OS resolves .. after a symlink component, and with the forward-reference ordering even a real-path check at creation wouldn't catch it, since the second link doesn't exist yet when the first is validated. So that mode can't promise every created link stays inside the root once they are followed together. I'd document it as best-effort and lexical, calling out that it cannot prevent escapes via chained or indirect symlinks, with REJECT (the default) or SKIP as the safe choices for untrusted archives. Worth noting that extraction itself never follows these links (the write path resolves every component no-follow), so nothing is written outside the root during extraction. The hazard is a created link that escapes when something follows it later, and the docs should say so plainly. If a hard guarantee is ever needed, the only way I see is a post-extraction pass that toRealPath-validates every created symlink and drops the escapers, which could be a future opt-in.

On resolveWithinRoot and resolved.equals(rootDirectory), agreed, better to be explicit than to rely on the downstream ops being safe today. Entries that resolve to the root itself (/, ., a/.. and friends) carry nothing to materialize, so I'd skip them rather than target the root path. I'll have resolveWithinRoot signal that case and let the caller skip, so a benign / entry doesn't fail the extraction and nothing ever writes to or replaces the root, plus tests for those names.

Before I push this, @garydgregory @ppkarwasz does this direction sound right to you (document ALLOW_WITHIN_ROOT as best-effort, and skip entries that resolve to the root)? If you are good with it, I'll proceed.

garydgregory:
> @garydgregory Thanks, this is a great direction, I'm keen to reuse PathFence instead of carrying my own containment code.
> 
> The syntactic part is a clean fit. toAbsolutePath().normalize() plus the within-root check is generic, and Compress, Configuration and Text can all share it as-is.
> 
> Links are the tricky bit, and I'd lean toward composing rather than folding them in. Part of why they resist living in a path predicate is that during extraction the entry doesn't exist yet, so I can't just toRealPath() the target. The real protection ends up being operational, resolve each component no-follow, create with CREATE_NEW so an existing link is never followed, and on Linux write relative to the directory handle (SecureDirectoryStream) to close the TOCTOU window. That's bound to the act of creating files, not to judging a path.
> 
> Where @ppkarwasz point lands well is the already-on-disk case, like a trusted config dir. There a resolving check (toRealPath then the same within-root test) makes sense, and that could be an optional mode in PathFence. The extractor would lean on the syntactic decision and keep its no-follow creation layer on top.
> 
> So my vote is composed. PathFence owns the containment decision, extraction-specific link safety sits around it, and PathFence stays light enough for Configuration to reuse.
> 
> Happy to dig into the IO PR with you and Piotr whenever it's a good time, and to bring over the link cases I already cover as tests. Thanks.

Let's iron out the PathFence in Commons IO if you are available. I'd love to get that out and reuse it as it fits best. If we need a separate link utility, that sounds like a good Commons IO addition as well. Then we could compose in Compress, Configuration, Text, and maybe others like JEXL.
WDYT?

src/main/java/org/apache/commons/compress/archivers/extractor/Extractor.java
```diff
@@ -0,0 +1,388 @@
...
+        Path current = rootDirectory;
+        for (final Path component : rootDirectory.relativize(directory)) {
+            current = current.resolve(component);
+            final BasicFileAttributes attrs;
+            try {
+                attrs = Files.readAttributes(current, BasicFileAttributes.class, LinkOption.NOFOLLOW_LINKS);
+            } catch (final NoSuchFileException e) {
+                Files.createDirectory(current);
+                continue;
+            }
```
Marcono1234, line 272: `Files#readAttributes` does not document that it can throw `NoSuchFileException`, and in general `NoSuchFileException` is an "optional specific exception" and might not be reported by all file system implementations (?). So this seems to rely on implementation details.

Maybe this should do `catch (IOException e)` and then when `if (e instanceof NoSuchFileException || Files.notExists(current, LinkOption.NOFOLLOW_LINKS)) { ... create directory }` and otherwise rethrow `e`?

src/main/java/org/apache/commons/compress/archivers/extractor/Extractor.java
Marcono1234: Hmm, just checking `resolved.equals(rootDirectory)` does not seem enough:
It seems Windows has some quirks with leading spaces and trailing dots, for example the name `" ."`:
- is not removed by `normalize()`
- effectively acts like `.`
- causes a trailing `\` _in the path name_ when obtaining the last name element
    ```java
    var dir = Path.of(".").toAbsolutePath();
    var bogus = dir.resolve(" .").toRealPath();
    System.out.println(bogus.getName(bogus.getNameCount() - 1));
    ```

tomasilluminati:
@Marcono1234 thanks, both were right, addressed both.

ALLOW_WITHIN_ROOT: you're right that it can't be a hard guarantee, so I rewrote the constant's Javadoc to say plainly it is best-effort and lexical (it does not stop a link reached through another link, or an in-root chain that escapes only once the OS resolves it), with REJECT (default) and SKIP called out as the safe modes for untrusted input, plus a new "Safe Extraction" section in the user guide. Your indirect-symlink PoC is now a test: blocked under REJECT and SKIP, and under ALLOW_WITHIN_ROOT/ALLOW_ALL the links are created but nothing is written outside the root during extraction.

resolveWithinRoot: it now returns null when the entry resolves to the root itself and the caller skips it, so a benign . or a/.. no longer fails and nothing ever targets or replaces the root. I kept it a nullable Path rather than Optional to match the codebase, escapes still throw, and I folded the four containment checks into one isWithinRoot helper with a sibling-prefix test.

Pushed. Thanks for the careful review.

src/test/java/org/apache/commons/compress/archivers/extractor/ExtractorRootBoundaryTest.java
```diff
@@ -0,0 +1,178 @@
...
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+
+/**
+ * Root-boundary invariant: {@code newExtractor} canonicalizes the target with {@code toRealPath}, so the containment guard is
+ * not sensitive to how the target is spelled (trailing separator, {@code .} component, or a symlinked path). This is why the
+ * extractor uses {@code resolveWithinRoot} against a canonical root rather than the lexical {@code ArchiveEntry.resolveIn},
+ * which compares against the parent string as given and misbehaves when that parent is not normalized.
+ */
+class ExtractorRootBoundaryTest {
```
Marcono1234, line 51: For completeness maybe this needs also a test where `target = Path.of("..")`.

If Extractor did not call `toRealPath`, then such a target path would render escape detection ineffective because `normalize` would not remove `..` entries from the path.

This is similar to the `toAbsolutePath().normalize()` concerns which you had already mentioned before and which were mentioned in the Commons IO PR; though the use of `toRealPath` here solves this already.

(Not sure if this is actually covered already by the tests and I just overlooked it.)

src/main/java/org/apache/commons/compress/archivers/extractor/Extractor.java
Marcono1234: Did you resolve this comment about `Files#readAttributes` intentionally?

src/main/java/org/apache/commons/compress/archivers/extractor/Extractor.java
```diff
@@ -217,11 +217,27 @@ final void setBeforeLeafWrite(final Runnable hook) {
...
+     *
+     * @return the resolved path within the root, or {@code null} if {@code name} resolves to the root itself (for example
+     *         {@code a/..}), which carries nothing to materialize; the caller skips such entries rather than writing at or
+     *         replacing the root.
+     * @throws ArchiveException if the resolved path escapes the extraction root.
      */
     private Path resolveWithinRoot(final String name) throws ArchiveException {
         final Path resolved = rootDirectory.resolve(name).normalize();
-        if (!resolved.startsWith(rootDirectory)) {
+        if (resolved.equals(rootDirectory)) {
```
Marcono1234, line 237: On Windows it seems this fails to detect to file names consisting only of spaces and dots, see https://github.com/apache/commons-compress/pull/776#discussion_r3507869476.
The paths are not equal but actually refer to the same directory.

Not sure if it is worth accounting for that corner case. It seems `Files.isSameFile(resolved, rootDirectory)` could be used as fallback, but it probably has overhead.

(Side note: I am planning to submit this as OpenJDK bug as well; but it does not help users with older Java versions.)

Marcono1234 (commented):
Thanks for addressing my comments so far; I am glad that they were useful!

Though I also don't want to disrupt this PR or the design decisions too much, so please let me know (you or the maintainers) if my comments are not helpful, or not helpful at this point.

tomasilluminati:
> Thanks for addressing my comments so far; I am glad that they were useful!
> 
> Though I also don't want to disrupt this PR or the design decisions too much, so please let me know (you or the maintainers) if my comments are not helpful, or not not helpful at this point.

@Marcono1234 they were very helpful, please keep them coming. The indirect-symlink PoC in particular made the best-effort caveat concrete instead of hand-wavy, and the root-resolving case was a real gap. Reviews like that are exactly what this kind of code needs, so no worry about disrupting, I'd rather get it right.

One heads up for you and the maintainers: I'm recovering from a hand surgery, so I may be a bit slower to respond for a little while. Still on it, just setting expectations. Thanks again.

src/test/java/org/apache/commons/compress/archivers/extractor/ExtractorRootBoundaryTest.java
tomasilluminati: Good call, added canonicalizesTargetWithParentComponentAndStaysContained. I anchored the ".." to a real in-root subdir (target/sub/..) rather than a bare Path.of(".."), since the bare relative one would resolve against the process working directory and extract outside the temp dir. toRealPath collapses it back to the canonical root as you said, and the test pins both halves: a normal entry lands in the real target, and ../escape is still rejected.

tomasilluminati:
@Marcono1234 Sorry about the resolve on the Files#readAttributes thread, that was me tidying the view by accident, not dismissing your point. I've reopened it. You're right that readAttributes only documents IOException and NoSuchFileException is an optional specific exception, so relying on it is fragile. Changed it to catch IOException and confirm with Files.notExists(current, NOFOLLOW_LINKS) before creating, kept e instanceof NoSuchFileException as the fast path, and rethrow anything else, so a real error like denied access (where notExists returns false) is no longer masked as "missing". Thanks for catching this.

src/main/java/org/apache/commons/compress/archivers/extractor/Extractor.java
tomasilluminati: I traced it, on Windows a name of only dots or spaces (like "...") aliases the root and the lexical equals here does not see it, but it stays contained, the entry never resolves outside the root, and the no-follow CREATE_NEW write hits the existing root directory and throws rather than replacing it, so it fails closed. I documented it as a known corner case instead of a per-entry Files.isSameFile, which would be two extra syscalls on every entry to guard something already fail-closed and not an escape. If you'd rather have these rejected outright, a cheap lexical reject of pure dot/space names (no I/O) would do it, happy to add that if you prefer it in this PR. And please link your OpenJDK issue here when it is filed, I'll reference it. @garydgregory @ppkarwasz

tomasilluminati:
> > @garydgregory Thanks, this is a great direction, I'm keen to reuse PathFence instead of carrying my own containment code.
> > The syntactic part is a clean fit. toAbsolutePath().normalize() plus the within-root check is generic, and Compress, Configuration and Text can all share it as-is.
> > Links are the tricky bit, and I'd lean toward composing rather than folding them in. Part of why they resist living in a path predicate is that during extraction the entry doesn't exist yet, so I can't just toRealPath() the target. The real protection ends up being operational, resolve each component no-follow, create with CREATE_NEW so an existing link is never followed, and on Linux write relative to the directory handle (SecureDirectoryStream) to close the TOCTOU window. That's bound to the act of creating files, not to judging a path.
> > Where @ppkarwasz point lands well is the already-on-disk case, like a trusted config dir. There a resolving check (toRealPath then the same within-root test) makes sense, and that could be an optional mode in PathFence. The extractor would lean on the syntactic decision and keep its no-follow creation layer on top.
> > So my vote is composed. PathFence owns the containment decision, extraction-specific link safety sits around it, and PathFence stays light enough for Configuration to reuse.
> > Happy to dig into the IO PR with you and Piotr whenever it's a good time, and to bring over the link cases I already cover as tests. Thanks.
> 
> Let's iron out the PathFence in Commons IO if you are available. I'd love to get that out and reuse it as it fits best. If we need a separate link utility, that sounds like a good Commons IO addition as well. Then we could compose in Compress, Configuration, Text, and maybe others like JEXL. WDYT?

@garydgregory

I'm in, happy to help with you and @ppkarwasz.

On "one class or composed", I'd compose and keep PathFence a purely lexical predicate (toAbsolutePath().normalize() then startsWith, no filesystem access), so it stays trivially reusable for Configuration, Text and JEXL.

One thing I'd make explicit in its contract though: lexical containment alone does not stop symlink traversal. normalize() never touches the filesystem, so a symlinked component slips straight through. It should therefore state plainly that PathFence is a syntactic gate, safe on its own only for trusted or link-free trees, and must be paired with real resolution for untrusted input.

That resolution is the separate utility you mentioned. Refining my own earlier comment, I'd keep the toRealPath check in the utility rather than as a mode inside PathFence, so PathFence never touches the filesystem. It carries the two modes the cases need:

- toRealPath then containment for the trusted on-disk case (the CATALINA_BASE conf symlink). It resolves links but needs the path to exist, so it fits reads, not targets that don't exist yet.
- no-follow resolution for untrusted input and for creation, never following a link, which is what the Extractor already does.

That puts the security-sensitive resolution in one place and keeps PathFence small enough to reuse everywhere.

If that sounds right to you, I can start there.

Meanwhile I'll keep #776 moving on @Marcono1234 two points, neither depends on PathFence.

src/test/java/org/apache/commons/compress/archivers/extractor/ExtractorRootBoundaryTest.java
Marcono1234: Hmm, ok but by using `target/sub/..` as output dir it does not cover the case anymore that I mentioned. Because then the test would most likely still succeed even if the implementation erroneously just used `normalize()` without `toRealPath()` or `toAbsolutePath()`.

To cover this, I think you would really need `Path.of("..")` (without any parent directory). But that would indeed not be ideal for this test here. Maybe at least the failing case could be covered though, where it is expected that an exception is thrown and no files or directories are created. And maybe give the escaping file a distinct name (e.g. `../commons-compress-escape.txt`) so that if the test unexpectedly fails and escape happens, the developer does not wonder about random files in the parent directory.

Not sure though if that would be a clean enough solution, and if this is really worth it then.

tomasilluminati: Good catch, my test wasn't actually asserting much, normalize() was collapsing target/sub/.. upfront, so it passed regardless. I swapped it for a bare Paths.get("..") target with an entry named ../commons-compress-escape.txt. Confirmed it fails if toRealPath() gets removed, so it's definitely doing its job now

https://github.com/apache/commons-compress/pull/776/changes/67e98269fbeaddef81115428d5f38d7490b580fd

src/main/java/org/apache/commons/compress/archivers/extractor/Extractor.java
Marcono1234: It seems my suggestion to use `isSameFile` is flawed / risky because it accesses the file system. So it should only occur as additional check, after a `Path#startsWith` check passed (which does not access the file system).

Otherwise I am worried that merely by calling `isSameFile` a Windows UNC path could leak information to a remote server.

Though as you mention, it might be desirable to avoid `isSameFile` completely (and I guess that is fine?).

However, either way, maybe it would be good to add a Windows-only test where a ZIP entry has a Windows UNC path as name (e.g. `\\localhost\invalid`) and ensure that the reported error is from the ZIP Slip protection and not an error like `FileSystemException: \\localhost\invalid\: The network name cannot be found`, which would indicate that some API was used which actually tried to perform a file system access _before_ they ZIP Slip check.

What do you think?
(though it seems the PR here is unfortunately a bit stale, maybe due to lack of time by the maintainers)

tomasilluminati: @Marcono1234 Agreed. The guard never used isSameFile and stays lexical: resolve, normalize, equals and startsWith, with nothing touching the file system before containment. The Javadoc now says so.

Tests added: a host/share name like //localhost/invalid/evil.txt on every platform through ZipFile, the zip stream and tar, plus Windows-only cases for \\localhost\invalid, an absolute drive path and a drive-relative name. Each accepts only the guard's own ArchiveException and an empty target, so a FileSystemException fails it. With isSameFile put in front of the check, the cross-platform cases fail with NoSuchFileException. The malformed-name test now covers the Windows illegal characters and a share-less UNC name.

They pass with a Windows JDK 17 under Wine. The windows-latest leg has not run since the first push, so a maintainer approving the workflow would help. The trailing-dots alias stays documented and fail-closed; the JDK issue is JDK-8248430.

Marcono1234: Thanks! The tests look really good to me.

----

> The trailing-dots alias stays documented and fail-closed; the JDK issue is JDK-8248430.

Thanks for digging this up; yes that was the existing JDK bug report about trailing dots. Though I have privately reported and suggested to the JDK maintainers to completely reject trailing dots on Windows (similar to how trailing whitespace is already rejected). But I am currently still waiting for their evaluation.

PR #779, opened 2026-07-01
Reject out-of-range directory header count in readDirectoryEntry

kali834x:
readDirectoryEntry computes datalen as `TP_SIZE * entry.getHeaderCount()` in int, but the header count comes straight from the tape segment header and isn't bounded at that point, so a directory inode claiming around 2^21 records wraps the product to a negative value that then flows into `raw.read(blockBuffer, 0, datalen)` and surfaces as an `IndexOutOfBoundsException` rather than a normal archive error. getNextEntry already rejects a count `>= CDATA_LEN` and the sibling skip sites in getNextEntry, readBITS and readCLRI cast to long, so this one path is the odd one out; applying the same bound before the multiply keeps a corrupt count from reaching the size math. this is the COMPRESS-712 testPart2 input, whose assertion moves from the raw `IndexOutOfBoundsException` to an `ArchiveException`.

- [x] Write unit tests that match behavioral changes, where the tests fail if the changes to the runtime are not applied.
- [x] Each commit in the pull request should have a meaningful subject line and body.

### DIFF
--- a/src/main/java/org/apache/commons/compress/archivers/dump/DumpArchiveInputStream.java
+++ b/src/main/java/org/apache/commons/compress/archivers/dump/DumpArchiveInputStream.java
@@ -526,12 +526,13 @@ public class DumpArchiveInputStream extends ArchiveInputStream<DumpArchiveEntry>
                 pending.put(entry.getIno(), entry);
             }
 
+            final int headerCount = entry.getHeaderCount();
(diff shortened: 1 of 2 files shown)


garydgregory:
Hello @kali834x 
Is there any reason not to do this on construction of the dump archive entry like https://github.com/apache/commons-compress/pull/781 ?

garydgregory:
Closing: no reply, merging #781 instead.

PR #780, opened 2026-07-02
Reject tar sparse blocks larger than the entry size

kali834x:
buildSparseInputStreams in TarArchiveInputStream and TarFile size the non-hole sub-streams from each sparse header's numbytes, and the only geometry guard, getOrderedSparseHeaders, bounds offset+numbytes against the logical realSize rather than the entry's stored size. a crafted gnu/pax sparse entry can set a small header size but a sparse map whose numbytes sum is much larger (realsize == numbytes so the existing realSize check passes), and reading that entry then pulls bytes belonging to the following entries. this accumulates the physical block bytes while building the streams and fails with an ArchiveException once they exceed currEntry.getSize(). the sum equals the entry size for every valid sparse variant (old gnu, pax 0.x, star), with slack only for the pax 1.x in-data map, so no valid archive is rejected. the added SparseFilesTest case reads a 4-byte entry whose map claims a 1 MiB block and now throws for both readers instead of over-reading into the next entry.

### DIFF
--- a/src/test/java/org/apache/commons/compress/archivers/tar/SparseFilesTest.java
+++ b/src/test/java/org/apache/commons/compress/archivers/tar/SparseFilesTest.java
@@ -53,6 +53,9 @@ import org.junit.jupiter.params.provider.ValueSource;
 
 class SparseFilesTest extends AbstractTest {
 
+    // numbytes claimed by the crafted sparse map, deliberately larger than the entry's stored size
+    private static final long SPARSE_MAP_NUMBYTES = 1024 * 1024;
+
     @TempDir
     private static Path tempDir;
 
@@ -539,9 +542,9 @@ class SparseFilesTest extends AbstractTest {
 
     private static void appendPadded(final ByteArrayOutputStream out, final byte[] data) {
         out.write(data, 0, data.length);
-        final int rem = data.length % 512;
+        final int rem = data.length % TarConstants.DEFAULT_RCDSIZE;
         if (rem != 0) {
-            out.write(new byte[512 - rem], 0, 512 - rem);
+            out.write(new byte[TarConstants.DEFAULT_RCDSIZE - rem], 0, TarConstants.DEFAULT_RCDSIZE - rem);
         }
     }
 
@@ -564,7 +567,7 @@ class SparseFilesTest extends AbstractTest {
     }
 
     private static byte[] ustarHeader(final String name, final long size, final char typeFlag) {
-        final byte[] header = new byte[512];
+        final byte[] header = new byte[TarConstants.DEFAULT_RCDSIZE];
         putField(header, 0, 100, name);
         putOctal(header, 100, 8, 0644);
         putOctal(header, 108, 8, 0);
@@ -591,7 +594,7 @@ class SparseFilesTest extends AbstractTest {
     // A second entry follows, whose bytes must not be reachable through the sparse entry.
     private byte[] sparseArchiveWithNumbytesLargerThanSize() {
         final ByteArrayOutputStream bos = new ByteArrayOutputStream();
-        final String pax = paxRecord("GNU.sparse.size", "1048576") + paxRecord("GNU.sparse.map", "0,1048576");
+        final String pax = paxRecord("GNU.sparse.size", String.valueOf(SPARSE_MAP_NUMBYTES)) + paxRecord("GNU.sparse.map", "0," + SPARSE_MAP_NUMBYTES);
         final byte[] paxData = pax.getBytes(java.nio.charset.StandardCharsets.US_ASCII);
         appendPadded(bos, ustarHeader("PaxHeaders/sparse.bin", paxData.length, 'x'));
         appendPadded(bos, paxData);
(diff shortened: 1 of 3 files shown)


src/test/java/org/apache/commons/compress/archivers/tar/SparseFilesTest.java
```diff
@@ -532,4 +536,89 @@ void testTarFilePaxGNU() throws IOException {
             assertPaxGNUEntry(entries.get(2), "1.0");
         }
     }
+
+    private static void appendPadded(final ByteArrayOutputStream out, final byte[] data) {
+        out.write(data, 0, data.length);
+        final int rem = data.length % 512;
```
garydgregory: Define or reuse a constant for the magic number `512`.

src/test/java/org/apache/commons/compress/archivers/tar/SparseFilesTest.java
```diff
@@ -532,4 +536,89 @@ void testTarFilePaxGNU() throws IOException {
...
+        final byte[] bytes = value.getBytes(java.nio.charset.StandardCharsets.US_ASCII);
+        System.arraycopy(bytes, 0, header, offset, Math.min(length, bytes.length));
+    }
+
+    private static void putOctal(final byte[] header, final int offset, final int length, final long value) {
+        putField(header, offset, length, String.format("%0" + (length - 1) + "o", value));
+    }
+
+    private static byte[] ustarHeader(final String name, final long size, final char typeFlag) {
+        final byte[] header = new byte[512];
```
garydgregory: Define or reuse a constant for the magic number 512.

src/test/java/org/apache/commons/compress/archivers/tar/SparseFilesTest.java
```diff
@@ -532,4 +536,89 @@ void testTarFilePaxGNU() throws IOException {
...
+        final String pax = paxRecord("GNU.sparse.size", "1048576") + paxRecord("GNU.sparse.map", "0,1048576");
+        final byte[] paxData = pax.getBytes(java.nio.charset.StandardCharsets.US_ASCII);
+        appendPadded(bos, ustarHeader("PaxHeaders/sparse.bin", paxData.length, 'x'));
+        appendPadded(bos, paxData);
+        appendPadded(bos, ustarHeader("sparse.bin", 4, '0'));
+        appendPadded(bos, new byte[] { 1, 2, 3, 4 });
+        final byte[] payload = "the following entry".getBytes(java.nio.charset.StandardCharsets.US_ASCII);
+        appendPadded(bos, ustarHeader("next.bin", payload.length, '0'));
+        appendPadded(bos, payload);
+        bos.write(new byte[1024], 0, 1024); // two zero EOF records
```
garydgregory: Define or reuse a constant for the magic number 1024. Is it a coincidence that this test uses this number which is 512 *2, another magic number in the test?

src/test/java/org/apache/commons/compress/archivers/tar/SparseFilesTest.java
```diff
@@ -532,4 +536,89 @@ void testTarFilePaxGNU() throws IOException {
...
+        header[155] = ' ';
+        return header;
+    }
+
+    // Builds a PAX GNU 0.1 sparse entry whose stored size is 4 bytes but whose sparse map claims a
+    // single 1 MiB data block. realsize equals the block size, so the existing realSize check passes.
+    // A second entry follows, whose bytes must not be reachable through the sparse entry.
+    private byte[] sparseArchiveWithNumbytesLargerThanSize() {
+        final ByteArrayOutputStream bos = new ByteArrayOutputStream();
+        final String pax = paxRecord("GNU.sparse.size", "1048576") + paxRecord("GNU.sparse.map", "0,1048576");
```
garydgregory: Is `1048576` used both on this line meant to be exactly the same? If so, refactor.

garydgregory (changes requested):
Hello @kali834x 
Thank you for the PR. Please see my comments.

src/test/java/org/apache/commons/compress/archivers/tar/SparseFilesTest.java
kali834x: Done, reused TarConstants.DEFAULT_RCDSIZE since the test lives in the same package.

src/test/java/org/apache/commons/compress/archivers/tar/SparseFilesTest.java
kali834x: Switched to TarConstants.DEFAULT_RCDSIZE here as well.

src/test/java/org/apache/commons/compress/archivers/tar/SparseFilesTest.java
kali834x: Not a coincidence, it's exactly two 512-byte end-of-archive records. Now written as 2 * TarConstants.DEFAULT_RCDSIZE so the relationship is explicit.

src/test/java/org/apache/commons/compress/archivers/tar/SparseFilesTest.java
kali834x: Yes, intentionally the same value: realsize has to equal the map's numbytes sum so the archive passes the existing realSize check and only the new bound rejects it. Pulled it into a SPARSE_MAP_NUMBYTES constant used by both records.

kali834x:
Pushed the cleanup: 512 and the 1024 EOF padding now use TarConstants.DEFAULT_RCDSIZE, and the 1 MiB value is a shared constant. SparseFilesTest and checkstyle are still green.

garydgregory:
Hello @kali834x 
Please request a Copilot review; let's see if it finds anything.

kali834x:
I tried, but GitHub rejects reviewer requests from my account since that needs write access on the repo, so Copilot will have to be requested from your side. I'll pick up anything it flags.

garydgregory:
@kali834x 
Check; I requested a copilot review...

PR #782, opened 2026-07-09
Reject negative element counts in BandSet band decoding

kali834x:
**Negative band counts reach array allocation in the Pack200 decoder**

The array-count overloads of `decodeBandInt`, `parseFlags` and `parseReferences` size each sub-array straight from `counts[i]` after only an upper-bound check, but count bands are decoded with signed codecs like `DELTA5`, so a crafted archive can make an element negative and turn `new int[counts[i]]` (and its `long[]`/`String[]` siblings) into a `NegativeArraySizeException` that escapes the declared `Pack200Exception` contract. The scalar `decodeBandInt(..., int)` overload already rejects `count < 0`, so this applies the same guard in the three siblings and a corrupt band surfaces as a normal `Pack200Exception`; the added `BandSetTest` case covers all three.

### DIFF
--- a/src/main/java/org/apache/commons/compress/harmony/unpack200/BandSet.java
+++ b/src/main/java/org/apache/commons/compress/harmony/unpack200/BandSet.java
@@ -151,6 +151,9 @@ public abstract class BandSet {
         final int[][] result = new int[counts.length][];
         int totalCount = 0;
         for (final int count : counts) {
+            if (count < 0) {
+                throw new Pack200Exception("count < 0");
+            }
             totalCount += count;
         }
         final int[] twoDResult = decodeBandInt(name, in, defaultCodec, totalCount);
@@ -549,6 +552,9 @@ public abstract class BandSet {
         int sum = 0;
         final long[][] result = new long[count][];
         for (int i = 0; i < count; i++) {
+            if (counts[i] < 0) {
+                throw new Pack200Exception("count < 0");
+            }
             sum += counts[i];
         }
         int[] hi = null;
@@ -630,6 +636,9 @@ public abstract class BandSet {
         }
         int sum = 0;
         for (int i = 0; i < count; i++) {
+            if (counts[i] < 0) {
+                throw new Pack200Exception("count < 0");
+            }
             sum += counts[i];
         }
         // TODO Merge the decode and parsing of a multiple structure into one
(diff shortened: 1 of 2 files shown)


garydgregory (commented):
@kali834x 
Is a 0 count legal in pack200?

kali834x:
Yes, 0 is legal and common. A 0 count just produces an empty sub-array, e.g. class_interface_count is 0 for every class that implements no interfaces, and ClassBands feeds those counts straight into decodeBandInt. That's why the guard only rejects negatives, same as the existing check in the scalar decodeBandInt(..., int) overload.

garydgregory:
@kali834x 
What about `parseCPSignatureReferences()`? What does the pack200 specification says?

garydgregory:
@kali834x 
What is the maximum legal value for each sum?

kali834x:
Yes, same pattern there. The int[] overloads of parseCPSignatureReferences and parseCPUTF8References sum the counts with a plain loop and size each sub-array straight from counts[i], so a counts array like {-1, 2} keeps the total non-negative, gets past the scalar guard in decodeBandInt, and then throws NegativeArraySizeException at new CPUTF8[-1]. parseCPUTF8References even allocates from counts[i] before decoding anything. ClassBands feeds these from band-decoded counts (localVariableTableN, anno_N, etc.), so the same crafted-archive path applies.

Spec-wise, count bands default to UNSIGNED5, which can't encode a negative, but the spec allows a secondary encoding to replace the primary for any band after the segment header, and signed codecs are eligible there. It never gives a negative count any meaning, so treating one as corrupt data is consistent with it.

On the maximum: the only explicit numeric limit the spec states is for the constant pool, "the arithmetic sum of all numbers in cp_counts must be less than 2^29" (536870912). The other count bands have no stated maximum of their own; they're bounded indirectly by the class file format (u2, so 65535, for interface/field/method/attribute style counts) and by each sum having to fit a Java array. Related: sumPositive accumulates with plain +=, so enough large counts can wrap the total back to a positive but wrong value and the failure becomes an OOME instead of a Pack200Exception. Pack200Exception.addExact would close that. I can put up a follow-up PR covering the two CP reference methods and the overflow check if you want it.

Thanks for merging this one.

garydgregory:
@kali834x 
Yes, let's do the check you mention.

PR #783, opened 2026-07-10
Reject negative element counts in pack200 decodeBandInt

kali834x:
the array-count form of decodeBandInt in BandSet slices a flat band into per-entry arrays but only rejects a counts[i] larger than the decoded length, never a negative one. a per-entry count read through UNSIGNED5 can pass Integer.MAX_VALUE and narrow to a negative int in BHSDCodec.decode, so a crafted pack200 archive whose counts still sum to a positive total reaches new int[counts[i]] and surfaces a NegativeArraySizeException instead of a Pack200Exception. the single-count decodeBandInt already guards count < 0; the same guard is added to the array form, which is what ClassBands feeds for class_attr_indexes and the line-number and local-variable tables.

### DIFF
--- a/src/main/java/org/apache/commons/compress/harmony/unpack200/BandSet.java
+++ b/src/main/java/org/apache/commons/compress/harmony/unpack200/BandSet.java
@@ -156,6 +156,9 @@ public abstract class BandSet {
         final int[] twoDResult = decodeBandInt(name, in, defaultCodec, totalCount);
         int index = 0;
         for (int i = 0; i < result.length; i++) {
+            if (counts[i] < 0) {
+                throw new Pack200Exception("count < 0");
+            }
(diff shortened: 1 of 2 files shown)


garydgregory:
@kali834x Please rebased on git master and resolve conflicts.

kali834x:
Rebased onto master locally and there's nothing left of this one. #782 already added the same negative count rejection to the array form of decodeBandInt, and the sumPositive refactor in 26bd84e throws before the band is even read, so the guard here would be dead code and the test duplicates testDecodeBandIntRejectsNegativeCount which is already on master (I reran it there to confirm). Fine to close this as redundant.

garydgregory:
Closing per @kali834x comment.

PR #789, opened 2026-07-22
Reject negative counts in pack200 parseCPUTF8/SignatureReferences

kali834x:
the int[] overloads of parseCPUTF8References and parseCPSignatureReferences in the pack200 bandset summed their per-entry counts with a plain loop and sized each sub-array straight from counts[i], unlike decodeBandInt, parseFlags and parseReferences which already route the total through sumNonNegative. a crafted archive whose code_LocalVariableTable_N, *_type_RS or anno_N band decodes to a negative count (unsigned5 is the default but a signed secondary codec can replace it for any band after the segment header) reaches new CPUTF8[-1] and throws NegativeArraySizeException out of ClassBands, escaping the declared Pack200Exception contract; the same unchecked sum can also wrap on large counts. route both through sumNonNegative so a corrupt band yields a normal Pack200Exception like the sibling overloads. follow-up to the sumNonNegative work in #782.


garydgregory:
TY @kali834x , merged 🚀

PR #790, opened 2026-07-28
Reject unknown file type in CPIO entry mode

kali834x:
**Invalid CPIO header fields escape as IllegalArgumentException**

`CpioArchiveEntry.setMode` accepts only the eight file types CPIO defines, so a crafted `c_mode` naming one of the seven other non-zero `S_IFMT` values (0170000, 030000, ...) throws a raw `IllegalArgumentException` out of `getNextEntry`, which declares `IOException`. `setSize` has the same problem in the two ASCII formats: a `-` in the size field parses to a negative value and the setter throws before the `getSize() < 0` checks are reached.

Rather than guarding individual call sites, `getNextCPIOEntry` now rethrows `IllegalArgumentException` from header parsing as `ArchiveException`, keeping the original exception as the cause, the same way `ArArchiveInputStream.getNextEntry` already does. This covers every entry setter the three header readers call. The `UnsupportedOperationException` sites in `CpioArchiveEntry` are format guards and each reader only calls the setters matching the format it constructs, so those are not reachable from input.

### DIFF
--- a/src/main/java/org/apache/commons/compress/archivers/cpio/CpioArchiveInputStream.java
+++ b/src/main/java/org/apache/commons/compress/archivers/cpio/CpioArchiveInputStream.java
@@ -307,27 +307,31 @@ public class CpioArchiveInputStream extends ArchiveInputStream<CpioArchiveEntry>
             IOUtils.consume(this);
         }
         readFully(buffer2, 0, buffer2.length);
-        if (CpioUtil.byteArray2long(buffer2, false) == MAGIC_OLD_BINARY) {
-            entry = readOldBinaryEntry(false);
-        } else if (CpioUtil.byteArray2long(buffer2, true) == MAGIC_OLD_BINARY) {
-            entry = readOldBinaryEntry(true);
-        } else {
-            System.arraycopy(buffer2, 0, buffer6, 0, buffer2.length);
-            readFully(buffer6, buffer2.length, buffer4.length);
-            final String magicString = ArchiveUtils.toAsciiString(buffer6);
-            switch (magicString) {
-            case MAGIC_NEW:
-                entry = readNewEntry(false);
-                break;
-            case MAGIC_NEW_CRC:
-                entry = readNewEntry(true);
-                break;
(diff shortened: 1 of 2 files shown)


garydgregory:
Hello @kali834x 
There plenty of other call sites in the class CpioArchiveEntry that throw IAE and here UOE. Do think the ones that are involved in other input call sites warrant the same treatment? I think we need a comprehensive solution instead of a one-off for a single call site, if this issue should be addressed at all, since IAE is thrown from so many call sites...

kali834x:
Went through CpioArchiveEntry to see what is actually reachable from the three header readers. From archive input it's setMode (any of the seven undefined S_IFMT values, in all three formats) and setSize (the two ASCII formats can parse a negative size via a `-` in the field, and the `getSize() < 0` checks sit after the setter so they never fire). The UOE sites are all checkNewFormat/checkOldFormat guards, and each reader only calls the setters matching the format it constructs, so those can't trigger from input.

So agreed, a one-off wasn't the right shape. I dropped the helper and moved the conversion into getNextCPIOEntry instead: a single catch that rethrows IllegalArgumentException from header parsing as ArchiveException with the cause kept, same as ArArchiveInputStream.getNextEntry already does. That covers every setter the readers call, now and in the future, and added a regression test for the negative size case. I do think it's worth addressing since getNextEntry only declares IOException, and byte-mutating the bundled cpio test files reaches the setMode IAE readily.

kali834x:
any update?

garydgregory:
I'll review again a bit later.

garydgregory:
@kali834x 
I added the tests from this PR but _not_ the main changes as they are no longer needed. You are credited in `changes.xml`.
Please verify and close or update.
Thank you.

kali834x:
Verified on master: setMode and setSize now throw the checked ArchiveException themselves, so the wrapping in getNextCPIOEntry is redundant. Ran CpioArchiveInputStreamTest at b0e13a9 and all 15 tests pass, including the four from this PR. Closing, and thanks for the credit in changes.xml.

PR #794, opened 2026-08-08
Reject negative ZIP64 offsets in positionAtCentralDirectory64

kali834x:
**Negative ZIP64 offsets escape ZipFile as IllegalArgumentException**

`positionAtCentralDirectory64` reads the 8-byte ZIP64 end-of-central-directory offsets as signed longs (`ZipEightByteInteger.getLongValue`) and passes them straight to the channel's `position()`, so an archive whose offset has the high bit set makes `position()` throw a raw `IllegalArgumentException` out of the `ZipFile` constructor, which declares only `IOException`. Every other size and offset field in this class already rejects negatives with `ArchiveException`; the four ZIP64 offset reads (locator and record, split and non-split) were missing that check, so a broken archive now fails with `ArchiveException`. The 32-bit path is unaffected because `ZipLong.getValue` is unsigned.


garydgregory:
Thank you @kali834x , merged 🚀

PR #801, opened 2026-08-11
Reject oversized four-byte literal length in snappy decoder

kali834x:
snappycompressorinputstream.readliterallength reads the four-byte literal length with byteutils.fromlittleendian and casts it straight to int, so a length field of 0xffffffff narrows to -1 and, after the +1 the method adds, becomes a zero-length literal. fill only guards against a negative literal size, so the zero-length literal slips through, consumes no input and emits no output, and read() then recurses into itself once per element until a crafted run of them overflows the stack with a stackoverflowerror that escapes the declared ioexception. reading the four-byte value as an unsigned long and rejecting anything that can't fit a positive int before narrowing keeps the check at the point of the conversion; the two- and three-byte forms can't overflow so they're untouched. reachable directly and through framedsnappycompressorinputstream, which decodes each chunk with this class.

- [x] Write unit tests that match behavioral changes, where the tests fail if the changes to the runtime are not applied.
- [x] Each commit in the pull request should have a meaningful subject line and body.

### DIFF
--- a/src/main/java/org/apache/commons/compress/compressors/snappy/SnappyCompressorInputStream.java
+++ b/src/main/java/org/apache/commons/compress/compressors/snappy/SnappyCompressorInputStream.java
@@ -246,7 +246,7 @@ public class SnappyCompressorInputStream extends AbstractLZ77CompressorInputStre
         case 63:
             final long fourByteLength = ByteUtils.fromLittleEndian(supplier, 4);
             if (fourByteLength + 1 > Integer.MAX_VALUE) {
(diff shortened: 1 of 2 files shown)


garydgregory:
Switched to draft. There are many callers of the API that downcast the result to an int and therefore this issue needs a general solution, which coincidentally, I happen to be working on.

kali834x:
Makes sense, a fix at the ByteUtils level covers all the call sites at once. Happy to rebase this onto your change once it lands, or close it if it's fully covered there.

PR #802, opened 2026-08-14
wrap gzip FNAME/FCOMMENT charset errors as CompressorException

kali834x:
The gzip reader parses the FNAME and FCOMMENT header fields by decoding their raw bytes with the configured file-name charset and handing the result to GzipParameters.setFileName/setComment. Those setters re-encode the string and reject a NUL byte with an IllegalArgumentException, a guard aimed at the write path. When a caller selects a non-default multibyte charset through Builder.setFileNameCharset (UTF-32BE for example), decoding a malformed field yields a replacement character that re-encodes with a NUL byte, so a crafted member makes the reader throw IllegalArgumentException out of the constructor and out of read(), both of which only declare IOException. I hit it while feeding a fuzzed member through a non-Latin-1 charset. This lines up with the recent work moving these readers to CompressorException, so I catch the IllegalArgumentException at the two parse sites and rethrow it there. I kept the change in the reader so the public GzipParameters setters stay as they are.

### DIFF
--- a/src/main/java/org/apache/commons/compress/compressors/gzip/GzipCompressorInputStream.java
+++ b/src/main/java/org/apache/commons/compress/compressors/gzip/GzipCompressorInputStream.java
@@ -411,11 +411,19 @@ public class GzipCompressorInputStream extends CompressorInputStream implements 
         }
         // Original file name
         if ((flg & FLG.FNAME) != 0) {
-            parameters.setFileName(new String(readToNull(inData), parameters.getFileNameCharset()));
+            try {
+                parameters.setFileName(new String(readToNull(inData), parameters.getFileNameCharset()));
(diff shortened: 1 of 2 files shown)


garydgregory:
Hello @kali834x

Thank you for the PR.

I took a more straightforward approach in commit 1ecd39bc5bda8b3b529aabd0a80384bed41a4984 in line with recent changes. Please verify your use case. I think you can then close this PR. You are credited in `changes.xml`.

kali834x:
Verified against master with 1ecd39b: my crafted member (FNAME read with UTF-32BE) now throws CompressorException from read() instead of the raw IllegalArgumentException, so my use case is covered. Closing, and thanks for the credit in changes.xml.

